PR #1, opened 2016-09-24
Added prototype for clock "macro" file emission

azidar:
Unfinished, do not merge.

I've sketched out what the Firrtl pass may look like. There is a lot more makefile hacking that needs to be done to make everything work.

There isn't a good way to detect a "clock" signal per se, because a PLL/clock divider will be a blackbox with a clock input and multiple clock outputs. My assumption is that I should annotate the top module's input clock, as well as any clock signals that are module outputs.

@palmer-dabbelt Any comments?


edwardcwang:
@azidar Can we close this, or were you still planning to do something with this?

PR #4, opened 2016-10-29
PFPMP rule now use CORE_DIR for firrtl location

colinschmidt:


colinschmidt:
I tied to go back and just push this to master but it already made the branch and PR so oh well.

PR #6, opened 2017-01-25
unset DISPLAY in dc run-synthesis

jackkoenig:
At the end of synthesis, dc tries to open a GUI that I don't have a license
for. Unsetting DISPLAY stops this from happening.


palmer-dabbelt:
Thanks!

PR #9, opened 2017-01-29
Formality

donggyukim:
We can just use the Synopsys's Formality script to run it. I also make it report matching points and optionally run verify for strober. Now, rocket-chip passes formality with saed32.


palmer-dabbelt:
This should fail if "SYNTHESIS_TOOL" is not "dc", since it looks into DC's workdir directly.

palmer-dabbelt:
Thanks!  This has been merged as https://github.com/ucb-bar/plsi/commit/209e4f7f6bbaef102f97d109cdef004c67e3d294 but it looks like github isn't picking it up so I'm closing the PR.

PR #12, opened 2017-02-02
Fix typos

edwardcwang:


palmer-dabbelt:
Thanks.  There's probably a lot more... :)

PR #16, opened 2017-02-03
Refactor ICC par

edwardcwang:


src/addons/par/icc/tools/icc_common.sh
```diff
@@ -71,7 +74,6 @@ cat >>"$run_dir"/enter <<EOF
 export ICV_HOME_DIR="$(dirname $(dirname $(dirname $icv)))"
 export PATH="$(dirname $icv):\$PATH"
 EOF
-fi
```
palmer-dabbelt: This was a mistake I made that was fixed in https://github.com/ucb-bar/plsi/commit/cabd73251f150e900a3b1b086941d4c54af0810c, can you move it over?

edwardcwang:
Fix ported over!

palmer-dabbelt:
Thanks!

PR #17, opened 2017-02-04
SAED32 redundant vias fix

edwardcwang:


colinschmidt:
Is this the only technology specific code in PLSI? You should probably at least ifdef it no?

edwardcwang:
Eventually it'll be refactored into the technology files themselves I think once we've figured out how exactly to make this work, but for now I think a TODO should suffice since time of the essence.

colinschmidt:
Not to be a jerk but these TODO's are why cad tool scripts are hard to use.
Would you mind ifdefing it out so this doesn't fail for some other technology needlessly?

edwardcwang:
After some further discussion, I think there is a way to do this more generically. I will do that instead with a more generic solution.

edwardcwang:
Closing this as refactor is about to happen anyway, and we will use the new config system avoid issues like this.

PR #23, opened 2017-03-12
pfpmp -> barstools

edwardcwang:


Makefile
```diff
@@ -646,15 +649,16 @@ $(OBJ_TOOLS_SRC_DIR)/pson/Configfile: $(shell find src/tools/pson -type f)
...
+	@mkdir -p $(dir $@)
+	$< --sbt $(abspath $(CMD_SBT)) --project tapeout --sbtargs "-DFIRRTL_HOME=$(abspath $(FIRRTL_HOME))" --basedir $(abspath $(OBJ_TOOLS_BIN_DIR)/barstools) --main "barstools.tapeout.transforms.$(notdir $@)" --output $@
 
-$(OBJ_TOOLS_BIN_DIR)/pfpmp/stamp: $(shell find src/tools/pfpmp/src -type f) $(shell find $(CORE_DIR)/firrtl/src -type f)
+$(OBJ_TOOLS_BIN_DIR)/barstools/stamp: $(shell find src/tools/barstools/tapeout -type f) $(shell find $(FIRRTL_HOME)/src -type f)
 	rm -rf $(dir $@)
 	mkdir -p $(dir $@)
-	rsync -a --delete src/tools/pfpmp/ $(dir $@)
-	$(SCHEDULER_CMD) --make -- $(MAKE) FIRRTL_HOME=$(abspath $(CORE_DIR)/firrtl) CMD_SBT=$(abspath $(CMD_SBT)) -C $(dir $@)
+	rsync -a --delete src/tools/barstools/ $(dir $@)
```
colinschmidt: I think this stuff was necessary before when palmer needed to change the firrtl source but I integrate with barstools without this hack in hurricane-chip now.

Makefile
```diff
@@ -121,8 +124,8 @@ CMD_SBT = $(OBJ_TOOLS_BIN_DIR)/sbt/sbt
 CMD_GCC = $(OBJ_TOOLS_BIN_DIR)/gcc-$(GCC_VERSION)/bin/gcc
 CMD_GXX = $(OBJ_TOOLS_BIN_DIR)/gcc-$(GCC_VERSION)/bin/g++
 CMD_PSON2JSON = $(OBJ_TOOLS_BIN_DIR)/pson/bin/pson2json
-CMD_FIRRTL_GENERATE_TOP = $(OBJ_TOOLS_BIN_DIR)/pfpmp/bin/GenerateTop
-CMD_FIRRTL_GENERATE_HARNESS = $(OBJ_TOOLS_BIN_DIR)/pfpmp/bin/GenerateHarness
+CMD_FIRRTL_GENERATE_TOP = $(OBJ_TOOLS_BIN_DIR)/barstools/bin/GenerateTop
+CMD_FIRRTL_GENERATE_HARNESS = $(OBJ_TOOLS_BIN_DIR)/barstools/bin/GenerateHarness
```
colinschmidt: There is a GenerateTopAndHarness now. If you need to generate both at the same time in saves some time.

Makefile
```diff
@@ -110,6 +110,9 @@ TRACE_PAR_DIR = trace/par-$(CORE_GENERATOR)-$(CORE_CONFIG)-$(SOC_CONFIG)-$(TECHN
 SIGNOFF_SYN_FORMAL_DIR = check/formal/syn-$(CORE_GENERATOR)-$(CORE_CONFIG)-$(SOC_CONFIG)-$(TECHNOLOGY)-$(MAP_CONFIG)-$(SYN_CONFIG)
 SIGNOFF_SYN_POWER_DIR = check/power/syn-$(CORE_GENERATOR)-$(CORE_CONFIG)-$(SOC_CONFIG)-$(MAP_CONFIG)-$(SYN_CONFIG)
 
+# Define the location of FIRRTL (used by barstools, etc).
+FIRRTL_HOME ?= src/tools/firrtl
```
colinschmidt: Will this work if it is overridden to point to e.g. the one in rocket-chip? I think its good to allow the user to do that but I just want to make sure that use case is supported.

edwardcwang: Yes, it would work fine.

Makefile
edwardcwang: I see. Currently I think they're invoked separately (and I'm not sure that GenerateHarness is actually invoked at all), so it might be more convenient to leave them separate for now until that mess is sorted out.

Makefile
edwardcwang: Yes, I agree. It might be nice too to have a mode where we don't touch the main directories and try to do as much work in object folders as possible (e.g. running multiple PLSI instances simultaneously from one folder), so I added a mode that lets us do that if the source dir is read-only.

edwardcwang:
@colinschmidt New review please?

colinschmidt:
LGTM

PR #24, opened 2017-04-04
Remove commercial technology targets

palmer-dabbelt:


palmer-dabbelt:
@kasanovic This is trivial.  I can rewrite history if you'd like, but that's a bit more of a pain.

colinschmidt:
This level of description needs to be private?

palmer-dabbelt:
Here's Krste's email, that's why I tagged him as a reviewer

````
From:       krste@berkeley.edu
To:         palmer.dabbelt@eecs.berkeley.edu
Subject:    removing tech path names from public repo
Date:       Tue, 04 Apr 2017 03:40:04 PDT (-0700)
Message-ID: <m260ikbgmz.fsf@berkeley.edu>


Palmer,

The following files have path names into NDA tech files - I realize it
technically doesn't expose much because only the file names are given,
but we'll show up on searches (that TSMC does regularly now after the
Stanford tech file incident) and legally this is grey ground.

https://github.com/ucb-bar/plsi/blob/bar/src/technologies/tsmc28hpm.tech.json
https://github.com/ucb-bar/plsi/blob/bar/src/technologies/st28fdsoi.tech.json

I realize this makes the stuff harder to use, but that's life using
NDA tools.  The public library versions are fine to leave in as
examples to help guide people write these files.

This is urgent, and the bad thing is git holds these in repo history,
so would ideally scrub from there too.

Thanks,
Krste
````

colinschmidt:
Thanks for the info.

edwardcwang:
What are the "public library versions"?

palmer-dabbelt:
The ones I didn't delete (SAED32 and OSU180)

PR #25, opened 2017-05-03
Donggyu's fixes

donggyukim:
This fixes PrimeTime scripts:
* `STRIP_PATH` is important when reading VCD/SAIF files, which is set by `TESTBENCH`.
* Power from PAR is enabled.
* Includes minor bug fixes.


donggyukim:
By the way, what's the difference between bar & master branches?

Makefile
```diff
@@ -22,6 +22,10 @@ CORE_GENERATOR ?= rocketchip
 # soc generator doesn't actually do anything at all.
 SOC_GENERATOR ?= nop
 
+# The testbench for simulation
+# FIXME: what's the default?
+TESTBENCH ?= none
```
edwardcwang: What exactly is this argument?

edwardcwang:
To first order, the bar branch is the "stable" branch whereas the master branch is the branch with more active development and features.

Makefile
donggyukim: This is necessary for setting `STRIP_PATH` in the PrimeTime script(path from the top module to DUT), which is generally the testbench. Otherwise, PrimeTime fails in signal activity annotations. I guess there may be a much better name, though.

edwardcwang: I see. Maybe something like PATH_TO_TESTBENCH and give e.g. main/module1 if main.module1.testbench is the hierarchy?

Also, if only PrimeTime uses it for now, could we make it a PrimeTime specific argument, like https://github.com/ucb-bar/plsi/blob/bar/src/addons/core-generator/chisel/vars.mk#L12 ?

donggyukim: Sure, can you directly fix it on this PR?

edwardcwang:
@donggyukim Using PRIMETIME_PATH_TO_TESTBENCH in place of TESTBENCH, can you let me know if it breaks anything? If it seems good I'll go ahead and merge this. Thanks!

PR #26, opened 2017-06-08
New Vivado synthesis fixes from tapeout

edwardcwang:


src/addons/soc-generator/vivado/pins.csv
```diff
@@ -57,3 +57,29 @@ Comment,Name of the pin,“input” or “output”
...
+,jd_5_o,input
+,jd_5_i,output
+,jd_5_oe,output
+,jd_6_o,input
+,jd_6_i,output
+,jd_6_oe,output
+,jd_7_o,input
+,jd_7_i,output
+,jd_7_oe,output
+,uart_rxd_out,output
```
edwardcwang: The UART pins are mono-direction pins that don't require a tristate IOBUF driver (like the LEDs, which also don't require a driver). But it can't hurt to have some, so I added them.

PR #27, opened 2017-06-08
DC, ICC, and general improvements from the tapeout class

edwardcwang:
There is more to come after the tapeout actually tapes out.


src/addons/par/icc/rules.mk
```diff
@@ -75,4 +75,4 @@ $(OBJ_PAR_DIR)/generated/$(PAR_TOP).macros.out: $(OBJ_PAR_DIR)/synopsys-icc-macr
 $(OBJ_PAR_DIR)/generated/$(PAR_TOP).floorplan.json: \
 		$(PAR_TOOL_ADDON)/tools/generate-floorplan-json \
 		$(OBJ_PAR_DIR)/generated/$(PAR_TOP).macros.out
-	$(abspath $<) --macros $(filter %.macros.out,$^) --rtl_top $(SYN_TOP) --config $(PAR_CONFIG) -o $@
+	PATH="$(abspath $(dir $(CMD_PYTHON3))):$(PATH)" $(abspath $<) --macros $(filter %.macros.out,$^) --rtl_top $(SYN_TOP) --config $(PAR_CONFIG) -o $@
```
colinschmidt: It feels like there must be some better way to organize this. Maybe roll it into a higher level call? Are there python2 programs still floating around that we want to avoid having called with python3?

src/tools/technology/generate-vars
```diff
@@ -38,12 +38,22 @@ for library_desc in techfile["libraries"]:
 		of.write(" {0}/{1}".format(tech_home, library_desc["nldm library file"]))
 of.write("\n")
 
+# Verilog files required for simulation only.
 of.write("TECHNOLOGY_VERILOG_FILES =")
```
colinschmidt: I think we should change this name to be explicit i.e. TECHNOLOGY_VERILOG_SIMULATION_FILES

src/tools/technology/generate-vars
```diff
@@ -76,13 +86,13 @@ for library_desc in techfile["libraries"]:
...
 
-of.write("TECHNOLOGY_MARCO_PROVIDE_SCRIPTS =")
+of.write("TECHNOLOGY_MACRO_PROVIDE_SCRIPTS =")
 for library_desc in techfile["libraries"]:
 	if "provide script" in library_desc:
 		of.write(" {0}".format(library_desc["provide script"]))
 of.write("\n")
 
-of.write("TECHNOLOGY_MARCO_DEPEND_SCRIPTS =")
+of.write("TECHNOLOGY_MACRO_DEPEND_SCRIPTS =")
```
colinschmidt: Whats going on with these to changes? Github says they are different but they look identical to me

edwardcwang: There's a typo fix (MA*R*CO) vs (MA*C*RO)

Makefile
```diff
@@ -1,5 +1,9 @@
 # Copyright 2016-2017 Palmer Dabbelt <palmer@dabbelt.com>
 
+# Disable the built-in rules to speed up execution.
+MAKEFLAGS += -r
```
edwardcwang: Yes, often when I'm debugging the makefile I'll put -d to see what rules it's running and having the -r reduces spam and makes the log a lot more readable.

src/addons/par/icc/rules.mk
edwardcwang: Yes, the problem is that the BWRC servers don't have python3, so any script that uses python3 has to have CMD_PYTHON3 appended to its path.

src/addons/par/icc/tools/icc_common.sh
```diff
@@ -3,6 +3,24 @@
...
+# http://stackoverflow.com/a/17841619
+function join_by { local d=$1; shift; echo -n "$1"; shift; printf "%s" "${@/#/$d}"; }
+
+# Older versions of readlink, like readlink 8.4, don't support multiple
+# arguments to readlink...
+# Usage: x=$(readlink_array my_array)
+readlink_array() {
+    local ref=$1[@]
+    local v=(${!ref})
+    local output=$(join_by " " "${v[@]}")
```
edwardcwang: Yes, on the BWRC servers the system readlink binary does not support multiple files, so I replaced it with a shell function.

That being said, it might perhaps be cleaner to implement by own readlink script and then just put that ahead in the path. That way we have a (slightly) simpler syntax and avoid this shell function. Thoughts?

colinschmidt: I guess my real issue was that this bash syntax is complicated. Is the ```!``` in front of ```ref``` what expands the path? I was thinking this would repeatedly call ```readlink -f``` but I can't see how it does that.

src/addons/par/icc/rules.mk
colinschmidt: Yeah so if there is no problem with just always using ```CMD_PYTHON3``` I don't see why we don't just set ```PATH``` that way at the top-level makefile to reduce this code duplication.

src/tools/technology/generate-vars
edwardcwang: Done

src/addons/par/icc/rules.mk
edwardcwang: I agree; the main annoying issue, however, is that Palmer set up PLSI to build CMD_PYTHON3 as a dependency in the rules (i.e. it builds python3 from source). Making it global would require me to refactor the way PLSI handles these kinds of binary/tool dependencies, which is something that is on my TODO list and will be handled in a future commit. Let me know what you think.

src/addons/par/icc/tools/icc_common.sh
edwardcwang: Yes, you're right that there's a bug - I neglected to actually call `readlink -f` repeatedly. Just ended up refactoring this into a separate script or readability.

Makefile
```diff
@@ -317,7 +321,7 @@ MAP_SIM_TOP = $(CORE_SIM_TOP)
 
 # There's always a mapped Verilog file since there's macros in it and I'm only
 # going to bother generating macros as Verilog.
-OBJ_MAP_RTL_V = $(OBJ_SOC_RTL_V) $(OBJ_MAP_DIR)/plsi-generated/$(MAP_TOP).macros_for_synthesis.v
+OBJ_MAP_RTL_V = $(OBJ_SOC_RTL_V) $(OBJ_MAP_DIR)/plsi-generated/$(MAP_TOP).macros_for_synthesis.v $(TECHNOLOGY_VERILOG_SYNTHESIS_FILES)
```
edwardcwang: Added example for how to use it. Will add the list of all variables to my TODO list.

edwardcwang:
@colinschmidt Addressed all the comments; would you mind taking another look?

src/addons/par/icc/tools/icc_common.sh
```diff
@@ -58,6 +58,11 @@ done
 
 set -ex
 
+if [[ "$script_dir" == "" ]]; then
+    echo "script_dir not set"
+    exit 1
+fi
+
```
colinschmidt: Don't you need to put script_dir in the PATH or something? Later you call readlink_array with no script_dir prefix and I don't understand how that will be available.

edwardcwang: The script_dir is not a PATH, but rather the path to the PLSI ICC scripts (typically src/addons/par/icc/tools). I've added comments to document this.

PR #28, opened 2017-07-10
No pcad

donggyukim:
I replaced `pcad-pipe-macro_compiler` & `pcad-pipe-list_macros`. I don't think we'll ever need `pcad-pipe-infer_decoupled`. Depends on ucb-bar/barstools#24


edwardcwang:
Will take a look at this and probably merge after #29 is merged.

PR #41, opened 2017-11-22
Don't assume current working directory is hammer-vlsi

edwardcwang:
Now we don't need $PWD and $(readlink -f) when calling hammer-vlsi from the CLI

Templates to be updated as well.

@colinschmidt


colinschmidt:
Thanks!

PR #54, opened 2018-02-04
Don't add empty configs if given empty strings

colinschmidt:
In addition to ignoring null lists also ignore empty strings in a list


edwardcwang:
Is the goal for hammer to ignore the empty string in  `hammer-vlsi ... -p my_snippet.json -p "" -p foo.json`? Or is there another way in which the empty string sneaks in?

colinschmidt:
If I don't set the environment variable then it puts a blank string in see https://github.com/ucb-bar/hammer/blob/master/src/hammer-vlsi/cli_driver.py#L74

colinschmidt:
I didn't want to mess up that one-liner since I didn't know enough python to make it look good. open to other suggestions

edwardcwang:
Since we want to prevent an unset environment variable from injecting a blank config but also want to catch `-p ""`, this seems like it will work.

PR #55, opened 2018-02-04
use absolute paths for cadence scripts

colinschmidt:
It seems like cadence tools don't like non-absolute paths or its just in the wrong place, either way this seems like a good solution.


colinschmidt:
There are a couple more of these coming so wait a bit to merge.

edwardcwang:
Thanks for alerting me to this - so it turns out that there is actually a bug in the cli_driver that causes obj_dir to be blank since it tries to get hammer_vlsi_path before it's ready. I'll push a PR.

edwardcwang:
#62

edwardcwang:
Closed via #62

PR #61, opened 2018-02-04
Remove old unused genus

edwardcwang:
Remove clutter


colinschmidt:
Great I was going to say this feels really silly/confusing to have around now.

edwardcwang:
Feel free to let me know if there's other old cruft lying around that can be cleaned up.

PR #75, opened 2018-02-08
with the new genus startup we write out sdc with view name

colinschmidt:
Missed this in my last changeset


colinschmidt:
yeah its hardcoded two places

PR #78, opened 2018-02-08
give a design node parameter and pass to innovus

colinschmidt:
Not sure if this is the right way to give options a default or if I should have nothing and then a check in the emission


edwardcwang:
I think this is fine, though it should probably still be overridden in technology plugins like https://github.com/ucb-bar/hammer/blob/master/src/hammer-vlsi/technology/saed32/defaults.yml

colinschmidt:
I see you are saying I should have also populated the current tech plugins?
If so that makes sense.

colinschmidt:
Ok I fixed conflicts and I believe addressed comments.

PR #83, opened 2018-02-12
Modularize hammer plugins into hookable/pausable steps

edwardcwang:
Genus is tested and appears to work. Innovus is next, hoping to get it done by tonight.

None of the other plugins are ported to this new style yet - they will come afterwards.

@colinschmidt Preliminary thoughts?

### DIFF
--- a/src/hammer-vlsi/hammer_vlsi.py
+++ b/src/hammer-vlsi/hammer_vlsi.py
@@ -680,6 +680,19 @@ class HammerTool(metaclass=ABCMeta):
         self._logger = value # type: HammerVLSILoggingContext
 
     ##############################
+    # Implementation helpers for properties
+    ##############################
+    def attr_getter(self, key: str, default: Any) -> Any:
+        """Helper function for implementing the getter of a property with a default."""
+        if not hasattr(self, key):
+            setattr(self, key, default)
+        return getattr(self, key)
+
+    def attr_setter(self, key: str, value: Any) -> None:
+        """Helper function for implementing the setter of a property with a default."""
+        setattr(self, key, value)
+
+    ##############################
     # Hooks
     ##############################
     def check_duplicates(self, lst: List[HammerToolStep]) -> Tuple[bool, Set[str]]:
--- a/src/hammer-vlsi/hammer_vlsi.py
+++ b/src/hammer-vlsi/hammer_vlsi.py
@@ -277,6 +277,19 @@ class HammerVLSISettings:
             "vlsi.builtins.hammer_vlsi_path": HammerVLSISettings.hammer_vlsi_path
         }
 
+    @classmethod
+    def set_hammer_vlsi_path_from_environment(cls) -> bool:
+        """
+        Try to set hammer_vlsi_path from the environment variable HAMMER_VLSI.
+        :return: True if successfully set, False otherwise
+        """
+        if "HAMMER_VLSI" not in os.environ:
+            return False
+        else:
+            cls.hammer_vlsi_path = os.environ["HAMMER_VLSI"]
+            return True
+
+
 class TimeValue:
     """Time value - e.g. "4 ns".
     Parses time values from strings.
--- a/src/hammer-vlsi/hammer_vlsi.py
+++ b/src/hammer-vlsi/hammer_vlsi.py
@@ -938,13 +938,13 @@ class HammerTool(metaclass=ABCMeta):
         return HammerTool.make_resume_hook(step, HookLocation.ResumePostStep)
 
     @staticmethod
-    def make_from_to_hooks(from_step: Optional[str] = None, to_step: Optional[str] = None) -> List[
-        HammerToolHookAction]:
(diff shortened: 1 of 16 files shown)


src/hammer-vlsi/hammer_vlsi.py
```diff
@@ -563,8 +679,313 @@ def logger(self, value: HammerVLSILoggingContext) -> None:
...
+                    do_step = False
+
+            if do_step:
+                try:
+                    if prev_step is None:
+                        # Run pre-step hook.
+                        self.do_pre_steps(step)
+                    else:
+                        # TODO: find a cleaner way of detecting a pause hook
+                        if step.name == "pause":
```
colinschmidt: It seems like you might not be able to have multiple pause steps since earlier you had been checking for duplicate names?

src/hammer-vlsi/synthesis/genus/__init__.py
```diff
@@ -3,21 +3,36 @@
...
-
-from functools import reduce
-from typing import Callable, Dict, List, Iterable, Any
+from typing import Dict, List, Any
 
 import os
 
+
 class Genus(HammerSynthesisTool, CadenceTool):
+    def fill_outputs(self) -> bool:
```
colinschmidt: Not sure I understand the name of this function. It seems to just check_outputs, not create or fill them.

src/hammer-vlsi/synthesis/genus/__init__.py
```diff
@@ -30,16 +45,69 @@ def export_config_outputs(self) -> Dict[str, Any]:
...
         return outputs
 
-    def do_run(self) -> bool:
-        self.create_enter_script()
+    @property
+    def steps(self) -> List[HammerToolStep]:
+        return self.make_steps_from_methods([
+            self.init_environment,
+            self.syn_generic,
+            self.syn_map,
```
colinschmidt: Do we want a syn_opt step by default?

src/hammer-vlsi/synthesis/genus/__init__.py
```diff
@@ -107,24 +193,16 @@ def verbose_append(cmd: str) -> None:
...
-        # TODO: move this check upwards?
-        mapped_v = output_verilog
-        if not os.path.isfile(mapped_v):
-            raise ValueError("Output mapped verilog %s not found" % (mapped_v)) # better error?
-        self.output_files = [mapped_v]
-
-        if not os.path.isfile(output_sdc):
-            raise ValueError("Output SDC %s not found" % (output_sdc)) # better error?
-        self.output_sdc = output_sdc
+        if bool(self.get_setting("synthesis.genus.generate_only")):
```
colinschmidt: Nice!

src/hammer-vlsi/hammer_vlsi.py
edwardcwang: Yeah, though the pause step is really meant to be injected once, so I'm not sure how much sense it makes to have multiple pause steps.

src/hammer-vlsi/hammer_vlsi.py
```diff
@@ -563,8 +679,313 @@ def logger(self, value: HammerVLSILoggingContext) -> None:
...
+                if has_step(action.step.name):
+                    self.logger.error("New step '{step}' already exists".format(step=action.step.name))
+                    return False
+                new_steps.insert(step_id, action.step)
+            elif action.location == HookLocation.InsertPostStep:
+                if has_step(action.step.name):
+                    self.logger.error("New step '{step}' already exists".format(step=action.step.name))
+                    return False
+                new_steps.insert(step_id + 1, action.step)
+            elif action.location == HookLocation.ResumePreStep or action.location == HookLocation.ResumePostStep:
```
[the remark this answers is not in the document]
edwardcwang: ResumePre means resume just before this step; ResumePost means resume after this step.

edwardcwang:
Okay, Innovus, DC, and ICC are all ported now, and tested.

colinschmidt:
Ok the one question I still have is about the do_betweens, does it happen between all steps? My example use case is something like I want to write_db(checkpoint/make resumable) after syn_map and syn_opt, but not necessarily `pause` at those points.

Either way I think we should merge now since I can just add those as hooks.

Thanks for all this hard work!

src/hammer-vlsi/par/innovus/__init__.py
```diff
@@ -23,14 +23,49 @@ def env_vars(self) -> Dict[str, str]:
...
+        self.verbose_tcl_append(cmd, self.output)
+
+    @property
+    def steps(self) -> List[HammerToolStep]:
+        return self.make_steps_from_methods([
+            self.init_design,
+            self.place_opt_design,
+            self.route_design,
+            self.opt_design,
+            self.write_design
```
colinschmidt: This just feels so much nicer!

edwardcwang:
Thanks! To answer the question: do_between by default dumps the db but doesn't pause by default. Will go ahead and merge soon (just realized that I didn't port the nop plugins).

PR #84, opened 2018-02-13
Use line buffering in popen

### LINKED ISSUE #64: Output from tools being buffered too long
Unclear if this is a hammer or outer build system problem (lsf) but it certainly is annoying, and running tools manually in lsf does not present this issue.

colinschmidt:
This makes it easier to tail hammer's output if using lsf or something else.
Some cad commands take a long time and so not knowing exactly which line hammer is on can be confusing.

Also this linebreak at the end keeps coming up in my diffs so maybe our editors are fighting, and we'll see if mine wins ;)

Closes #64

### DIFF
--- a/src/hammer-vlsi/hammer_vlsi.py
+++ b/src/hammer-vlsi/hammer_vlsi.py
@@ -1217,12 +1217,12 @@ class HammerTool(metaclass=ABCMeta):
         prog_tag = prog_name + " " + prog_args
         subprocess_logger = self.logger.context("Exec " + prog_tag)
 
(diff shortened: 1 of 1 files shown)


edwardcwang:
According to https://docs.python.org/3/library/subprocess.html#popen-constructor

```1 means line buffered (only usable if universal_newlines=True i.e., in a text mode)```

Do we need to set `universal_newlines=True` as well?

colinschmidt:
Experimentally it worked for me, so either its true by default or this is `a text mode` which seems likely.

colinschmidt:
Hmm it does seem not to be set by default. Do you want me to also set that?

edwardcwang:
If you can briefly check that `universal_newlines=True` doesn't cause any harm I think we may as well set it.

colinschmidt:
Still testing. Don't merge

colinschmidt:
I don't think so. closing until further notice

PR #91, opened 2018-02-13
Test hook on hook

colinschmidt:
This should work


edwardcwang:
If you run `src/test/unittests.sh` does it pass?

colinschmidt:
I get an "OK" at the bottom but also get this in the middle:
```
[synthesis] Target step 'change3' does not exist
[<global>] Synthesis tool mocksynth failed! Please check its output.
```
Which was my experience when I tried to do this for real.

edwardcwang:
Investigating.

PR #95, opened 2018-02-17
Make sure you have all sources included

colinschmidt:
I couldn't come up with a good name for the option to control this and wasn't quite sure you would want to disable it, so feedback on either would be good.


edwardcwang:
I think this is reasonable by default. If it's an issue we can always make it into a setting.

PR #100, opened 2018-02-26
Implement floorplan constraint infrastructure

edwardcwang:
Example usage at https://github.com/ucb-bar/hammer-nda-saed32-aspire-synthesis/blob/master/SodorTile.yml


edwardcwang:
The dummy is for debugging purposes mostly. We can remove it if you'd like.

colinschmidt:
No I think we should just merge things as they stand

PR #104, opened 2018-03-08
add orientation support to floorplan json

colinschmidt:
Did this a while ago but forgot to PR it.

I'm not sure if the strings are tech or tool specific so if we figure that out we might want a config that maps between independent versions and the specific ones.


edwardcwang:
Looks great. Does it still pass `unittests.sh`?

colinschmidt:
Yes it does.

edwardcwang:
Awesome sauce

PR #151, opened 2018-06-21
Floorplan obstructions API

colinschmidt:
This creates an API to add blockages/obstructions to your placement constraints.

The three types of constraints simply reflect my current usage and could be expanded.

### DIFF
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -131,11 +131,17 @@ vlsi.inputs:
   # - "toplevel" (top-level chip dimensions; may only occur once, for the top-level module)
   # - "hardmacro" (places this hard macro at a particular spot)
   # - "hierarchical" (marks this instance as part of hierarchical place and route)
+  # - "obstruction" (creates a blockage at a particular spot see obs_type)
   # TODO(edwardw): add rotation
   # x (float) - x coordinate in um
   # y (float) - y coordinate in um
   # width (float) - width in um
   # height (float) - height in um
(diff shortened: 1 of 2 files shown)


src/hammer-vlsi/defaults.yml
```diff
@@ -131,11 +131,17 @@ vlsi.inputs:
   # - "toplevel" (top-level chip dimensions; may only occur once, for the top-level module)
   # - "hardmacro" (places this hard macro at a particular spot)
   # - "hierarchical" (marks this instance as part of hierarchical place and route)
+  # - "obstruction" (creates a blockage at a particular spot see obs_type)
   # TODO(edwardw): add rotation
   # x (float) - x coordinate in um
   # y (float) - y coordinate in um
   # width (float) - width in um
   # height (float) - height in um
+  # layers (str) - String that enumerates layer(s) blocked by the obstruction
```
edwardcwang: How do we specify multiple layers in a single string? Is it CAD tool dependent?

Also might be useful to note which ones are optional here

colinschmidt: I have to figure out if its CAD tool dependent. I suppose it probably eventually will have to be but for now I was hoping to leave it as is.

For reference my current usage just looks like `layers: "M1 M2 M3 M4"`

edwardcwang: Would it be a lot of extra work to have it be a `List[str]`?

colinschmidt: No not much work at all. Thats a better solution.

colinschmidt:
Updated documentation and converted layers to a list

PR #155, opened 2018-06-27
No nested enum

edwardcwang:
Due to a Python bug/missing feature. This causes `import hammer_vlsi` to fail which is a fatal error for trying to use hammer.

Until python/cpython#7950 is merged (and Python 3.8 becomes widespread) we'll have to avoid nested Enum classes.

Sorry @colinschmidt  :-(

### DIFF
--- a/src/hammer-vlsi/hammer_vlsi_impl.py
+++ b/src/hammer-vlsi/hammer_vlsi_impl.py
@@ -192,6 +192,30 @@ OutputLoadConstraint = NamedTuple('OutputLoadConstraint', [
 ])
 
 
+class ObstructionType(Enum):
+    Place = 1
+    Route = 2
+    Power = 3
+
+    @classmethod
+    def __mapping(cls) -> Dict[str, "ObstructionType"]:
+        return {
+            "place": ObstructionType.Place,
+            "route": ObstructionType.Route,
+            "power": ObstructionType.Power
+        }
+
+    @staticmethod
+    def from_str(x: str) -> "ObstructionType":
+        try:
+            return ObstructionType.__mapping()[x]
+        except KeyError:
+            raise ValueError("Invalid obstruction type: " + str(x))
+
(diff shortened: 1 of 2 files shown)


src/hammer-vlsi/hammer_vlsi_impl.py
```diff
@@ -221,28 +245,6 @@ def from_str(x: str) -> "PlacementConstraintType":
     def __str__(self) -> str:
         return reverse_dict(PlacementConstraintType.__mapping())[self]
 
-    class ObstructionType(Enum):
```
eric-wieser: You could also just set `_ignore_ = ['ObstructionType']`

edwardcwang: That's fair, though I think that property is only present in Python 3.7+.

Python 3.6 results in an error message:
```
Traceback (most recent call last):
  File "/tmp/test3.py", line 2, in <module>
    class A(Enum):
  File "/tmp/test3.py", line 5, in A
    _ignore_ = ["B"]
  File "/usr/lib/python3.6/enum.py", line 84, in __setitem__
    raise ValueError('_names_ are reserved for future Enum use')
ValueError: _names_ are reserved for future Enum use
```

edwardcwang: Though even in 3.7+ it seems like `_ignore_` isn't what we want since we can't access the ignored member:
```
>>> A
<enum 'A'>
>>> A.B
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/mnt/data/dev/edwardcwang_cpython/Lib/enum.py", line 337, in __getattr__
    raise AttributeError(name) from None
AttributeError: B
```

eric-wieser: I stand corrected

PR #166, opened 2018-08-03
 Add custom/extra IP block API

edwardcwang:
Custom IP block API.

Provides a way for us to pass generated/pre-provided analog blocks to hammer without having to hack the technology library's JSON file.


edwardcwang:
@colinschmidt Any thoughts?

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -714,6 +714,30 @@ def check_input_files(self, extensions: List[str]) -> bool:
...
+            # If the key doesn't exist we can safely say there are no extra libraries.
+            return []
+
+        extra_libs = self.get_setting("vlsi.technology.extra_libraries")
+        if not isinstance(extra_libs, list):
+            raise ValueError("extra_libraries was not a list")
+        else:
+            return list(map(hammer_tech.HammerTechnology.parse_library, extra_libs))
+
+    def get_available_libraries(self) -> List[hammer_tech.Library]:
```
colinschmidt: This made me think about how we should handle "extra" libraries that don't exist. In particular I'm thinking about the case where we would generate them.
I assume the plan-of-record is:
I generate a json snippet with my IP library
I point to that json from hammer's extra_libraries feature
And if I ever want a new version of more libraries I'm going to update my hammer config to point to the new version(name/directory change of the json) or new file(for a freshly added library)

This seems sane but wanted to double check that this is how you were thinking about it to.

edwardcwang: Yes, I think we are on the same page. It is:

1) IP block generator generates hammer IR snippet of hammer_tech `libraries`
2) Use extra_libraries to point to the snippet

Also, the API is undergoing a slight revision to make it easier to handle paths (coming within a day or two).

PR #188, opened 2018-09-12
Work on Vivado to get it synthesizing for me

grebe:
@edwardcwang 

Unless I'm wrong, Vivado doesn't seem to be kept up to date. Here's my attempt to get Vivado back up. Does this look reasonable? I only kind of know what's going on in here, so any feedback is welcome.

### DIFF
--- a/src/hammer-vlsi/synthesis/vivado/__init__.py
+++ b/src/hammer-vlsi/synthesis/vivado/__init__.py
@@ -11,12 +11,6 @@ class VivadoSynth(HammerSynthesisTool):
         new_dict.update({})  # TODO: stuffs
         return new_dict
 
-    def temp_file(self, filename: str) -> str:
-        """Helper function to get the full path to a filename under temp_folder."""
-        if self.get_setting("synthesis.mocksynth.temp_folder", nullvalue="") == "":
-            raise ValueError("synthesis.mocksynth.temp_folder is not set correctly")
-        return os.path.join(self.get_setting("synthesis.mocksynth.temp_folder"), filename)
-
     @property
     def steps(self) -> List[HammerToolStep]:
         return self.make_steps_from_methods([
@@ -33,7 +27,7 @@ class VivadoSynth(HammerSynthesisTool):
 
     def fill_outputs(self) -> bool:
         # This tool doesn't really have outputs
-        self.output_files = []
+        self.output_files = ["post_opt.dcp"]
         return True
 
 tool = VivadoSynth
--- a/src/hammer-vlsi/synthesis/vivado/__init__.py
+++ b/src/hammer-vlsi/synthesis/vivado/__init__.py
@@ -1,0 +1,39 @@
+from hammer_vlsi import HammerSynthesisTool, HammerToolStep, deepdict
+
+from typing import Dict, List, Any
+
+import os
+
+class VivadoSynth(HammerSynthesisTool):
+    @property
+    def env_vars(self) -> Dict[str, str]:
+        new_dict = deepdict(super().env_vars)
+        new_dict.update({})  # TODO: stuffs
+        return new_dict
+
+    def temp_file(self, filename: str) -> str:
+        """Helper function to get the full path to a filename under temp_folder."""
+        if self.get_setting("synthesis.mocksynth.temp_folder", nullvalue="") == "":
+            raise ValueError("synthesis.mocksynth.temp_folder is not set correctly")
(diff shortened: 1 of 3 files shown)


edwardcwang:
@grebe Thanks so much for looking into this! Yes, the vivado plugin is rusty. I'll take a quick look at the code. This is actually really useful since I've been wanting to run some FPGA stuff for a while but haven't had the chance to port the plugin.

Also it looks like the hammer bot doesn't know how to fetch from branches of forks. Remind me to look into this at some point.

src/hammer-vlsi/synthesis/vivado/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+import os
+
+class VivadoSynth(HammerSynthesisTool):
+    @property
+    def env_vars(self) -> Dict[str, str]:
+        new_dict = deepdict(super().env_vars)
+        new_dict.update({})  # TODO: stuffs
+        return new_dict
+
+    def temp_file(self, filename: str) -> str:
```
edwardcwang: Stray method?

src/hammer-vlsi/synthesis/vivado/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+        self.run_executable([
+            os.path.join(self.tool_dir, "run-synthesis"),
+            "-o", self.run_dir,
+            "--top", self.top_module,
+        ] + self.input_files)
+        return True
+
+    def fill_outputs(self) -> bool:
+        # This tool doesn't really have outputs
+        self.output_files = []
```
edwardcwang: We should fill the outputs (post-synthesis netlist Verilog) here

src/hammer-vlsi/synthesis/vivado/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+
+    @property
+    def steps(self) -> List[HammerToolStep]:
+        return self.make_steps_from_methods([
+            self.run_synthesis,
+        ])
+
+    def run_synthesis(self) -> bool:
+        self.run_executable([
+            os.path.join(self.tool_dir, "run-synthesis"),
```
edwardcwang: I'm surprised this worked out of the box :P

src/hammer-vlsi/synthesis/vivado/__init__.py
```diff
@@ -0,0 +1,39 @@
+from hammer_vlsi import HammerSynthesisTool, HammerToolStep, deepdict
+
+from typing import Dict, List, Any
+
+import os
+
+class VivadoSynth(HammerSynthesisTool):
```
edwardcwang: We should probably add a `defaults.yml` to allow hammer to locate the vivado binary

src/hammer-vlsi/synthesis/vivado/__init__.py
grebe: Yeah, there's some copypasta.

src/hammer-vlsi/synthesis/vivado/__init__.py
grebe: In this case, is that a dcp?

edwardcwang: I think dcp is good although it would also be great to include the post-mapped Verilog even if just for debugging/introspection

src/hammer-vlsi/synthesis/vivado/__init__.py
grebe: Generally speaking, is hammer supposed to work without sourcing tool-specific scripts? I.e., should I have to source vivado's settings64.sh, or by having the correct defaults.yml should hammer just work?

edwardcwang: It's desirable since it turns to be possible to waste a lot of time in digging through `source.sh`-esque scripts and it's nice to encode exactly what in the world we need to run a tool (encodes information & reduces messiness). (Of course you could always pre-source stuff and run hammer so we don't remove any functionality, but it's not ideal for re-use.)

grebe: OK, I'm not sure how to do this but it seems reasonable

edwardcwang:
First I'll try to fix the CI bot then will sketch something for the setup parameters and test it. Does that sound reasonable?

edwardcwang:
@grebe It looks like Eric got around to adding defaults.yml and testing, so I'll go ahead and merge this one; let's continue the discussion in #197?

PR #197, opened 2018-09-19
Eric vivado fixes

pkerichang:
Cleanup synthesis/P&R with vivado.

Main changes:

1. Use pkg_resources + jinja2 template to generate tcl and run scripts (If you're not fine with this new dependency, let me know).

2. added defaults.yml file which is used to configure tcl scripts.


src/hammer-vlsi/synthesis/vivado/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+    @property
+    def env_vars(self) -> Dict[str, str]:
+        new_dict = deepdict(super().env_vars)
+        return new_dict
+
+    @property
+    def steps(self) -> List[HammerToolStep]:
+        return self.make_steps_from_methods([
+            self.run_synthesis,
+            self.run_place_and_route,
```
edwardcwang: I think we may want to do these in separate classes so that we can run synthesis and par separately, particularly for a large design. I'd envision putting common stuff into a `VivadoCommon` class and having separate `VivadoSynthesis` and `VivadoPlaceAndRoute` classes.

src/hammer-vlsi/synthesis/vivado/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+        ])
+
+    def _get_dcp_path(self):
```
edwardcwang: We can just use `def get_dcp_path` here methinks. Also in general we've been trying to add type annotations for all functions, including return types.

src/hammer-vlsi/synthesis/vivado/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+        return os.path.join(self.run_dir, _synth_tag, 'obj', 'post_opt.dcp')
+    
+    def setup_workspace(self, tag_name, add_files, run_script, **kwargs) -> bool:
```
edwardcwang: Arguments missing type annotations

src/hammer-vlsi/synthesis/vivado/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+        run_script = 'run-par'
+        add_files = ['synth_dcp_path.tcl', 'par.tcl', 'mcs.tcl']
+        work_dir = self.setup_workspace('vivado_par', add_files, run_script,
+                                        dcp_path=self._get_dcp_path())
+        self.run_executable([os.path.join(work_dir, run_script)])
+        return True
+    
+    def fill_outputs(self) -> bool:
+        # This tool doesn't really have outputs
+        self.output_files = [self._get_dcp_path()]
```
edwardcwang: It would be good to export the post-synthesis netlist here as well for debugging purposes and being consistent with the rest of the API.

src/hammer-vlsi/synthesis/vivado/defaults.yml
```diff
@@ -0,0 +1,20 @@
+# Default settings for synthesis in Vivado, for project/technology configuration and overriding.
+synthesis.vivado:
+  # Location of the binary.
+  binary: "vivado"
+  # Location of the vivado setup script
+  setup_script: "/tools/projects/erichang/hammer-examples/vivado_setup.sh"
```
edwardcwang: Seems like we should check some stuff in?

src/hammer-vlsi/synthesis/vivado/file_templates/board.tcl
```diff
@@ -0,0 +1,5 @@
+# Vivado fragment to define the FPGA board parameters.
+
+set name {{'{'}}{{ board_name }}{{'}'}}
```
edwardcwang: I think it makes sense to use jinja2 if the templates we're subsituting are very complex. However in our case if it's just simple substitutions, I think we may be able to just get away with just regular Python string substitution.

Also I would say for fragments this simple, we may as well just embed them into the Python source code.

src/hammer-vlsi/synthesis/vivado/file_templates/paths.tcl
```diff
@@ -0,0 +1,10 @@
+# This file is dynamically-generated; do not edit this file.
+# TCL fragment to set some paths.
+
+set_param board.repoPaths [list {{ board_files }}];
```
edwardcwang: Ditto above

edwardcwang:
@pkerichang Thanks for the fixes; I think it looks great overall! Left review comments inline above.

Two major things:

1) Can you rebase this against the new master?

2) If possible, can we try to generate as much of the TCL inside the Python as possible? The legacy plugin model was to have a set of templates and have Hammer substitute strings (i.e. what the DC plugin does), but I think we'd like to move towards more of what the Genus/Innovus plugins do, which is generate all components in the Python, so we don't have these large templates with code that isn't generated from an obvious source).

Again overall this looks great and thanks so much for helping to maintain this plugin!

src/hammer-vlsi/synthesis/vivado/__init__.py
pkerichang: That is fine.  However, for testing purpose, supposed I finished synthesis, how would I make hammer start place and route on command line?

src/hammer-vlsi/synthesis/vivado/file_templates/board.tcl
pkerichang: So you prefer to have them embedded in source code instead of having separate template files?

Just asking because for BAG I went the other direction (preferring files + pkg_resources instead of embedding in source) as I found it to be editor friendly; you can open it in an editor and the file format gets recognized, you have syntax highlighting, completions, auto indent, etc., and it's a lot more readable.

I'm fine with either approach, just let me know your decision.

pkerichang:
I'll work on the changes you comments.  However, on tcl script generation, I looked at the way genus generates its tcl script, and as I responded in the inline comments I personally prefer the template file + jinja2 approach as I think it is more readable (jinja2 is more necessary for genus because there's some logic in tcl file generation that can be embedded, but in vivado's case I agree we can get away with a simpler approach).  But if you insist I can change to the other way too.

Also please let me know which hammer command I should run to start P&R for testing purposes.  Thanks

src/hammer-vlsi/synthesis/vivado/__init__.py
edwardcwang: Assuming we have the output of synthesis as `syn-output.json`, we can run par as follows:

```
hammer-vlsi synthesis_to_par -p syn-output.json -o par-run.json
hammer-vlsi par -p par-run.json
```

For a small enough design, we can also try running both with `hammer-vlsi syn-par [...]`.

src/hammer-vlsi/synthesis/vivado/__init__.py
```diff
@@ -1,33 +1,157 @@
...
+        os.symlink(source, dest)
+    except OSError as e:
+        if e.errno == errno.EEXIST:
+            os.remove(dest)
+            os.symlink(source, dest)
+        else:
+            raise e
+
+
+class VivadoCommon(HammerSynthesisTool):
```
edwardcwang: `VivadoCommon` probably doesn't need to inherit from synthesis tool

src/hammer-vlsi/synthesis/vivado/__init__.py
```diff
@@ -1,33 +1,157 @@
...
+
+    _temp_dir = 'file_templates'
+    _base_files = ['board.tcl', 'paths.tcl', 'project.tcl']
+    synth_tag = 'vivado_synth'
```
edwardcwang: I think we can probably inline some/all of these unless there's extensive references to it?

In particular `synth_tag` is redundant since Hammer will autopopulate `self.run_dir` to be (by default) either "obj/syn-rundir" or "obj/par-rundir".

src/hammer-vlsi/synthesis/vivado/__init__.py
```diff
@@ -1,33 +1,157 @@
...
         new_dict = deepdict(super().env_vars)
-        new_dict.update({})  # TODO: stuffs
         return new_dict
 
+    def get_dcp_path(self) -> str:
+        return os.path.join(self.run_dir, self.synth_tag, 'obj',
+                            'post_opt.dcp')
+
+    def setup_workspace(self, tag_name: str, add_files: List[str],
+                        run_script: str, **kwargs: Any) -> bool:
```
edwardcwang: Can we eliminate this `kwargs` unless absolutely necessary?

src/hammer-vlsi/synthesis/vivado/__init__.py
pkerichang: Do I need to do something special with the "tools" variable in __init__.py so hammer is smart enough to know it should create VivadoPlaceAndRoute?

src/hammer-vlsi/synthesis/vivado/__init__.py
```diff
@@ -1,33 +1,157 @@
...
+    def run_place_and_route(self) -> bool:
+        run_script = 'run-par'
+        add_files = ['synth_dcp_path.tcl', 'par.tcl', 'mcs.tcl']
+        work_dir = self.setup_workspace(
+            'vivado_par', add_files, run_script, dcp_path=self.get_dcp_path())
+        self.run_executable([os.path.join(work_dir, run_script)])
+        return True
+
+
+class VivadoSynth(VivadoCommon):
```
edwardcwang: This should inherit both VivadoCommon and HammerSynthesisTool

src/hammer-vlsi/synthesis/vivado/__init__.py
```diff
@@ -1,33 +1,157 @@
...
+        dcp_path = self.get_dcp_path()
+        dcp_dir = os.path.dirname(dcp_path)
+        self.output_files = [
+            dcp_path,
+            os.path.join(dcp_dir, self.top_module + '_post_synth.v')
+        ]
+        return True
+
+
+class VivadoPlaceAndRoute(VivadoCommon):
```
edwardcwang: This should inherit VivadoPlaceAndRoute tool as well

src/hammer-vlsi/synthesis/vivado/file_templates/vivado_setup.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+# sample vivado setup script.
+
+# setup license server
+source /tools/flexlm/flexlm.sh
+# setup vivado environment variables
+source /tools/xilinx/Vivado/2018.1/settings64.sh
```
edwardcwang: The setup script seems obsolete since we have synthesis.vivado.binary. We should also specify as much as possible the license variables in the defaults.yml to avoid over reliance on CAD shell scripts

pkerichang: Do you mean you want to parse through those scripts (which ultimately just sets various environment variables), and list all those variables in defaults.yml instead?  Or something else?  I don't think having synthesis.vivado.binary is sufficient, as Vivado may need some environment variables to be set by the shell script to function.

In general, I find the approach of asking user to provide a "master shell script" which sets up everything needed by the tool (including licenses) to be most portable across different setups.  As a very contrived example, if someone in Xilinx is trying to use Hammer with a developer version of Vivado, he may not need to set up licenses at all.

What do you think?

src/hammer-vlsi/synthesis/vivado/file_templates/board.tcl
grebe: I like having separate template files for the reasons you just mentioned.

pkerichang:
Hello,

I pushed another comment, which basically just partition the synthesis/place-and-route flow properly this time, with some minor edits you suggested.  I think the two main issues left is template files v.s. python source embedding, and whether we want to source a shell script (defined by defaults.yml) that sets up environment, or extract those environment variables into defaults.yml.

Eric

src/hammer-vlsi/synthesis/vivado/core.py
```diff
@@ -0,0 +1,80 @@
+from typing import Dict, List
+
+import os
+import errno
+import pkg_resources
+from itertools import chain
+
+from jinja2 import Template
+
+from hammer_vlsi import deepdict
```
edwardcwang: Would import from hammer_utils I think

src/hammer-vlsi/synthesis/vivado/core.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class VivadoCommon(object):
```
edwardcwang: No need to explicitly inherit object

src/hammer-vlsi/synthesis/vivado/core.py
```diff
@@ -0,0 +1,80 @@
...
+            if e.errno == errno.EEXIST:
+                os.remove(dest)
+                os.symlink(source, dest)
+            else:
+                raise e
+
+    def setup_workspace(self, add_files: List[str], run_script: str,
+                        extra_params: Dict[str, str]) -> None:
+        # make working directory
+        os.makedirs(self.run_dir, exist_ok=True)
```
edwardcwang: The hammer framework will do this automatically methinks.

src/hammer-vlsi/synthesis/vivado/core.py
```diff
@@ -0,0 +1,80 @@
...
+        os.makedirs(os.path.join(self.run_dir, 'obj'), exist_ok=True)
+        # create symlink to constraint file
+        cons_fname = os.path.abspath(
+            self.get_setting('synthesis.vivado.constraints_file'))
+        cons_dir = os.path.join(self.run_dir, 'constrs')
+        os.makedirs(cons_dir, exist_ok=True)
+        cons_targ = os.path.join(cons_dir, os.path.basename(cons_fname))
+        self.sym_link_force(cons_fname, cons_targ)
```
edwardcwang: I think it the constraints file isn't going to be that big, we may as well copy it to be deterministic.

src/hammer-vlsi/synthesis/vivado/defaults.yml
edwardcwang: I think hammer has a way to reference files relative to the tool dir (please poke me about this)

src/hammer-vlsi/synthesis/vivado/defaults.yml
```diff
@@ -0,0 +1,22 @@
...
+  # Location of the constraints file
+  constraints_file: "/tools/projects/erichang/hammer-examples/vc707.xdc"
```
edwardcwang: Ditto above

src/hammer-vlsi/synthesis/vivado/par.py
```diff
@@ -0,0 +1,34 @@
+from typing import List
+
+import os
+
+from hammer_vlsi import HammerPlaceAndRouteTool, HammerToolStep
+
+from .core import VivadoCommon
```
edwardcwang: FYI without `__init__.py` the default HammerDriver will get confused when it tries to import this plugin.

We probably need to move `core.py` to something like `src/hammer-vlsi/common/vivado_core.py` and move synth.py and par.py to `src/hammer-vlsi/synthesis/vivado/__init__.py` and `src/hammer-vlsi/par/vivado/__init__.py` respectively.

src/hammer-vlsi/synthesis/vivado/file_templates/board.tcl
edwardcwang: Long-form thoughts below below

src/hammer-vlsi/synthesis/vivado/file_templates/vivado_setup.sh
edwardcwang: Long-form thoughts below

src/hammer-vlsi/synthesis/vivado/defaults.yml
edwardcwang: Yes I agree in general they need not (and probably aren't) in the tool directory. It would be nice to have it work "out of the box" in some way and hardcoding BWRC-specific paths in here might not be the best idea.

I'd envision that these could point to the default scripts and generators that use Hammer can customize/override them as they wish.

src/hammer-vlsi/synthesis/vivado/par.py
pkerichang: The problem I have is that the current hammer source directory structure is really weird; it seems like you have many disjoints python modules/packages lying all over the place, with no clear hierarchy between them.

For example, src/hammer-vlsi/common is not a python package (no __init__.py), so is the synthesis folder and par folder, so organization wise they are just three disjoint folders that each contains some python stuffs.  with the suggestion you have then the synthesis/place-and-route files need to magically know the existence and path to the vivado_core.py file, and subclass a class defined there.  I'm not so comfortable with that approach.

My current solution now is to have a vivado python package living in the synthesis folder (although admittedly that's not a good place to be in anymore), with a core module, synth module, and par module, so the classes are tightly coupled together.  To import the synthesis plugin, you set the synthesis path to "src/hammer-vlsi/synthesis", and the tool name to "vivado.synth".  To import the P&R plugin, you use the same tool path, but change the name to "vivado.par".  importlib is able to resolve it properly.

Eric

edwardcwang: Yeah I agree that plugin common files don't have a good structure and just filed #198 to deal with this. I do think in the meantime it would be nice if we could follow the convention for importing hammer plugins so ideally we would put the par portion of vivado to `src/hammer-vlsi/par/vivado`.

If it's difficult I can work on #198 first if you'd like?

edwardcwang:
Also as long as the main structure is not too disturbed I'm happy to just merge this PR when it's in a decent shape, and merge any future improvements in subsequent PRs, btw.

pkerichang:
Let me just put all my responses here (going to sleep soon so I'll work on these tomorrow):

1. Regarding par/synth file location: I agree with you, if there's a better "common" location that would help a lot.  Vivado is the corner case where the same tool does both synthesis and P&R and I understand why the current structure doesn't handle it in a nice way.  I think it'll probably be cleaner to keep the current structure and accept this pull request, then after you finish #198 I can help you move files to the right place.

2. Regarding template files: I see your concern and I agree, I can refactor the code to do that.  Does that mean I should try to remove the jinja2 dependency before accepting this pull request then?

3. For environment variables, part of it is just I'm too busy to read through the vivado setup scripts and track down everything it's doing, but like you say I think the hammer structure can support both approach (either set environment variables or source setup script if that's what the user want).  Maybe accept the current pull request and we can slow add in the envrionment variables as improvement in the future?

I think other issues are minor enough that I can fix them tomorrow morning.

edwardcwang:
@pkerichang 
1) I think we are on the same page generally. I think merging is fine, but I am a bit concerned about having to reference "vivado.synth" and "vivado.par" as tools in the IR - to conform to the existing style it's best if we can reference e.g. `vlsi.core.synthesis_tool: vivado` and `vlsi.core.par_tool`. I think unfortunately the easiest way to do that the moment is to move them to the appropriate places under `src/hammer-vlsi/{synthesis,par}`.
Would it be okay if I moved them for now and then when we have the proper fix in #198 we'll re-do it properly then?

Alternatively if you know of a way to make `vlsi.core.synthesis_tool: vivado` work then I'm open to that as well.

2) Thanks for taking care of this. Also since we still use `pkg_resources`, this will mean that this plugin will depend on the setuptools. https://stackoverflow.com/questions/22148144/python-importerror-no-module-named-pkg-resources

Do we only use it for finding files?

3) Agreed, let's do it later in #199.

edwardcwang:
I think if we can take care of the last 2 points above and address the detailed points in the code review we should be good to go.

I still have some minor itches but will probably just take care of them later on :)

pkerichang:
1. Please feel free to move them, then we can accept this pull request (and then I can use this setup in EE290C).

2. pkg_resources is the only solution I know of that will work for both directory structure and egg zip files, which will be the case if people want to install hammer through pip in the future.  In my opinion if they have Python 3 they should have setuptools too, as that's the preferred package.  But if this is not something you think we'll care about I can remove the setuptools dependency.

edwardcwang:
Resolution:

Eric to remove direct pkg_resources dependency (filed #200 for using pkg_resources under the hood in hammer infrastructure)

Edward to move files to proper place and use hammer's relative file location functionality.

pkerichang:
Seems like the P&R code is failing because it cannot find VivadoCommon?  Should we just remove the P&R code for now and add back later when a good place for VivadoCommon is found?

edwardcwang:
Yes, the P&R code is failing because I didn't finish the migration. I will take care of it today.

edwardcwang:
@pkerichang I fixed as much as I could - the error above looks like an error in your code?

Also you may want to review the changes I did to make sure I didn't make a mistake.

pkerichang:
The bug is fixed and the code still works, feel free to merge whenever you can.

edwardcwang:
Merged, thanks!

PR #205, opened 2018-09-30
Add two tool apis. is_newer_version and tool_config_prefix

colinschmidt:
One enables checking for specific versions of tools
The other gives a pointer to the tool specific configs

### DIFF
--- a/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
+++ b/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
@@ -95,6 +95,23 @@ class HammerTool(metaclass=ABCMeta):
         """
         return {}
 
+    def tool_config_prefix(self) -> str:
+        """
+        Returns the config string that contains all tool specific settings
+
+        :return: A string that is the prefix for all tool specific settings
+        """
+        return ""
+
+    def is_newer_version(self, test_version: str) -> bool:
+        """
+        Based on the tool figures out if the test_version is at least as new as the current version
+
+        :param test_version: The version to compare against the in-use version
+        :return: True if the test_version is at least as new as the current version
+        """
+        return True
+
     # Setup functions.
     def run(self, hook_actions: List[HammerToolHookAction] = []) -> bool:
         """Run this tool.
--- a/src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
+++ b/src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
@@ -397,6 +397,28 @@ class CadenceTool(HasSDCSupport, HammerTool):
 
         return reduce(update_dict, list_of_vars + [cadence_vars], {})
 
+    # Assumes versions look like MAJOR_ISRMINOR
+    def is_newer_version(self, test_version: str) -> bool:
+        current_version = self.get_setting(self.tool_config_prefix() + ".version") # type: str
+        current_main_version = int(current_version.split("_")[0]) # type: int
+        test_main_version = int(test_version.split("_")[0]) # type: int
+        if(current_main_version > test_main_version):
+            return True
+        if(current_main_version < test_main_version):
+            return False
+        if("_" in current_version):
+            if("_" in test_version):
+                current_minor_version = int(current_version.split("_")[1][3:]) # type: int
+                test_minor_version = int(test_version.split("_")[1][3:]) # type: int
+                if(current_minor_version >= test_minor_version):
+                    return True
+                else:
+                    return False
+            else:
+                return False
+        else: # If there is no minor version they are at least equal
+            return True
+
     def get_timing_libs(self, corner: Optional[MMMCCorner] = None) -> str:
         """
         Helper function to get the list of ASCII timing .lib files in space separated format.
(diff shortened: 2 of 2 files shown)


src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -95,6 +95,23 @@ def export_config_outputs(self) -> Dict[str, Any]:
         """
         return {}
 
+    def tool_config_prefix(self) -> str:
+        """
+        Returns the config string that contains all tool specific settings
+
+        :return: A string that is the prefix for all tool specific settings
+        """
+        return ""
```
edwardcwang: Maybe we can make this method abstract?

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -95,6 +95,23 @@ def export_config_outputs(self) -> Dict[str, Any]:
...
+
+    def is_newer_version(self, test_version: str) -> bool:
```
edwardcwang: Hmm, what if instead we had an API that converts the string version into a comparable integer instead? That way maybe tools can do something like 

```python
if version >= self.versionInt("10.0") and version < self.versionInt("13.0"):
  ...
```

Alternatively maybe something like `def versionCompare(self, version_1: str, version_2: str) -> int` where it returns either -1, 0, or 1. What do you think?

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
colinschmidt: Sure. I wasn't exactly sure what that annotation did.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
colinschmidt: Sure. I like the first one better. I thought about the second one but it didn't seem needed.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
edwardcwang: Sounds good. Also if we make it abstract we can just do `pass` instead of `return ""` in the body.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
edwardcwang: The second one is needed if we want to sort a list of version strings though it does seem like a unlikely task. Also we can always implement it using the first so it's more redundant.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -592,17 +602,26 @@ def env_vars(self) -> Dict[str, str]:
...
 
         :param product: Either "DC" or "ICC"
         :param settings_key: Key to retrieve the version for the product. Leave blank for DC and ICC.
         """
-        key = settings_key # type: str
-        if product == "DC":
-            key = "synthesis.dc.dc_version"
-        elif product == "ICC":
-            key = "par.icc.icc_version"
+        key = self.tool_config_prefix() + "." + "version" # type: str
```
edwardcwang: Is `product` obsolete? If so we should remove the parameter.

src/hammer-vlsi/par/nop.py
```diff
@@ -16,6 +16,12 @@ def fill_outputs(self) -> bool:
         self.output_ilms = []
         return True
 
+    def tool_config_prefix(self):
+        return "empty"
+
+    def version_number(self, version:str):
+        return 1
+
```
edwardcwang: The above are missing return type annotations and needs a space between `version` and `:str`.

Perhaps we should return a blank string or maybe change the return type to `Optional[str]` for `tool_config_prefix`?

src/hammer-vlsi/synthesis/mocksynth/__init__.py
```diff
@@ -21,6 +21,12 @@ def env_vars(self) -> Dict[str, str]:
         new_dict.update({})  # TODO: stuffs
         return new_dict
 
+    def tool_config_prefix(self):
+        return "empty"
+
+    def version_number(self, version:str):
+        return 1
+
```
edwardcwang: Ditto above

src/hammer-vlsi/test.py
```diff
@@ -194,6 +194,12 @@ def step(self) -> bool:
         """
         pass
 
+    def tool_config_prefix(self):
+        return "empty"
+
+    def version_number(self, version:str):
+        return 1
+
```
edwardcwang: Maybe all the test tools should mix in this instead of copying over and over again?

src/hammer-vlsi/test.py
```diff
@@ -464,6 +470,12 @@ def steps(self) -> List[hammer_vlsi.HammerToolStep]:
                     self.step
                 ])
 
+            def tool_config_prefix(self):
+                return "empty"
+
+            def version_number(self, version:str):
+                return 1
+
```
edwardcwang: Ditto above

edwardcwang:
I think this looks much better now - just added a new round of comments.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
colinschmidt: No we still use product to come up with the tarball name a few lines below.

src/hammer-vlsi/par/nop.py
colinschmidt: I think an empty string is fine.
It shouldn't really be optional, its just that mock/dummy tools still need to implement abstract methods.
I'm gonna try to create a single DummyTool that implements the abstract methods for all of these dummy tools to use.

edwardcwang: Now that I think about it a blank string might have issues if parts of the Hammer core are just adding ".<something>" but ".whatever" isn't going to be a valid key...

colinschmidt: I think if a dummy tool requires/uses config keys then it will need to override that method.

PR #208, opened 2018-10-04
Add the ability for tech plugins to use yaml

colinschmidt:
This enables people to write comments in the tech plugins

### DIFF
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -213,23 +213,24 @@ class HammerTechnology:
         self.config = None  # type: TechJSON
 
     @classmethod
-    def load_from_dir(cls, technology_name: str, path: str, is_yaml: bool = False) -> "HammerTechnology":
-        """Load a technology from a given folder.
+    def load_from_dir(cls, technology_name: str, paths: List[str]) -> Optional["HammerTechnology"]:
+        """Load a technology from a given set of folders.
 
         :param technology_name: Technology name (e.g. "saed32")
-        :param path: Path to the technology folder (e.g. foo/bar/technology/saed32)
-        :param is_yaml: Whether the technology information is stored in yaml or not. If false hammer will assume json.
+        :param paths: A list of paths to the technology folders (e.g. foo/bar/technology/saed32)
         """
-
-        if is_yaml:
-            with open(os.path.join(path, "{technology_name}.tech.yaml".format(technology_name=technology_name))) as f:
-                yaml_str = f.read()
-                return HammerTechnology.load_from_yaml(technology_name, yaml_str, path)
-        else:
-            with open(os.path.join(path, "{technology_name}.tech.json".format(technology_name=technology_name))) as f:
-                json_str = f.read()
-                return HammerTechnology.load_from_json(technology_name, json_str, path)
-
+        for path in paths:
+            tech_blob_path = os.path.join(path, technology_name, "%s.tech.json" % technology_name)
+            if os.path.exists(tech_blob_path):
+                with open(tech_blob_path) as f:
+                    json_str = f.read()
+                    return HammerTechnology.load_from_json(technology_name, json_str, path)
+            tech_blob_path = os.path.join(path, technology_name, "%s.tech.yaml" % technology_name)
+            if os.path.exists(tech_blob_path):
+                with open(tech_blob_path) as f:
+                    yaml_str = f.read()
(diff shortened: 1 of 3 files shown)


src/hammer-vlsi/hammer_vlsi/driver.py
```diff
@@ -121,24 +121,30 @@ def update_project_configs(self, project_configs: List[dict]) -> None:
...
-            if os.path.exists(tech_json_path):
+            tech_blob_path = os.path.join(path, tech_str, "%s.tech.json" % tech_str)
+            if os.path.exists(tech_blob_path):
+                is_yaml = False
                 break
-        if tech_json_path == "":
-            self.log.error("Technology {0} not found or missing .tech.json!".format(tech_str))
+            tech_blob_path = os.path.join(path, tech_str, "%s.tech.yaml" % tech_str)
+            if os.path.exists(tech_blob_path):
+                is_yaml = True
```
edwardcwang: Generally looks good - I would move the `is_yaml` detection logic inside `load_from_dir` though. (e.g. when we load `defaults.yml/json` the exact one is determined inside the load plugin logic)

colinschmidt:
I had to make a few changes to the load_dir API, so this deserves another look but I think that API is mostly internal.
It now directly takes in the list of tech paths, rather than performing that selection outside in load_tech.
I also uncovered a strange assumption where load_technology prefixed the tech json/yaml with another directory named after the technology but load_from_dir did not. Unifying this meant a few small tweaks to the unittests, but since load_from_dir is an internal API this should be fine.

src/hammer-vlsi/test.py
```diff
@@ -276,8 +276,11 @@ def test_yaml_tech_file(self) -> None:
         tech_yaml_filename = tech_dir + "/dummy28/dummy28.tech.yaml"
         with open(tech_yaml_filename, "w") as f:
             f.write(tech_yaml)
-        tech = hammer_tech.HammerTechnology.load_from_dir("dummy28", [tech_dir])
-        self.assertTrue(tech is not None, "Unable to load technology")
+        tech_opt = hammer_tech.HammerTechnology.load_from_dir("dummy28", [tech_dir])
+        if tech_opt is None:
+            self.assertTrue(tech_opt is not None, "Unable to load technology")
```
edwardcwang: Seems like this became an impossible assertion...?

colinschmidt: Oops. I had an refactor for the type checker that made that useless.

colinschmidt: It's actually an assertion that always fires which is correct(the test should fail if we can't load the technology).
Its a bit unclear in this case so I'm changing it.

colinschmidt:
Trying to use this with a technology plugin and getting something weird. Lets hold off on merging until I resolve what (if any) changes need to be made there.

colinschmidt:
Ok I fixed the issue I was seeing. I've now done a real test with this.

src/hammer-tech/hammer_tech.py
```diff
@@ -212,17 +213,24 @@ def __init__(self):
...
         :param technology_name: Technology name (e.g. "saed32")
-        :param path: Path to the technology folder (e.g. foo/bar/technology/saed32)
+        :param paths: A list of paths to the technology folders (e.g. foo/bar/technology/saed32)
         """
-
-        with open(os.path.join(path, "{technology_name}.tech.json".format(technology_name=technology_name))) as f:
-            json_str = f.read()
-
-        return HammerTechnology.load_from_json(technology_name, json_str, path)
+        for path in paths:
```
edwardcwang: I'm not sure I would put this `paths` for loop in here. It might be related to the unit test failures. Can take a look at some point.

colinschmidt: This passed before its probably something silly.

edwardcwang: Yeah I think some technology prefixing is inconsistent somewhere. I do think `load_from_dir` makes slightly more sense if it loads a single dir though, and instead we let the driver load from multiple dirs (this is consistent with tool plugin loading)

colinschmidt: Yeah I agree. My original load_from_dir change was keeping the name accurate, but in order to do the yaml/json detection in there I had to move the paths loop, or duplicate code/work.

colinschmidt:
Thanks for the fix ups looks good to me.

PR #210, opened 2018-10-04
Typo mili->milli

jwright6323:
Only changing a comment...


colinschmidt:
we should have support for mils 🤣

jwright6323:
📏 🔥

PR #216, opened 2018-10-10
Driver and CLIDriver refactor

### LINKED ISSUE #183: Shell utility to migrate changes from .yml into a par-input.json without rerunning synthesis
This would be super useful since I do it manually all the time. Seems easy to do superficially but risk is that we miss edge cases; maybe restrict only to `vlsi.*` things and exclude control settings (e.g. `input_files`) for now?

edwardcwang:
The main breaking change:
1) Outputs are no longer full outputs by defaults
2) -full outputs are dumped by default for syn and par actions

syn-par's behaviour should be preserved.

This should enable #183 using syn-to-par now.
Previously there was no way to get the output-only parts of the config and combine it with new input configs. This is a major pain point since many parts of the Hammer IR are consumed only in par and it's a pain to manually translate the input YAML to `par-input.json`.

The rest is mostly infrastructural work that increases DRY and reduces boilerplate.

Close #183


src/hammer-vlsi/cli_driver_test.py
```diff
@@ -39,6 +42,199 @@ def generate_dummy_config(self, syn_rundir: str, config_path: str, top_module: s
...
+                                   syn_out_path: str,
+                                   syn_to_par_out_path: str) -> None:
+        # Check that running the CLIDriver executes successfully (code 0).
+        with self.assertRaises(SystemExit) as cm:  # type: ignore
+            CLIDriver().main(args=[
+                "syn",  # action
+                "-p", config_path,
+                "--output", syn_out_path,
+                "--syn_rundir", syn_rundir,
+                "--par_rundir", par_rundir
```
colinschmidt: Is there a reason to give it this extra path? Are you checking that it doesn't use it?

src/hammer-vlsi/cli_driver_test.py
```diff
@@ -39,6 +42,199 @@ def generate_dummy_config(self, syn_rundir: str, config_path: str, top_module: s
...
+        syn_to_par_out_path = os.path.join(syn_rundir, "syn_par_out.json")
+        self.generate_dummy_config(syn_rundir, config_path, top_module)
+
+        self.run_syn_to_par_with_output(config_path, syn_rundir, par_rundir,
+                                        syn_out_path, syn_to_par_out_path)
+
+        # synthesis output should NOT keep other settings
+        with open(syn_out_path, "r") as f:
+            syn_output = json.loads(f.read())
+            self.assertEqual(syn_output["synthesis.outputs.output_files"], [])
```
colinschmidt: Isn't this one of the few things that should be in the syn_output?

src/hammer-vlsi/cli_driver_test.py
```diff
@@ -39,6 +42,199 @@ def generate_dummy_config(self, syn_rundir: str, config_path: str, top_module: s
...
+            self.assertFalse("vlsi.core.technology" in syn_output)
+        # Generated par input should have other settings
+        with open(syn_to_par_out_path, "r") as f:
+            par_input = json.loads(f.read())
+            self.assertEqual(par_input["par.inputs.top_module"], top_module)
+            # par-input should preserve other settings
+            self.assertEqual(par_input["vlsi.core.technology"], "nop")
+
+        # Cleanup
+        shutil.rmtree(syn_rundir)
```
colinschmidt: Do these runs also create output.json in current directory still? If so should it be deleted?

src/hammer-vlsi/cli_driver_test.py
```diff
@@ -39,6 +42,199 @@ def generate_dummy_config(self, syn_rundir: str, config_path: str, top_module: s
...
+        top_module = "dummy"
+        config_path = os.path.join(syn_rundir, "run_config.json")
+        syn_out_full_path = os.path.join(syn_rundir, "syn-output-full.json")
+        syn_to_par_out_path = os.path.join(syn_rundir, "syn_par_out.json")
+        self.generate_dummy_config(syn_rundir, config_path, top_module)
+
+        # Check that running the CLIDriver executes successfully (code 0).
+        with self.assertRaises(SystemExit) as cm:  # type: ignore
+            CLIDriver().main(args=[
+                "syn", # action
```
colinschmidt: Without the output flag it creates the full json config file in `syn-output-full.json`? Or is this a typo in the settings?

src/hammer-vlsi/cli_driver_test.py
```diff
@@ -39,6 +42,199 @@ def generate_dummy_config(self, syn_rundir: str, config_path: str, top_module: s
...
+                "--output", syn_to_par_out_path,
+                "--syn_rundir", syn_rundir,
+                "--par_rundir", par_rundir
+            ])
+        self.assertEqual(cm.exception.code, 0)
+
+        # synthesis full output should keep other settings
+        with open(syn_out_full_path, "r") as f:
+            syn_output = json.loads(f.read())
+            self.assertEqual(syn_output["synthesis.outputs.output_files"], [])
```
colinschmidt: Again this seems reversed.

src/hammer-vlsi/cli_driver_test.py
```diff
@@ -39,6 +42,199 @@ def generate_dummy_config(self, syn_rundir: str, config_path: str, top_module: s
...
+        # Check that the syn-to-par generated par input and the syn-par
+        # generated par input are the same, modulo any synthesis.outputs.*
+        # settings since they don't matter for par input.
+        with open(syn_to_par_out_path, "r") as f:
+            with open(os.path.join(syn_rundir, "par-input.json"), "r") as f2:
+                syn_to_par_output = json.loads(f.read())
+                syn_to_par_output = dict(filter(lambda i: "synthesis.outputs." not in i[0], syn_to_par_output.items()))
+                syn_par_output = json.loads(f2.read())
+                syn_par_output = dict(filter(lambda i: "synthesis.outputs." not in i[0], syn_par_output.items()))
+                self.assertEqual(syn_to_par_output, syn_par_output)
```
colinschmidt: What is expected to differ between the non-filtered versions? Do we even want the filter?

src/hammer-vlsi/cli_driver_test.py
```diff
@@ -68,14 +264,63 @@ def test_syn_par_config_dumping(self) -> None:
         with open(os.path.join(syn_rundir, "syn-output.json"), "r") as f:
             syn_output = json.loads(f.read())
             self.assertEqual(syn_output["synthesis.outputs.output_files"], [])
+            # syn-output should NOT keep other settings
+            self.assertFalse("vlsi.core.technology" in syn_output)
+        with open(os.path.join(syn_rundir, "syn-output-full.json"), "r") as f:
+            syn_output_full = json.loads(f.read())
+            self.assertEqual(syn_output_full["synthesis.outputs.output_files"], [])
```
colinschmidt: As above.

src/hammer-vlsi/cli_driver_test.py
edwardcwang: I think in this specific case there should be no use of par_rundir but I generally leave it in there for safety purposes. Do you want me to remove it?

src/hammer-vlsi/cli_driver_test.py
edwardcwang: It is in syn_output. The comment for "other settings" refers to settings like `vlsi.core.technology`.

src/hammer-vlsi/cli_driver_test.py
edwardcwang: Responded above

src/hammer-vlsi/cli_driver_test.py
edwardcwang: See the line below ("other setting" like `vlsi.core.technology` is kept)

src/hammer-vlsi/cli_driver_test.py
edwardcwang: Yes the synthesis tool driver creates a full output json in `syn-rundir/syn-output-full.json` by default.

src/hammer-vlsi/cli_driver_test.py
edwardcwang: The filter simply removes `synthesis.outputs.*` settings. The reason there is a discrepancy is because `syn-par` takes the output-only result from synthesis and combines it with the original (synthesis.output free config), whereas `syn-to-par` combines it with the synthesis output which includes "synthesis.output.*" settings. Of course we could hack/special-case to `syn-to-par` but it seems like it doesn't really matter and avoids an unnecessary hack.

src/hammer-vlsi/hammer_vlsi/cli_driver.py
```diff
@@ -223,12 +269,13 @@ def syn_par_action(driver: HammerDriver, append_error_func: Callable[[str], None
                 return None
             else:
                 # Generate place-and-route input from the synthesis output.
-                par_input = HammerDriver.generate_par_inputs_from_synthesis(syn_output)  # type: dict
+                syn_output_converted = HammerDriver.synthesis_output_to_par_input(syn_output)
+                assert syn_output_converted is not None, "syn_output was generated by CLIDriver"
```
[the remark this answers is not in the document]
edwardcwang: Done

src/hammer-vlsi/hammer_vlsi/driver.py
```diff
@@ -170,27 +247,20 @@ def load_par_tool(self, run_dir: str = "") -> bool:
...
         par_tool.technology = self.tech
         par_tool.set_database(self.database)
         par_tool.run_dir = run_dir
-        par_tool.hierarchical_mode = HierarchicalMode.from_str(self.database.get_setting("vlsi.inputs.hierarchical.mode"))
+        par_tool.hierarchical_mode = HierarchicalMode.from_str(
+            self.database.get_setting("vlsi.inputs.hierarchical.mode"))
 
         # TODO: automate this based on the definitions
         par_tool.input_files = self.database.get_setting("par.inputs.input_files")
         par_tool.top_module = self.database.get_setting("par.inputs.top_module")
```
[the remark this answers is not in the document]
edwardcwang: Done

src/hammer-vlsi/cli_driver_test.py
edwardcwang: Yes. I've filed #218 for fixing this since it's a bit more involved.

src/hammer-vlsi/cli_driver_test.py
colinschmidt: Hmm. I don't think its a big deal, I'll leave it up to you

src/hammer-vlsi/cli_driver_test.py
colinschmidt: Yeah its fine to leave as is. Thanks for the explanation.

Issue #237, opened 2018-10-25
./generate_properties.py should be more automatic


jwright6323:

edwardcwang:
We should also update export_config_outputs() automatically

PR #238, opened 2018-10-25
Add GDS output

jwright6323:
This adds the GDS output as a default interface output


edwardcwang:
@jwright6323 Can you add it to https://github.com/ucb-bar/hammer/blob/0998b543c0993f351d3146aa5fe89eea75e0cd14/src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py#L247 as well? (yes we should probably make it automatically update from generate_properties)

jwright6323:
Done #237 should note this.

John

PR #239, opened 2018-10-25
Ensure LSF jobs are killed when hammer exits

### LINKED ISSUE #236: Kill LSF jobs after ctrl+C
Currently this leaves zombies behind.

colinschmidt:
Fixes #236


jwright6323:
So the solution ended up being to just remove `atexit.register(...)` which seems to effect the desired behavior.

PR #244, opened 2018-10-30
Refactor check_function_type to not raise an exception by default

edwardcwang:
Added assert_function_type


edwardcwang:
Sometimes you want to check the type and know that it failed without it raising an exception.

PR #246, opened 2018-10-31
LEF utilities

edwardcwang:
Also move VerilogUtils into hammer_utils


colinschmidt:
How many different LEFs did you test this with?

edwardcwang:
Just a handful so far.

jwright6323:
We probably want to make `get_sizes` a more generic `parse_lef` method eventually.

src/hammer-vlsi/hammer_utils/lef_utils.py
```diff
@@ -0,0 +1,74 @@
...
+
+            # Check for "END <my_macro>"
+            regex = "END " + re.escape(in_macro)
+            if re.search(regex, line) is not None:
+                # END found
+                in_macro = None
+                continue
+
+            # Check for SIZE
+            regex = "SIZE\s+([\d\.]+)\s+BY\s+([\d\.]+)\s+;"
```
jwright6323: This might be fragile- can the semicolon come directly after the number? This regex requires a space, maybe make the last `\s+` a `\s*`?

edwardcwang:
Agreed for making it more generic eventually, just wanted to be agile and not overdesign.

src/hammer-vlsi/hammer_utils/lef_utils.py
edwardcwang: Yes we can use `\s*`. It doesn't hurt us to be more compatible.

[The LEF reference](http://www.ispd.cc/contests/18/lefdefref.pdf) as well as generated LEFs both have a space before the semicolon.

jwright6323:
What is `PROPERTYDEFINITIONS`?

edwardcwang:
It's a way to specify certain properties not supported directly in the LEF syntax. It seems to appear in certain LEF files in certain technologies. See the LEF reference for more details on that.

jwright6323:
Why does it need to be in the `get_sizes` method, though? Does it interact with `SIZE` somehow?

edwardcwang:
You can have statements within it that have `MACRO` which confuse the current parser:
```
PROPERTYDEFINITIONS
 MACRO LEF58_EDGETYPE STRING ;
END PROPERTYDEFINITIONS
```

PR #251, opened 2018-11-01
Move filtering and tech-centred functions to hammer_tech

edwardcwang:
HammerTool has accumulated these functions but since they primarily only reference the technology, they can move to hammer_tech.

It also enables library users to perform technology-centered action without having to spawn a tool instance.


src/hammer-vlsi/test.py
```diff
@@ -649,6 +656,7 @@ def test_tool_format(lib, filt) -> List[str]:
             ]
         }])
         test.set_database(database)
+        tech.set_database(database)
```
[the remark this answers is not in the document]
edwardcwang: Not quite since the tech plugins now use the database directly, so tech needs its database set. The Driver does this already.

src/hammer-vlsi/test.py
```diff
@@ -456,24 +454,28 @@ def test_read_libs(self) -> None:
...
+                self._test_filter_output = tech.read_libs([HammerToolTestHelpers.make_test_filter()], test_tool_format, must_exist=False)
                 return True
         test = Tool()
         test.logger = HammerVLSILogging.context("")
         test.run_dir = tempfile.mkdtemp()
         test.technology = tech
-        test.set_database(hammer_config.HammerDatabase())
+        database = hammer_config.HammerDatabase()
+        test.set_database(database)
+        tech.set_database(database)
```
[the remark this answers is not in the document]
edwardcwang: See above

src/hammer-vlsi/test.py
```diff
@@ -551,6 +555,7 @@ def step(self) -> bool:
         test.run_dir = tempfile.mkdtemp()
         test.technology = tech
         test.set_database(hammer_config.HammerDatabase())
+        tech.set_database(hammer_config.HammerDatabase())
```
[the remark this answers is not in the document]
edwardcwang: See above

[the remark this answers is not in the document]
jwright6323: Oh, I just can't read. I read both `test` and `tech` as `tech`. LGTM

src/hammer-tech/hammer_tech.py
```diff
@@ -5,19 +5,28 @@
...
 
 import hammer_config
-
-from hammer_logging import HammerVLSILoggingContext
-
 import python_jsonschema_objects  # type: ignore
-from hammer_utils import deeplist, get_or_else, optional_map
 from hammer_config import load_yaml
+from hammer_logging import HammerVLSILoggingContext
+from hammer_utils import (LEFUtils, add_lists, deeplist, get_or_else,
```
colinschmidt: Is there some style guide at which point you import * or is that not possible in python?

src/hammer-tech/hammer_tech.py
```diff
@@ -164,6 +173,37 @@ def library_from_json(json: str) -> Library:
     return Library.from_json(json)
 
 
+# Struct that holds an extra library and possible prefix.
+class ExtraLibrary(NamedTuple('ExtraLibrary', [
+    ('prefix', Optional[PathPrefix]),
+    ('library', Library)
+])):
+    __slots__ = ()
```
colinschmidt: This seems like an over optimization at this point. Did you see some memory usage problem?

src/hammer-tech/hammer_tech.py
```diff
@@ -164,6 +173,37 @@ def library_from_json(json: str) -> Library:
...
+
+    def to_setting(self) -> dict:
+        raise NotImplementedError("No clean implementation for JSON export of library yet")
```
colinschmidt: Is this something we want and should be an issue?

src/hammer-tech/hammer_tech.py
edwardcwang: I usually just put down `__slots__` by habit at this point...

src/hammer-tech/hammer_tech.py
edwardcwang: It's not really important right now. Fixing this is hard since it requires wrangling with/replacing python_jsonschema_objects

src/hammer-tech/hammer_tech.py
edwardcwang: IIRC the general rule is to avoid wildcard imports unless the package is dead simple e.g. has a bunch of constants.

PR #213, opened 2018-10-09
[WIP do not merge] DRC/LVS additions

jwright6323:
I'm just trying to see what the CI says right now. We'll CR this with updates to plugins sometime this week.

John

PR #257, opened 2018-11-03
Auto generate properties

### LINKED ISSUE #237: ./generate_properties.py should be more automatic

jwright6323:
Closes #237 

This is reliant on #213. Running `./generate_properties.py` now directly edits `hammer_vlsi_impl.py`.

### DIFF
--- a/src/hammer-vlsi/generate_properties.py
+++ b/src/hammer-vlsi/generate_properties.py
@@ -6,12 +6,12 @@
 #  Copyright 2017-2018 Edward Wang <edward.c.wang@compdigitec.com>
 #  Helper script to generate properties for hammer-vlsi tool classes.
 
-from collections import namedtuple
-import sys
-
+import argparse
 import os
 import re
-import argparse
+import sys
+from collections import namedtuple
+from typing import Dict
 
 InterfaceVar = namedtuple("InterfaceVar", 'name type desc')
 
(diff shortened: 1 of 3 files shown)


src/hammer-vlsi/generate_properties.py
```diff
@@ -128,16 +143,11 @@ def main(args):
...
-    output.extend(generate_interface(HammerSynthesisTool))
-    output.append("")
-    output.extend(generate_interface(HammerPlaceAndRouteTool))
-    output.append("")
-    output.extend(generate_interface(HammerDRCTool))
-    output.append("")
-    output.extend(generate_interface(HammerLVSTool))
-    print("\n".join(output))
- 
+    generate_interface(HammerSynthesisTool)
```
edwardcwang: Can we maybe make updating the implementation filename a flag, perhaps one that's enabled by default? I think it will help with testing/determinism if we can just print to stdout as well.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -781,6 +782,26 @@ def hcells_list(self, value: List[str]) -> None:
         self.attr_setter("_hcells_list", value)
 
 
+    @property
+    def test_attr(self) -> str:
+        """
+        Get the foobar.
```
edwardcwang: Looks like some test code snuck into the actual implementation

src/hammer-vlsi/generate_properties.py
jwright6323: yes, good idea.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
jwright6323: Oops.

src/hammer-vlsi/generate_properties.py
jwright6323: Fixed

PR #261, opened 2018-11-14
Remove obsolete code

edwardcwang:


colinschmidt:
Its git so we can always revive any of this if we need to but as it hasn't been used by anoyone in years it LGTM.

PR #262, opened 2018-11-16
Technology and library filters refactor

edwardcwang:
Will be used in the macro sizes PR and reduces the amount of spaghetti code needed for it.

### DIFF
--- a/src/hammer-tech/library_filter.py
+++ b/src/hammer-tech/library_filter.py
@@ -6,7 +6,9 @@
 #  Copyright 2017-2018 Edward Wang <edward.c.wang@compdigitec.com>
 
 from numbers import Number
-from typing import TYPE_CHECKING, Any, Callable, List, NamedTuple, Optional, Union
+from typing import TYPE_CHECKING, Any, Callable, List, NamedTuple, Optional, Tuple, Union
+
+from hammer_utils import get_or_else
 
 if TYPE_CHECKING:
     # grumble grumble, we need a better Library class generator
@@ -24,14 +26,22 @@ if TYPE_CHECKING:
         @property
         def verilog_synth(self) -> Optional[str]: pass
 
+PathsFunctionType = Callable[["Library"], List[str]]
+ExtractionFunctionType = Callable[["Library", List[str]], List[str]]
+
 
 class LibraryFilter(NamedTuple('LibraryFilter', [
     ('tag', str),
     ('description', str),
     # Is the resulting string intended to be a file?
     ('is_file', bool),
-    # Function to extract desired string(s) out of the library.
-    ('extraction_func', Callable[["Library"], List[str]]),
+    # Function to extract desired path(s) out of the library.
+    # Returns either a list of library-relative paths.
+    ('paths_func', PathsFunctionType),
+    # Function to extract desired string(s) out of the library, given full
+    # paths and the Library.
+    # Returns a list of strings.
+    ('extraction_func', Optional[ExtractionFunctionType]),
     # Additional filter function to use to exclude possible libraries.
     ('filter_func', Optional[Callable[["Library"], bool]]),
     # Sort function to control the order in which outputs are listed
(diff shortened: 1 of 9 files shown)


src/hammer-tech/library_filter.py
```diff
@@ -24,14 +26,45 @@ def spice_file(self) -> Optional[str]: pass
...
 
 class LibraryFilter(NamedTuple('LibraryFilter', [
     ('tag', str),
     ('description', str),
     # Is the resulting string intended to be a file?
     ('is_file', bool),
-    # Function to extract desired string(s) out of the library.
-    ('extraction_func', Callable[["Library"], List[str]]),
+    # Function to extract desired path(s) out of the library.
+    # Returns either a list of library-relative paths.
```
colinschmidt: Delete either

src/hammer-tech/library_filter.py
```diff
@@ -24,14 +26,45 @@ def spice_file(self) -> Optional[str]: pass
...
+    ('paths_func', PathsFunctionType),
```
colinschmidt: There is no sane default that allows this to be optional?

src/hammer-vlsi/hammer_utils/__init__.py
```diff
@@ -254,11 +255,23 @@ def is_union(t: Any) -> bool:
...
         import typing
-        if a == dict and b == typing.Dict:
+
+        if isinstance(a, str) and isinstance(b, str):
+            # Two strings (e.g. if both are stringly-typed mypy annotations).
+            # Just check if they are identical.
+            return a == b
+        elif isinstance(a, str) and isinstance(b, type):
+            # A string and a type.
+            # Likely the first is a mypy stringly-typed annotation and the
```
colinschmidt: What happens in the "unlikely" case?

src/hammer-vlsi/test.py
```diff
@@ -212,6 +212,117 @@ class HammerTechnologyTest(unittest.TestCase):
...
+        """
+        Test that filters whose extraction functions return extra (non-path)
+        metadata.
+        """
+        import hammer_config
+
+        tech_dir, tech_dir_base = HammerToolTestHelpers.create_tech_dir("dummy28")
+        tech_json_filename = os.path.join(tech_dir, "dummy28.tech.json")
+
+        def add_dont_use_list(d: Dict[str, Any]) -> Dict[str, Any]:
```
colinschmidt: I don't get the name of this function? Is it just copy pasta?

src/hammer-vlsi/test.py
```diff
@@ -212,6 +212,117 @@ class HammerTechnologyTest(unittest.TestCase):
...
+            assert lib.milkyway_techfile is not None
+            return [lib.milkyway_techfile]
+
+        def extraction_func(lib: hammer_tech.Library, paths: List[str]) -> List[str]:
+            assert len(paths) == 1
+            if lib.name is None:
+                name = ""
+            else:
+                name = str(lib.name)
+            return [json.dumps({"path": paths[0], "name": name}, indent=4)]
```
colinschmidt: Is this indent necessary?

src/hammer-vlsi/test.py
```diff
@@ -212,6 +212,117 @@ class HammerTechnologyTest(unittest.TestCase):
...
+                                   is_file=True, filter_func=filter_func,
+                                   paths_func=paths_func,
+                                   extraction_func=extraction_func,
+                                   sort_func=sort_func)
+
+        database = hammer_config.HammerDatabase()
+        tech.set_database(database)
+        raw = tech.process_library_filter(pre_filts=[], filt=filter,
+                                          must_exist=False,
+                                          output_func=lambda str, _: [str])
```
colinschmidt: Can you name the variable something else? I'm pretty sure its not referencing the type "str" but its a bit confusing.

src/hammer-vlsi/hammer_utils/__init__.py
edwardcwang: What I mean is that if someone writes nonsensical in the stringly-typed annotation (or is abusing the annotation field for other stuff), it will obviously fail as expected.

```python
def my_func(x: "this is a totally nonsensical type annotation", y: int) -> bool: ...
```

src/hammer-vlsi/test.py
edwardcwang: It's not necessary especially since it's test code but it does make our lives easier if the test fails since it's at least formatted nicely in that case.

src/hammer-tech/library_filter.py
edwardcwang: Done, thanks for the catch.

src/hammer-vlsi/test.py
edwardcwang: Yes this is accidental copy pasta. Will fix.

src/hammer-vlsi/test.py
edwardcwang: Good catch, fixed.

colinschmidt:
Cool thanks for fixing this stuff.

PR #265, opened 2018-11-19
Support TLU+ map files

edwardcwang:
Updated version of #209 meant to be merged to master.


src/hammer-tech/filters.py
```diff
@@ -244,3 +244,13 @@ def select_tlu_min_cap(lib: "Library") -> List[str]:
                 return []
 
         return LibraryFilter.new("tlu_min", "TLU+ min cap db", is_file=True, paths_func=select_tlu_min_cap)
+
+    @property
+    def tlu_map_file_filter(self) -> LibraryFilter:
+        """Select tlu map files."""
```
colinschmidt: I think the comment should read "Select TLU+ map files."

edwardcwang: Done

PR #266, opened 2018-11-20
Add power specification support

colinschmidt:
Not tested yet but wanted to see if anyone had early feedback

### DIFF
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -199,7 +199,7 @@ vlsi.inputs:
   # how many domains you have and what their relation is.
   # Valid options:
   # auto - Hammer will generate a power specification based on other settings. Currently unimplemented
-  # manual - Use the file given in power_spec_path with type given in power_spec_type
+  # manual - Use the value given in power_spec_contents with type given in power_spec_type
   # empty - Do not use any power specification
   power_spec_mode: empty
   # Optional: Power specification type. (str)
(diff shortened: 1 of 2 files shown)


edwardcwang:
Are we interested in an `auto` mode and corresponding higher-level API to generate the power spec automatically?

colinschmidt:
Yes we do want that. It won't be in this PR but I can add the mode setting and mark it not implemented if you would prefer.

edwardcwang:
Yes, how about let's add an `auto` mode and mark it as unimplemented for now. e.g. `raise NotImplementedError("...")`

colinschmidt:
Rebased and fixed feedback

edwardcwang:
Generally looks good. What do you think about `power_spec_path` -> `power_spec_contents`? This will make things a bit easier and more deterministic for programs that generate Hammer IR (i.e. not having to worry about finding a temporary folder and dumping the file). It also gives more options for specifying the location of the file (e.g. we could do a subst to get the filename then transclude to get the contents).

```
vlsi.inputs:
  power_spec_contents: "mychip.cpf"
  power_spec_contents_meta: transclude
```

colinschmidt:
Hmm. That means hammer will have to write the contents to a file, and this will be a very long string. I think your suggestion does mirror what happened with floorplan constraints though. And I can see the evolution of power specifications following that path as well. Initially just raw power specification constraints as a dump in Hammer, then IR updates that let you specify similar constraints as hammerIR and then either a DSL or higher level IR constraints.

I'll make the updates soon. Thanks for the feedback.

colinschmidt:
Fixed. Tested on a demo.

PR #272, opened 2018-11-23
Implement crossappendref

edwardcwang:


src/hammer_config_test/test.py
```diff
@@ -212,6 +212,25 @@ def test_meta_crossappend(self) -> None:
...
+        self.assertEqual(db.get_setting("foo.bar.languages"), ["scala", "python"])
+
+    def test_meta_crossappendref(self) -> None:
+        """
+        Test that the meta attribute "crossappendref" works.
+        """
+        db = hammer_config.HammerDatabase()
+        base = hammer_config.load_config_from_string("""
+foo.bar.dsl: ["scala"]
+"snakes": ["python"]
```
colinschmidt: Why is `snakes` in quotes here?

edwardcwang: It's an oversight. Fixing

PR #291, opened 2019-02-05
Add hier-par-to-syn helper

colinschmidt:
This is useful if your hierarchical `auto` job fails
in the middle and you need to restart it manually


colinschmidt:
So if I run a manually `syn-par-Mod1` step. In order to get `syn-input-toplevel.json` for `syn-par-toplevel` I need to muck with json right now. This just does that mucking for me. To me it was analogous to the `syn-to-par` existing even though we have `syn-par`. Rather than it being an explicit function before it was just kind of shoved into the `auto` portion of `cli_driver.py`
https://github.com/ucb-bar/hammer/blob/master/src/hammer-vlsi/hammer_vlsi/cli_driver.py#L729

PR #301, opened 2019-02-11
Addition of hierarchical drc and lvs actions

colinschmidt:
doesn't do much yet but it runs


colinschmidt:
I've tested the DRC change for my hierarchical designs, and it works.

edwardcwang:
So it seems that this hierarchical DRC/LVS is different from the "hierarchical" flag in Calibre right? I'm not sure how we go about running DRC/LVS at each level of a hierarchical run which only ILMs are generated at each level as opposed to full GDS(?)

colinschmidt:
This is sort of a precursor for better block-level drc/lvs support.
This enables targets like `drc-MyBlock` for simliarity with other targets like `syn-par-MyBlock`.
Current plugin's don't do much for this but will eventually consume that info I believe.

edwardcwang:
Okay this is just infrastructural support and doesn't actually work with current drc/lvs plugins right?

colinschmidt:
Its currently only infrastructural. No current plugins or known hooks override the `extra_hier...drc/lvs_hooks` .

You can use the targets but they are the same as just `drc` or `lvs` i.e.
```
./hammer-vlsi.py -p par-MyBlock/par-output.json -o drc-input-MyBlock.json par-to-drc
```
plus either this
```
./hammer-vlsi.py -p drc-input-MyBlock.json drc
```
or this
```
./hammer-vlsi.py -p drc-input-MyBlock.json drc-MyBlock
```
will currently work and produce the same output

PR #302, opened 2019-02-11
Always log stdout and stderr of lsf to file

colinschmidt:
Some tools won't produce output to their own log
or won't produce output until after the tool may fail
this makes sure we have something to show for our time


jwright6323:
cc @vighneshiyer This is the reason I said not to use the native LSF settings; once this is in you should give that a shot.

jwright6323:
LGTM, squashing per in-person request by @colinschmidt

PR #308, opened 2019-02-14
Add support for stackups in the tech json

jwright6323:
This adds basic fields for the stackup info (width, spacing, index, etc.) and a method for getting info.

### DIFF
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -23,7 +23,7 @@ from hammer_utils import (LEFUtils, add_lists, deeplist, get_or_else,
 
 from library_filter import LibraryFilter
 from filters import LibraryFilterHolder
-from stackup import Stackup
+from stackup import RoutingDirection, WidthSpacingTuple, Metal, Stackup
 
 # Holds the list of pre-implemented filters.
 # Access it like hammer_tech.filters.lef_filter
@@ -795,10 +795,9 @@ class HammerTechnology:
 
         return check_isfile
 
-
-    def get_stackup_by_name(self, name: str) -> "Stackup":
+    def get_stackup_by_name(self, name: str) -> Stackup:
         """
-        Returns the stackup details for the given key
+        Return the stackup details for the given key.
         """
         if self.config.stackups is not None:
             for item in list(self.config.stackups):
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -23,6 +23,7 @@ from hammer_utils import (LEFUtils, add_lists, deeplist, get_or_else,
 
 from library_filter import LibraryFilter
 from filters import LibraryFilterHolder
+from stackup import Stackup
 
 # Holds the list of pre-implemented filters.
 # Access it like hammer_tech.filters.lef_filter
@@ -793,6 +794,19 @@ class HammerTechnology:
                 return path
 
         return check_isfile
+
+
+    def get_stackup_by_name(self, name: str) -> "Stackup":
+        """
+        Returns the stackup details for the given key
+        """
+        if self.config.stackups is not None:
+            for item in list(self.config.stackups):
+                if item["name"] == name:
+                    return Stackup.from_setting(item)
+            raise ValueError("Stackup named %s is not defined in tech JSON" % name)
+        else:
+            raise ValueError("Tech JSON does not specify any stackups")
 
 
 class HammerTechnologyUtils:
--- a/src/hammer-tech/schema.json
+++ b/src/hammer-tech/schema.json
@@ -74,21 +74,6 @@
           "gds file" : {
             "type" : "string"
           },
-          "metal layers" : {
-            "type" : "array",
-            "items" : {
-              "title" : "MetalLayer",
-              "type" : "object",
-              "properties" : {
-                "name" : {
-                  "type" : "string"
-                },
-                "preferred routing direction" : {
-                  "type" : "string"
-                }
-              }
-            }
-          },
           "milkyway lib in dir" : {
             "type" : "string"
           },
@@ -194,7 +179,64 @@
             "type" : "string"
           }
         },
-        "additionalProperties" : false
+        "additionalproperties" : false
+      }
+    },
+    "stackups" : {
+      "type" : "array",
+      "items" : {
+        "title" : "Stackup",
+        "type" : "object",
+        "additionalproperties" : false,
+        "properties" : {
+          "name" : {
+            "type" : "string"
+          },
+          "metals" : {
+            "type" : "array",
+            "items" : {
+              "title" : "Metal",
+              "type" : "object",
+              "additionalproperties" : false,
+              "properties" : {
+                "name" : {
+                  "type" : "string"
+                },
+                "index" : {
+                  "type" : "integer"
+                },
+                "direction" : {
+                  "type" : "string"
+                },
+                "min_width" : {
+                  "type" : "number"
+                },
+                "pitch" : {
+                  "type" : "number"
+                },
+                "offset" : {
+                  "type" : "number"
+                },
+                "power_strap_widths_and_spacings" : {
+                  "type" : "array",
+                  "items" : {
+                    "title" : "WidthSpacingTuple",
+                    "type" : "object",
+                    "additionalproperties" : false,
+                    "properties" : {
+                      "width_at_least" : {
+                        "type" : "number"
+                      },
+                      "min_spacing" : {
+                        "type" : "number"
+                      }
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
       }
     }
   }
--- a/src/hammer-tech/stackup.py
+++ b/src/hammer-tech/stackup.py
@@ -2,15 +2,20 @@
 # -*- coding: utf-8 -*-
 #
 #  stackup.py
-#  Data structures to represent technology stackup options
+#  Data structures to represent technology stackup options.
 #
 #  See LICENSE for licence details.
 
 from enum import Enum
-from typing import Any, Callable, Iterable, List, NamedTuple, Optional, Tuple, Dict
+from typing import List, NamedTuple, Tuple, Dict
 from hammer_utils import reverse_dict
 
 class RoutingDirection(Enum):
+    """
+    Represents a preferred routing direction for a metal layer.
+    Note that this represents a *preferred* direction, not a DRC rule.
+    """
+
     Vertical = 1
     Horizontal = 2
     Redistribution = 3
@@ -25,6 +30,7 @@ class RoutingDirection(Enum):
 
     @staticmethod
     def from_str(input_str: str) -> "RoutingDirection":
+        # pylint: disable=missing-docstring
         try:
             return RoutingDirection.__mapping()[input_str]
         except KeyError:
@@ -34,6 +40,11 @@ class RoutingDirection(Enum):
         return reverse_dict(RoutingDirection.__mapping())[self]
 
     def opposite(self) -> "RoutingDirection":
+        """
+        Return the opposite routing direction.
+        For Redistribution, this returns itself.
+        :return: Opposite routing direction
+        """
         if self == RoutingDirection.Vertical:
             return RoutingDirection.Horizontal
         elif self == RoutingDirection.Horizontal:
@@ -41,68 +52,99 @@ class RoutingDirection(Enum):
         else:
             return self
 
-# Note to maintainers: This is mirrored in schema.json- don't change one without the other
 
-# A tuple of wire width limit and spacing for generating a piecewise linear rule for spacing based on wire width
-# width_at_least: Any wires larger than this must obey the minSpacing rule
-# min_spacing: The minimum spacing for this bin. If a wire is wider than multiple entries, the worst-case (larger) minSpacing wins.
 class WidthSpacingTuple(NamedTuple('WidthSpacingTuple', [
-    ('width_at_least', float),
-    ('min_spacing', float)
+        ('width_at_least', float),
+        ('min_spacing', float)
 ])):
+    """
+    A tuple of wire width limit and spacing for generating a piecewise linear rule
+    for spacing based on wire width.
+
+    width_at_least: Any wires larger than this must obey the minSpacing rule.
+    min_spacing: The minimum spacing for this bin.
+                 If a wire is wider than multiple entries, the worst-case (larger)
+                 minSpacing wins.
+
+    Note to maintainers: This is mirrored in schema.json - don't change one without the other!
+    """
     __slots__ = ()
 
     @staticmethod
     def from_setting(d: dict) -> "WidthSpacingTuple":
-        w = float(d["width_at_least"])
-        s = float(d["min_spacing"])
-        assert w >= 0.0
-        assert s > 0.0
+        # pylint: disable=missing-docstring
+        width_at_least = float(d["width_at_least"])
+        min_spacing = float(d["min_spacing"])
+        assert width_at_least >= 0.0
+        assert min_spacing > 0.0
         return WidthSpacingTuple(
-            width_at_least=w,
-            min_spacing=s
+            width_at_least=width_at_least,
+            min_spacing=min_spacing
         )
 
     @staticmethod
     def from_list(l: List[dict]) -> List["WidthSpacingTuple"]:
-        out = sorted(list(map(lambda x: WidthSpacingTuple.from_setting(x), l)), key=lambda x: x.width_at_least)
-        s = 0.0
+        out = sorted(list(map(WidthSpacingTuple.from_setting, l)), key=lambda x: x.width_at_least)
+
+        # Check that spacings increase.
+        current_spacing = 0.0
         for wst in out:
-            assert wst.min_spacing >= s
-            s = wst.min_spacing
+            assert wst.min_spacing >= current_spacing
+            current_spacing = wst.min_spacing
         return out
 
 
-# A metal layer and some basic info about it
-# name: M1, M2, etc.
-# index: The order in the stackup (lower is closer to the substrate)
-# direction: The preferred routing direction of this metal layer, or "redistribution" for non-routing top-level redistribution metals like Aluminum
-# min_width: The minimum wire width for this layer
-# pitch: The minimum cross-mask pitch for this layer (NOT same-mask pitch for multiply-patterned layers)
-# offset: The routing track offset from the origin for the first track in this layer (0 = first track is on an axis)
-# power_strap_widths_and_spacings: A list of WidthSpacingTuples that specify the minimum spacing rules for an infinitely long wire of variying width
 class Metal(NamedTuple('Metal', [
-    ('name', str),
-    ('index', int),
-    ('direction', "RoutingDirection"),
-    ('min_width', float),
-    ('pitch', float),
-    ('offset', float),
-    ('power_strap_widths_and_spacings', List[WidthSpacingTuple])
+        ('name', str),
+        ('index', int),
+        ('direction', RoutingDirection),
+        ('min_width', float),
+        ('pitch', float),
+        ('offset', float),
+        ('power_strap_widths_and_spacings', List[WidthSpacingTuple])
 ])):
+    """
+    A metal layer and some basic info about it.
+
+    name: Metal layer name (e.g. M1, M2).
+    index: The order in the stackup (lower is closer to the substrate).
+    direction: The preferred routing direction of this metal layer, or
+               RoutingDirection.Redistribution for non-routing top-level
+               redistribution metals like Aluminium.
+    min_width: The minimum wire width for this layer.
+    pitch: The minimum cross-mask pitch for this layer (NOT same-mask pitch
+           for multiply-patterned layers).
+    offset: The routing track offset from the origin for the first track in this layer.
+            (0 = first track is on an axis).
+    power_strap_widths_and_spacings: A list of WidthSpacingTuples that specify the minimum
+                                     spacing rules for an infinitely long wire of variying width.
+    """
     __slots__ = ()
 
-    # TODO this is assuming that a manufacturing grid is 0.001
+    @property
     def grid_unit(self) -> float:
+        """
+        Return the manufacturing grid unit.
+
+        TODO: this assumes a manufacturing grid of 0.001
+        """
         return 0.001
 
-    # TODO internally represent numbers as integers or fixed-point to obviate the need for this
-    # ucb-bar/hammer#319
-    def snap(self, x: float) -> float:
-        return float(round(x / self.grid_unit())) * self.grid_unit()
+    def snap(self, num: float) -> float:
+        """
+        Snap a number to the grid unit.
+
+        TODO: internally represent numbers as integers or fixed-point to
+        obviate the need for this (see #319).
+
+        :param num: Number to snap
+        :return: Number snapped to grid_unit
+        """
+        return float(round(num / self.grid_unit)) * self.grid_unit
 
     @staticmethod
     def from_setting(d: dict) -> "Metal":
+        # pylint: disable=missing-docstring
         return Metal(
             name=str(d["name"]),
             index=int(d["index"]),
@@ -113,8 +155,13 @@ class Metal(NamedTuple('Metal', [
             power_strap_widths_and_spacings=WidthSpacingTuple.from_list(d["power_strap_widths_and_spacings"])
         )
 
-    # Get the minimum spacing for a provided width
     def get_spacing_for_width(self, width: float) -> float:
+        """
+        Get the minimum spacing for a provided width.
+
+        :param width: Width to calculate minimum spacing for.
+        :return: Minimum spacing for `width`
+        """
         spacing = 0.0
         for wst in self.power_strap_widths_and_spacings:
             if width >= wst.width_at_least:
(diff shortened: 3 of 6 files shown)


edwardcwang:
Is there a unit test/example for how to use this? In addition to this I still think it would be good to have a way to use Python or something to parametrize the tech plugin schema.

jwright6323:
I don't disagree that we need a parameterizable plugin but I don't want to worry about that right now. I don't think this needs a test beyond what exists.

edwardcwang:
Right now I'm not sure I understand the API and how to use it, which is why I asked for an example/test.

jwright6323:
Look at the tech plugin MR.

edwardcwang:
OK most of it appears to make sense - not sure what `spacingtable_*` is supposed to represent though.

jwright6323:
SPACINGTABLE is a standard LEF property that expresses spacing as a function of wire width and run length. This is just the entry for maximum run length (which is conservative, i.e. will be guaranteed DRC-free)

edwardcwang:
I'd like to add a Hammer object for encoding the metal layer information, along the following lines (sketch below) for data validation/documentation's sake:

```python
class MetalLayerInfo(NamedTuple):
  name: str
  index: int
  ...
```

src/hammer-tech/schema.json
```diff
@@ -194,7 +194,43 @@
...
+          "min_width" : {
+            "type" : "number"
+          },
+          "pitch" : {
+            "type" : "number"
+          },
+          "offset" : {
+            "type" : "number"
+          },
+          "spacingtable_widths" : {
```
edwardcwang: If this is meant to model the following LEF info, maybe we can have spacingtable be an array of tuples instead?
```
SPACINGTABLE ORTHOGONAL 
WITHIN 0.15 SPACING 0.11
WITHIN 0.13 SPACING 0.13
WITHIN 0.11 SPACING 0.15 
```

jwright6323: Hmm. Maybe, let me sit on this. I think that'll be more verbose than I'd like, but this is not a strong opinion.

jwright6323: I think my decision would be made by whichever makes the code easier to decipher.

edwardcwang: From a data structure perspective I would prefer the list of tuples as it better represents the structure in the LEF, unless I am mistaken there

jwright6323: True, the list sizes should be the same but that should be an error imo. I think the example you listed is a degenerate case, though- it's usually a multidimensional structure and we've distilled it into a simpler case by sacrificing area efficiency a bit by assuming infinite run length.

edwardcwang: Would the following structure be more future proof then?

```python
class SpacingEntry(NamedTuple):
  # if width < than this
  width: float
  # if length < than this
  length: Optional[float]
  # then follow this spacing
  spacing: float
```

We could also be agile about it and not include the `length` field until we need it

jwright6323: I don't see us ever needing this info, TBH. I only included it so we can calculate the power strap widths, but even that is a bit of a micro-optimization IMO. I would rather go with this and continue making progress on stuff that matters, this is getting to be bikeshedding a bit.

edwardcwang: I suppose if we are going to include it we may as well do it in an extensible way. If it's extra work/burden for you I can help with this.

jwright6323: I think it's a waste of time right now. We can change it later. Feel free to file an issue.

edwardcwang: Resolved offline: file an issue and deal with API breakages later

edwardcwang:
@jwright6323 Were you still planning to add the Hammer object things or did you want me to do that?

jwright6323:
Yes, I'll get to it. I want to merge this, the tech plugin MR, and the cad plugin PR at the same time when everything's ready. Not there yet.

jwright6323:
Marking this "Do Not Merge" since I think I want to experiment with TLEF parsing a bit. If it's harder that I think we'll still go with what's here.

src/hammer-tech/hammer_tech.py
```diff
@@ -10,7 +10,7 @@
 import os
 import subprocess
 from abc import ABCMeta, abstractmethod
-from typing import Any, Callable, Iterable, List, NamedTuple, Optional, Tuple
+from typing import Any, Callable, Iterable, List, NamedTuple, Optional, Tuple, Dict
```
jwright6323: We can revert this, I think it's no longer needed.

jwright6323:
Removing the DO NOT MERGE tag since this is passing and more or less complete, modulo a code review.

src/hammer-tech/schema.json
```diff
@@ -194,7 +194,64 @@
             "type" : "string"
           }
         },
-        "additionalProperties" : false
+        "additionalproperties" : false
+      }
+    },
+    "stackups" : {
```
colinschmidt: Does this make the old metal layers (https://github.com/ucb-bar/hammer/pull/308/files#diff-57aef1ccd100a4823016fa230d58d5e7L77) defunct?
If so we should remove that.

src/hammer-tech/schema.json
```diff
@@ -194,7 +194,64 @@
...
+      "type" : "array",
+      "items" : {
+        "title" : "Stackup",
+        "type" : "object",
+        "additionalproperties" : false,
```
colinschmidt: What does this do?

src/hammer-tech/stackup.py
```diff
@@ -0,0 +1,184 @@
...
+    @staticmethod
+    def from_setting(d: dict) -> "WidthSpacingTuple":
+        return WidthSpacingTuple(
+            width_at_least=float(d["width_at_least"]),
+            min_spacing=float(d["min_spacing"])
+        )
+
+# A metal layer and some basic info about it
+# name: M1, M2, etc.
+# index: The order in the stackup (lower # is closer to the wafer)
```
colinschmidt: Maybe closer to the substrate? To me the whole thing is the wafer so all the metals are inside the wafer.

src/hammer-tech/stackup.py
```diff
@@ -0,0 +1,184 @@
...
+            pitch=float(d["pitch"]),
+            offset=float(d["offset"]),
+            power_strap_widths_and_spacings=sorted(list(map(lambda x: WidthSpacingTuple.from_setting(x), list(d["power_strap_widths_and_spacings"]))), key=lambda x: x.width_at_least)
+        )
+
+    def get_spacing_for_width(self, width: float) -> float:
+        spacing = 0.0
+        for wst in self.power_strap_widths_and_spacings:
+            if width >= wst.width_at_least:
+                spacing = max(spacing, wst.min_spacing)
```
colinschmidt: I assumed you were sorting them before so you could early-out here. Is there some other reason to sort, or a reason to not early-out?

src/hammer-tech/stackup.py
```diff
@@ -0,0 +1,184 @@
...
+        return spacing
+
+    def min_spacing_from_pitch(self, pitch: float) -> float:
+        ws = self.power_strap_widths_and_spacings
+        spacing = ws[0].min_spacing
+        for first, second in zip(ws[:-1], ws[1:]):
```
colinschmidt: I don't really get why you are comparing with the next larger width?

src/hammer-tech/stackup.py
```diff
@@ -0,0 +1,184 @@
...
+            if pitch >= (first.min_spacing + second.width_at_least):
+                spacing = second.min_spacing
+        return spacing
+
+    # This method will return the maximum width a wire can be to consume a given number of routing tracks
+    # This assumes the neighbors of the theick wire are minimum-width routes
```
colinschmidt: Typo: `theick` -> `thick`

src/hammer-vlsi/defaults.yml
```diff
@@ -367,3 +367,13 @@ lvs.submit:
...
   settings: "vlsi.submit.settings"
   settings_meta: lazycrossref
+
+# Key name for the technology stackup
+# This should exist in the stackups list in the tech json
+technology.stackup: null
+# This should specify the TOPMOST metal layer the standard cells use for power rails
+# Note that this is not usually stackup specific; it is based on the std cell library
+technology.std_cell_rail_layer: null
+# This is used to provide a reference master for generating standard cells
```
colinschmidt: `for generating standard cell rails`

src/hammer-tech/stackup.py
jwright6323: I'll document this better. I need to figure out a good way to express it, but it makes sense in my head. Would be helpful to draw it on a whiteboard later today.

src/hammer-tech/stackup.py
jwright6323: This code predates the sorting, and I neglected to update it. I'll fix.

src/hammer-tech/schema.json
jwright6323: @edwardcwang ? This is in all the other objects. I assume it disallows adding additional properties not specified in the schema.

src/hammer-tech/stackup.py
jwright6323: lol yes, that's a horrible comment. I'll fix.

src/hammer-tech/schema.json
jwright6323: oh, huh. Yes, I think so.

colinschmidt: As discussed in meeting this should be dead and will be removed.

src/hammer-tech/stackup.py
jwright6323: Changed to wide, which is confusing with the T and W acronyms

colinschmidt: That was the other suggestion I almost made :)

src/hammer-tech/stackup.py
```diff
@@ -0,0 +1,198 @@
...
+        width = (s3w2 - spacing*3)/2
+        start = self.min_width / 2 + spacing
+        # TODO this is assuming that a manufacturing grid is 0.001
+        # Note the rounding here seems to get around some floating-point roundoff issues in prior calculations
+        if force_even and ((int(round(width * 1000)) % 2) != 0):
+                width = width - 0.001
+                start = start + 0.001
+        return (width, spacing, start)
+
+    # TODO implement M W X* W M style wires, where X is slightly narrower than W and centered on-grid
```
edwardcwang: General comments about the methods on `Metal`: I think that, after reading it through, they make sense, though:

1) would be good if we can have some unit tests for these functions
2) I think some of them could be written more succinctly/functionally but I'm hesitant without unit tests.

I can help with tests if you think that would help with me understanding it

src/hammer-tech/stackup.py
```diff
@@ -0,0 +1,198 @@
...
+    @staticmethod
+    def from_setting(d: dict) -> "Stackup":
+        return Stackup(
+            name=str(d["name"]),
+            metals=list(map(lambda x: Metal.from_setting(x), list(d["metals"])))
+        )
+
+
+    def get_metal(self, name: str) -> "Metal":
+        # TODO don't brute force this
```
edwardcwang: Not sure what is meant by this; is this about writing it in a more functional way?

src/hammer-vlsi/defaults.yml
```diff
@@ -367,3 +367,13 @@ lvs.submit:
   command_meta: lazysubst
   settings: "vlsi.submit.settings"
   settings_meta: lazycrossref
+
+# Key name for the technology stackup
+# This should exist in the stackups list in the tech json
+technology.stackup: null
```
edwardcwang: They seem generally sane, 3 comments about these:

1) Would be good to add `# type: <type>` comments for each of these to facilitate future documentation scraping
2) Let's put this under `technology.core` maybe? So `vlsi.technology` is about how hammer-vlsi interacts with technology. I'd imagine these settings could go under `technology.<mytech>` since we may want different settings per technology. Take a look at extracted_tarballs_dir?
3) Would put them all under one heading:
```
technology.core:
  # foo bar foo bar
  # type: ...
  stackup: null
  # foo bar
  # type: ...
  std_cell_rail_layer: null
```

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -934,6 +934,13 @@ def get_mmmc_corners(self) -> List[MMMCCorner]:
             output.append(corn)
         return output
 
+    def get_stackup(self) -> "Stackup":
+        """
+        Get the stackup provided by the technology key
+        """
+        # TODO how does python cache this? Do we need to avoid re-processing this every time?
```
edwardcwang: I think we can memoize this if it becomes a performance issue

src/hammer-tech/stackup.py
```diff
@@ -0,0 +1,198 @@
...
+# index: The order in the stackup (lower is closer to the substrate)
+# direction: The preferred routing direction of this metal layer, or "redistribution" for non-routing top-level redistribution metals like Aluminum
+# min_width: The minimum wire width for this layer
+# pitch: The minimum cross-mask pitch for this layer (NOT same-mask pitch for multiply-patterned layers)
+# offset: The routing track offset from the origin for the first track in this layer (0 = first track is on an axis)
+# power_strap_widths_and_spacings: A list of WidthSpacingTuples that specify the minimum spacing rules for an infinitely long wire of variying width
+class Metal(NamedTuple('Metal', [
+    ('name', str),
+    ('index', int),
+    ('direction', "RoutingDirection"),
```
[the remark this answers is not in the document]
jwright6323: IIRC mypy was unhappy with this, which is why it's there, but I could be mistaken. This works, though, so I don't really want to change it / retest it.

src/hammer-tech/stackup.py
```diff
@@ -0,0 +1,198 @@
...
+    @staticmethod
+    def from_setting(d: dict) -> "Metal":
+        return Metal(
+            name=str(d["name"]),
+            index=int(d["index"]),
+            direction=RoutingDirection.from_str(d["direction"]),
+            min_width=float(d["min_width"]),
+            pitch=float(d["pitch"]),
+            offset=float(d["offset"]),
+            power_strap_widths_and_spacings=sorted(list(map(lambda x: WidthSpacingTuple.from_setting(x), list(d["power_strap_widths_and_spacings"]))), key=lambda x: x.width_at_least)
```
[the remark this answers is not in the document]
jwright6323: Yeah, that's a good idea.

src/hammer-tech/schema.json
jwright6323: @edwardcwang It appears this is unused in hammer. It's used in icc and dc... what are the plans for those plugins? I don't want to spend a lot of time "correcting" them if they're broken to begin with.

jwright6323: I commented them out of those tools, so it behaves as if the field is set to `None` There is a link to this PR in the comments.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
jwright6323: OK, I'll leave the comment in. I doubt it will be an issue.

src/hammer-tech/stackup.py
jwright6323: Yes, I deleted the comment. I think it's fine as-is.

src/hammer-tech/stackup.py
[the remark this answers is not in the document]
jwright6323: done

src/hammer-tech/stackup.py
```diff
@@ -0,0 +1,198 @@
...
+        )
+
+    def get_spacing_for_width(self, width: float) -> float:
```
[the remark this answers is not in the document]
jwright6323: Added comments.

src/hammer-tech/stackup.py
jwright6323: I don't think there's much room to write these any more succinctly than they already are. I also think if you did, it would be confusing (more so than they are currently).

I can throw some tests together- I agree this is somewhat "magical" code and would be easy for someone to break.

src/hammer-vlsi/defaults.yml
jwright6323: I think these should go where `placement_site` is, which is currently`vlsi.technology`. I can move `placement_site` if you like `technology.core` better. No preference.

src/hammer-tech/stackup.py
edwardcwang: Some tests on this would be good. For example some of the checks done with loops+ifs could be done with `filter` etc. Either way I think tests can help with both understanding and regression catching.

src/hammer-vlsi/defaults.yml
edwardcwang: So far settings like `placement_site`, this is the order of precedence I have in my head. Say I'm interested in technology property `foobar` (or `placement_site` etc). Then the code will look in the following order (highest to lowest priority):

1. `vlsi.technology.overrides.foobar`
1. `technology.<mytech>.foobar`
2. `technology.core.foobar`, if the above does not exist

This enables a config database of the following e.g.:

In techs.yml
```
# technology defaults
technology.mytech45.foobar: 12345
technology.footech99.foobar: 67876

# any unknown technology, use this
technology.core.foobar: 1000
```

In myproject.yml
```
# in this project use this regardless of technology (for incremental adoption)
technology.core.
vlsi.technology.overrides.foobar: 9001
```

Also, not sure it makes sense for everything in `vlsi.technology` to move to `technology.core` e.g. `extra_macro_sizes` is design-specific.

jwright6323: Does this currently exist? I get 2 and 3, but I don't really understand 1. Can't your personal yml just override `technology.<mytech>.foobar`? This seems a little complicated to me.

edwardcwang: 2 and 3 sort of exist (see `extracted_tarballs_dir` for an example). Maybe we don't need (1), but then there's no way to globally override `foobar` which may not really make sense without a technology anyway.

jwright6323: I am inclined to leave this as-is then (or at least move things to `vlsi.technology` to be consistent with `placement_site`) if this requires additional pythoning that is not automatic. IMO it's pretty simple to just have separate `myproject.techA.yml` and `myproject.techB.yml` files that separate out technology-specific settings. I actually think it's undesirable to put all possible technology settings in a single file or database for NDA reasons.

jwright6323: Filed #317, I think this has uncovered some things we need to take a look at, IMO.

jwright6323: I'm going to leave them as they currently are, and we should clean them up with #317 when we get to that.

edwardcwang: Dealing with this later in #317 is fine.

Would it still be possible for you to add the type annotations and put them under one heading (`technology:` is fine for now)?

Regarding "I actually think it's undesirable to put all possible technology settings in a single file or database for NDA reasons." --> the hammer database is actually aggregated from many separate files, so this "global" override would just be something mixed in. It's not one giant file.

jwright6323: ...Actually no, I'm going to make them `technology.core.*` because that way if people agree with my thoughts in #317 they need not be changed.

edwardcwang: For putting them into one YAML heading it doesn't actually matter what it is, no?

```
technology: # or technology.core
  # foo bar foo bar
  # type: ...
  stackup: null
  # foo bar
  # type: ...
  std_cell_rail_layer: null
```

jwright6323: oh, yes, I agreed with that change and forgot to mention it. I already fixed that :).

jwright6323: Marking this resolved, we can debate in #317 if needed.

src/hammer-tech/stackup.py
jwright6323: Tests added.

edwardcwang:
Looking again

src/hammer-tech/stackup.py
```diff
@@ -0,0 +1,230 @@
...
+#  stackup.py
+#  Data structures to represent technology stackup options
+#
+#  See LICENSE for licence details.
+
+from enum import Enum
+from typing import Any, Callable, Iterable, List, NamedTuple, Optional, Tuple, Dict
+from hammer_utils import reverse_dict
+
+class RoutingDirection(Enum):
```
colinschmidt: Random thought that could be filed as issue, do older technologies have bidirectional metals? Would those be labeled as a fourth category or would treating them as redistribution be ok?

src/hammer-tech/stackup.py
```diff
@@ -0,0 +1,230 @@
...
+    __slots__ = ()
+
+    # TODO this is assuming that a manufacturing grid is 0.001
+    def grid_unit(self) -> float:
+        return 0.001
+
+    # TODO internally represent numbers as integers or fixed-point to obviate the need for this
+    # ucb-bar/hammer#319
+    def snap(self, x: float) -> float:
+        return float(round(x / self.grid_unit())) * self.grid_unit()
```
colinschmidt: Do you need the float conversion? I would have assumed `int` * `float` = `float`

src/hammer-tech/stackup.py
```diff
@@ -0,0 +1,230 @@
...
+    # i.e. T W W T
+    # T = thin / min-width
+    # W = wide
+    # Returns width, spacing, and start
+    # See min_spacing_from_pitch documentation for an explanation of the calculation
+    # force_even: True to force the strap width to be an even multiple of the grid unit
+    def get_width_spacing_start_twwt(self, tracks: int, force_even: bool = False) -> Tuple[float, float, float]:
+        ws = self.power_strap_widths_and_spacings
+        spacing = ws[0].min_spacing
+        # the T W W T pattern contains two wires (W2) and 3 spaces (S3)
```
colinschmidt: I think this needs to be slightly more clear given the next line.
Does s3w2 include all the wires and spaces above?
It doesn't appear to but I'm not exactly sure what is excluded.
I initially thought it would hold this: 
`1minW + 1Space + tPitches + 1Space + tPitches + 1Space + 1minW = (2t+2)Pitches+1Space`
assuming `Pitch = 1Space + 1 minW`
but instead it seems to hold `(2t+1)Pitches - 1minW` so you would be exlcluding the first and last `1minW +1Space` from above.
So s3w2 holds just the distance from the left hand side of the first Wide wire to the right hand side of the second Wide wire?
Do I understand this correctly?

src/hammer-tech/stackup.py
```diff
@@ -0,0 +1,230 @@
...
+        s3w2 = self.snap(((2*tracks) + 1) * self.pitch - self.min_width)
+        width = self.snap((s3w2 - spacing*3)/2.0)
+        for first, second in zip(ws[:-1], ws[1:]):
+            if s3w2 >= (second.min_spacing*3 + second.width_at_least*2):
```
colinschmidt: Some future PR should try to abstract the number of Ts and Ws.

src/hammer-tech/stackup.py
jwright6323: This is referring to the preferred routing direction rather than a DRC rule. Even older techs still do VHVHV... because having two same-direction metals back-to-back is bad for routing.

src/hammer-tech/stackup.py
jwright6323: Still missing a space. I'll draw it for you.

src/hammer-tech/stackup.py
jwright6323: Not sure what you mean here.

colinschmidt: You give an argument to the function that is a string "TWWWT" or TWWTWWT". Just a pie in the sky.

jwright6323: Oooh. Yes that would be slick.

src/hammer-tech/stackup.py
jwright6323: Probably not.

src/hammer-tech/stackup.py
jwright6323: Resolved offline

edwardcwang:
@jwright6323 FYI this commit appears to disable all other tests...?

```
Commit:     John Wright <johnwright@eecs.berkeley.edu>
CommitDate: Sat Feb 23 18:12:26 2019 -0800

    Fix bogus assertions
---
 src/test/unittests.sh | 16 ++++++++--------
 1 file changed, 8 insertions(+), 8 deletions(-)

diff --git a/src/test/unittests.sh b/src/test/unittests.sh
index 7f2862b..f4ae2c0 100755
--- a/src/test/unittests.sh
+++ b/src/test/unittests.sh
@@ -5,14 +5,14 @@
 err=0
 trap 'err=1' ERR

-python3 ../hammer-vlsi/test.py
+#python3 ../hammer-vlsi/test.py
 python3 ../hammer-vlsi/tech_test.py
-python3 ../hammer-vlsi/constraints_test.py
-python3 ../hammer-vlsi/cli_driver_test.py
-python3 ../hammer-vlsi/utils_test.py
-python3 ../hammer-vlsi/units_test.py
-python3 ../hammer-vlsi/verilog_utils_test.py
-python3 ../hammer-vlsi/lef_utils_test.py
-python3 ../hammer_config_test/test.py
+#python3 ../hammer-vlsi/constraints_test.py
+#python3 ../hammer-vlsi/cli_driver_test.py
+#python3 ../hammer-vlsi/utils_test.py
+#python3 ../hammer-vlsi/units_test.py
+#python3 ../hammer-vlsi/verilog_utils_test.py
+#python3 ../hammer-vlsi/lef_utils_test.py
+#python3 ../hammer_config_test/test.py
```

jwright6323:
Oops. Not intended. I can fix or you can get it in your cleanup commit, whatever you prefer.

edwardcwang:
I'll take care of it, just an FYI.

src/hammer-tech/stackup.py
```diff
@@ -0,0 +1,340 @@
...
+
+        start = self.snap(self.min_width / 2.0 + spacing)
+        return (width, spacing, start)
+
+    def get_width_spacing_start_twwt(self, tracks: int, force_even: bool = False) -> Tuple[float, float, float]:
+        """
+        This method will return the maximum width a wire can be in order
+        to consume a given number of routing tracks.
+        This assumes that the wires are in the following configuration.
+        TODO: double check that this pydoc is correct
```
edwardcwang: @jwright6323 If it's easy for you to do so, would you mind helping me double check this pydoc and fill out the documentation for `force_even` below?

src/hammer-vlsi/tech_test.py
```diff
@@ -18,6 +18,7 @@
 import hammer_tech
 from hammer_tech import LibraryFilter
 from hammer_utils import deepdict
+from stackup import Stackup, Metal, WidthSpacingTuple
```
edwardcwang:
```suggestion
from hammer_tech import Stackup, Metal, WidthSpacingTuple
```

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -934,6 +934,13 @@ def get_mmmc_corners(self) -> List[MMMCCorner]:
             output.append(corn)
         return output
 
+    def get_stackup(self) -> "Stackup":
```
```suggestion
    def get_stackup(self) -> Stackup:
```

src/hammer-vlsi/tech_test.py
```diff
@@ -525,6 +526,163 @@ def add_lib_with_lef(d: Dict[str, Any]) -> Dict[str, Any]:
         # Cleanup
         shutil.rmtree(tech_dir_base)
 
+class StackupTestHelper:
```
edwardcwang: Filed #322

src/hammer-tech/stackup.py
jwright6323: done.

jwright6323:
Cleaned this up, merging.

Issue #304, opened 2019-02-13
Type-safe get_setting


edwardcwang:
Right now `get_setting` just returns `Any` which is a type-safety hazard (e.g. with `Optional[str]` getting passed in as `"null"`). Would be good to have type-safe versions of `get_setting` that have a bit more knowledge of the expected return type.

harrisonliew:
@edwardcwang I see multiple ways of addressing this, and am not sure what your intention was.

A `get_setting` calls the method in `hammer_tool`:
https://github.com/ucb-bar/hammer/blob/a627b888f5aa6957a3b73e90981881340704b5e9/src/hammer-vlsi/hammer_vlsi/hammer_tool.py#L698-L711

which then calls the method in `config_src`:
https://github.com/ucb-bar/hammer/blob/d7b29a58009ecbe6ca3bb1d87bfd318998cdb838/src/hammer_config/config_src.py#L555-L567

So if `get_setting` in `config_src` correctly passes a `null` value as `None`, it will return `"null"` to `hammer_tool` if `nullvalue` isn't specified as not `None` there.

So, do you want `hammer_tool`'s `nullvalue` to override that in `config_src` or just make all `nullvalue`s `Optional[str] = None`?

edwardcwang:
The issue is that `get_setting`'s static return type can't change depending on its parameters. The solution that I was thinking of was type-specific/safe versions of get_settings. The more sustainable solution is for `get_setting` to not be used by users directly, though...

jwright6323:
You are proposing something like `get_setting_str`, `get_setting_int`, `get_setting_float`, ... ?

edwardcwang:
In the current Python type-checking regime I don't know of another way to do this.

(On a side note: some of the more fancy programming languages have this concept of "dependent types" which appear to allow the type checker to determine the return type of a function/class based on parameters. With dependent types one is able to side-step this problem)

edwardcwang:
To add a concrete example we ran into recently: `pitch=self.get_setting("vlsi.inputs.bumps.pitch"),` gives the wrong default type even though the YAML defines this as a Decimal, not a float

jwright6323:
> To add a concrete example we ran into recently: `pitch=self.get_setting("vlsi.inputs.bumps.pitch"),` gives the wrong/unconstrained default type even though the YAML defines this as a Decimal

Fortunately @harrisonliew is fixing this particular issue in #372, but we still need a bigger solution to the issue tracked here.

PR #324, opened 2019-02-27
Change get_setting to use None as a default nullvalue

jwright6323:
This makes it return slightly saner Nones instead of string "null"s. No current calls to get_setting use "null" so this seems safe.

It's also worth noting that you may want a null value that is not type `Optional[str]` (and in fact we do this in the calibre plugin).

@harrisonliew does this solve your issue on ucb-bar/hammer-cad-plugins#40 ?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@
 *.log
 *rundir/
 /*.json
-src/test/output.json
+src/test/output*.json
 
 # Sphinx/doc junk
 doc/_build/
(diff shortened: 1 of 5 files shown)


jwright6323:
This is related to #304 but I do think a type-safe method would be preferred. IMO this is just a band-aid, but I think a useful one at least.

jwright6323:
Hmm, I thought the unit tests passed.. Let me look at it again.

edwardcwang:
Turns out we did use `"null"`in a few places, oops :/

jwright6323:
Yeah unfortunately I only checked the `get_settings` calls, not comparisons on what they return. I think I got all the `"null"`s now but it's still failing tests.

jwright6323:
@edwardcwang passing now

PR #326, opened 2019-02-27
Hammer sim

dpgrubb13:
Draft PR for feedback on Hammer Sim.  Currently supports vanilla verilog sims with VCS.  Rocketchip support coming soon.  I would like feedback on the structure of the tool I've written in addition to any stylistics things (ie.  I primarily used 'sim' instead of 'simulation').


src/hammer-vlsi/defaults.yml
```diff
@@ -35,6 +35,12 @@ vlsi.core:
   lvs_tool_path: ["${vlsi.builtins.hammer_vlsi_path}/lvs"]
   lvs_tool_path_meta: subst
 
+  # TODO(daniel)
```
edwardcwang: FYI the stylistic convention for this has been `TODO(daniel): message here`

src/hammer-vlsi/hammer_vlsi/driver.py
```diff
@@ -707,6 +776,56 @@ def run_lvs(self, hook_actions: Optional[List[HammerToolHookAction]] = None, for
...
+        run_succeeded = self.sim_tool.run(hooks_to_use)
+        if not run_succeeded:
+            self.log.error("Simulation tool %s failed! Please check its output." % self.sim_tool.name)
+            # Allow the flow to keep running, just in case.
+            # TODO: make this an option
+
+        # Record output from the tool into the JSON output.
+        # Note: the output config dict is NOT complete
+        output_config = {}  # type: Dict[str, Any]
+        #try:
```
edwardcwang: Is there something problematic here?

edwardcwang:
Also I would rebase this on top of master regularly especially as this is a draft PR

src/hammer-vlsi/hammer_vlsi/driver.py
dpgrubb13: I think this is an artifact of using some of the other 'run_*' functions as an example for how to do this.  Since the simulation currently does not have any outputs that get written out to be consumed downstream, I think I just ended up commenting this section out.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -772,6 +772,151 @@ def hcells_list(self, value: List[str]) -> None:
...
+class HammerSimTool(HammerTool):
+  
+    #@abstractmethod
+    #def fill_outputs(self) -> bool:
+    #  pass
+
+    # export_config_outputs not necessary because not passing anything downstream?
+
+    @property
+    def input_files(self) -> List[str]:
```
edwardcwang: Please use `generate_properties.py` for this rather than manually writing these getters/setters

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -772,6 +772,151 @@ def hcells_list(self, value: List[str]) -> None:
...
+
+    @compiler_opts.setter
+    def compiler_opts(self, value: List[str]) -> None:
+        """Set the simulation compiler options."""
+        if not (isinstance(value, List[str])):
+            raise TypeError("compiler_opts must be a List")
+        self.attr_setter("_compiler_opts", value)
+
+    @property
+    def timescale(self) -> str:
```
edwardcwang: Can we use a `TimeValue` instead of a raw `str` here?

jwright6323: I agree, although it should be a tuple of `TimeValue`s (unit and resolution). e.g. `(TimeScale("1 ns"), TimeScale("1 ps"))` for `` `timescale 1ns / 1ps``

A `NamedTuple` would also work.

jwright6323:
@dpgrubb13 Can you run the tests before you push so we keep the comments readable?

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -1075,8 +1077,6 @@ def get_input_ilms(self) -> List[ILMStruct]:
         """
         ilms = self.get_setting("vlsi.inputs.ilms")  # type: List[dict]
         assert isinstance(ilms, list)
-        return list(map(ILMStruct.from_setting, ilms))
```
jwright6323: where'd this go?

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -995,6 +1133,29 @@ def cpf_power_specification(self) -> str:
         return "\n".join(output)
 
 
+
+class HasSDCSupport(HammerTool):
+    """Mix-in trait with functions useful for tools with SDC-style
+    constraints."""
+    @property
+    def benchmarks(self) -> List[str]:
```
jwright6323: Why is this in `HasSDCSupport`?

src/hammer-vlsi/defaults.yml
```diff
@@ -608,3 +619,51 @@ technology.core:
...
+  # Simulation timescale
+  timescale: null
+  # C source files 
+  csrc: []
+  # Benchmarks to run on simulator
+  benchmarks: []
+  # Flags to run with the simulator
+  exec_flags: []
+  # List of TCL scripts to execute in a vcs call
+  tcl_scripts: []
```
jwright6323: this seems to be breaking the abstraction a bit

src/hammer-vlsi/defaults.yml
```diff
@@ -608,3 +619,51 @@ technology.core:
...
+  # C compiler options
+  compiler_opts: []
+  # Simulation timescale
+  timescale: null
+  # C source files 
+  csrc: []
+  # Benchmarks to run on simulator
+  benchmarks: []
+  # Flags to run with the simulator
+  exec_flags: []
```
jwright6323: this should probably live in the tool plugin

src/hammer-vlsi/defaults.yml
```diff
@@ -608,3 +619,51 @@ technology.core:
...
+  # Black box command line option escape hatch
+  options: [] 
+  # Define options
+  defines: []
+  # C compiler options
+  compiler_opts: []
+  # Simulation timescale
+  timescale: null
+  # C source files 
+  csrc: []
```
jwright6323: this is usually simulator-specific, so IMO it should go in the tool plugin namespace

src/hammer-vlsi/defaults.yml
```diff
@@ -608,3 +619,51 @@ technology.core:
...
+  # top_module_config: null
+  input_files: []
+  # Debug option
+  debug: false
+  # Black box command line option escape hatch
+  options: [] 
+  # Define options
+  defines: []
+  # C compiler options
+  compiler_opts: []
```
jwright6323: this should go in the tool plugin

src/hammer-vlsi/defaults.yml
```diff
@@ -608,3 +619,51 @@ technology.core:
...
+
+sim.inputs:
+  # Top RTL module
+  top_module: null
+  # top_module_config: null
+  input_files: []
+  # Debug option
+  debug: false
+  # Black box command line option escape hatch
+  options: []
```
jwright6323: what does this do?

src/hammer-vlsi/defaults.yml
```diff
@@ -608,3 +619,51 @@ technology.core:
...
+
+  # Valid options are:
+  #   "rtl" for rtl level simulation
+  #   "gl-syn" for gate level post-synthesis simulation
+  #   "gl-par" for gate level post-par simulation
+  level: "rtl"
+
+  gl-syn:
+    input_files: []
+    post_synth_sdc: null
```
jwright6323: can we just call this `sdc`? If it's post-PNR this name doesn't make sense

src/hammer-vlsi/defaults.yml
```diff
@@ -608,3 +619,51 @@ technology.core:
...
+    time_annotated: false
```
jwright6323:
```suggestion
    timing_annotated: false
```

src/hammer-tech/filters.py
```diff
@@ -167,6 +167,21 @@ def sort_func(lib: "Library"):
...
+        """
+
+        def filter_func(lib: "Library") -> bool:
+            return lib.verilog_sim is not None
+
+        def paths_func(lib: "Library") -> List[str]:
+            assert lib.verilog_sim is not None
+            return [lib.verilog_sim]
+
+        return LibraryFilter.new("verilog_sim", "VSIM gate level sim verilog sources", is_file=True, filter_func=filter_func, paths_func=paths_func)
```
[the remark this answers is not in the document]
jwright6323: I'm not sure I understand why this is a library filter- I was thinking this would be implemented using the inputs/output interface like how syn passes the mapped netlist to PAR

src/hammer-vlsi/generate_properties.py
```diff
@@ -119,7 +119,11 @@ def main(args) -> int:
                                         InterfaceVar("output_files", "List[str]",
                                                      "output collection of mapped (post-synthesis) RTL files"),
                                         InterfaceVar("output_sdc", "str",
-                                                     "(optional) output post-synthesis SDC constraints file")
+                                                     "(optional) output post-synthesis SDC constraints file"),
+                                        InterfaceVar("output_access", "str",
+                                                     "output access file for gate level simulation"),
+                                        InterfaceVar("output_force_regs", "str",
```
jwright6323: see other comment, this should be the JSON

src/hammer-vlsi/generate_properties.py
```diff
@@ -119,7 +119,11 @@ def main(args) -> int:
...
+                                                     "output force regs tcl file for gate level simulation register initialization")
                                         # TODO: model CAD junk
```
jwright6323: this is missing a few outputs we probably want: SAIF for activity factors (maybe call it `activity_file`, VCD/VPD (maybe call it `waveform_file`)

src/hammer-vlsi/defaults.yml
```diff
@@ -608,3 +619,99 @@ technology.core:
...
+  compiler_opts: []
+
+  # Simulation timescale
+  # type: str
+  # This is passed as a plain string
+  # For example, for VCS, timescale may be set as "sim.inputs.timescale: '1ns/10ps' which would be passed as "-timescale=1ns/10ps" 
+  timescale: null
+
+  # C source files passed when generatin the simulation executable
+  # type: List[str[
```
```suggestion
  # type: List[str]
```

edwardcwang:
Unit tests wise - I would suggest taking a look at what John did for the SubmitCommand test or the mockpar stuff for inspiration.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -941,13 +981,112 @@ def ilms(self, value: List[ILMStruct]) -> None:
...
+
+
 class HasUPFSupport(HammerTool):
     """Mix-in trait with functions useful for tools with UPF style power
     constraints"""
     @property
     def upf_power_specification(self) -> str:
         raise NotImplementedError("Automatic generation of UPF power specifications is not supported yet.")
 
+
```
jwright6323: random newline

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -994,7 +1133,6 @@ def cpf_power_specification(self) -> str:
 
         return "\n".join(output)
 
-
```
jwright6323: random newline deleted

src/hammer-tech/filters.py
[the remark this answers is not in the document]
jwright6323: Talked about this offline, the library filter is fine

src/hammer-vlsi/defaults.yml
dpgrubb13: Done

src/hammer-vlsi/hammer_vlsi/cli_driver.py
```diff
@@ -177,7 +186,13 @@ def action_map(self) -> Dict[str, CLIActionType]:
...
             "par-to-lvs": self.par_to_lvs_action,
             "drc": self.drc_action,
-            "lvs": self.lvs_action
+            "lvs": self.lvs_action,
+            "sim": self.sim_action,
+            "simulation": self.sim_action,
+            "synthesis_to_sim": self.synthesis_to_sim_action,
+            "synthesis-to-sim": self.synthesis_to_sim_action,
+            "syn_to_sim": self.synthesis_to_sim_action,
+            "syn-to-sim": self.synthesis_to_sim_action
```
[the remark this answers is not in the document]
dpgrubb13: I agree that it's pretty similar, but I think it makes sense to just put it into the next PR at this point. It's trivial either way once this is in place.

src/hammer-vlsi/defaults.yml
dpgrubb13: More comments were added to the defaults.yml since.  The debug flag enable vcd generation and the options flag is for general black box options that will be passed directly as a flag when compiling the test executable as an exact string.

src/hammer-vlsi/defaults.yml
dpgrubb13: This is removed since it's not actually being used.  I decided to remove the `gl-syn` section because options and input_files inputs can be shared between the different sim levels, and the sim mode is moved to the `level` flag.

src/hammer-vlsi/defaults.yml
dpgrubb13: I agree with this.

src/hammer-vlsi/generate_properties.py
dpgrubb13: I'm not sure that we need to pass the VCD/VPD files downstream (feel free to correct me).  I agree though about adding some outputs.  We will need to pass SDF files and SPF files for different sims.

src/hammer-vlsi/hammer_vlsi/cli_driver.py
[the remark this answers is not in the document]
jwright6323: sounds good.

src/hammer-vlsi/generate_properties.py
jwright6323: The VCD files are consumed by the dynamic power analysis tools. We can add them later.

jwright6323:
@dpgrubb13 the unit tests should pass now. I read through the code a bit as I was fixing things. Here are a few action items I'd like you to fix before we merge this.

- [ ] There are no equivalent interface vars for synthesis (`output_seq_cells`, `output_all_regs`, `output_sdf`)
- [ ] There are no`par_to_sim` or `par-to-sim` actions despite there being a method to do so- need to add those to the action map like you did `syn_to_par` and `syn-to-par`
- [ ] The output SDF file should have some deterministic name for both `par` and `syn`- right now I've just set it to `""` for `par`. This should probably be a hammer IR setting (basically make it look like `post_synth_sdc`)
- [ ] `output_all_regs` and `output_seq_cells` need a description somewhere of how you should generate them.
- [ ] Lines 998-1009 should be uncommented (and fixed if buggy)

harrisonliew:
Decided to merge and push any remaining unaddressed comments to future PRs.

Issue #280, opened 2018-12-30
Port parameter of clock constraints is not fully vendor-independent


edwardcwang:
The syntax `pin:<path>` may not be vendor-independent: better to do just `<path>` and have each tool plugin translate.

```
vlsi.inputs.clocks:
- name: "digital_system_clock"
  generated: true
  port: "pin:MY_TOP/modA/modB/my_macro/clk_out"
  period: "100 ns"
  uncertainty: "1 ns"
```

edwardcwang:
Related to metrics

edwardcwang:
Turns out that even with SDC each tool may have different ways of referring to nets/pins, so this is slightly trickier than expected.

PR #331, opened 2019-03-05
Add metrics traits

jwright6323:
This is a draft of the metrics mix-in traits.


jwright6323:
- [ ] Also solve #280 here

colinschmidt:
I think it might be better to break the fix for #280 out from a new metrics API. Just from a CR perspective I can imagine it getting confusing.

edwardcwang:
I agree with @colinschmidt here - in general for CI/CR it's nicer to have lots of small PRs than big PRs. (Sometimes it can't be helped though)

PR #345, opened 2019-04-05
Handle None case for generated flag

abejgonzalez:


edwardcwang:
LGTM. I filed #346 for the longer term fix

Issue #184, opened 2018-09-07
ILM Terminology


jwright6323:
"ILM" is a strictly Cadence term- it seems the abstraction is being broken a bit here.

https://github.com/ucb-bar/hammer/blob/b8ef3a9a12275a77df7a989e6728771112e5f531/src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py#L233

colinschmidt:
This is easy to do once we come up with the right name

jwright6323:
Going to do this with #242 

Any name preferences? I propose `hier_output_modules` or `hierarchical_output_modules`

colinschmidt:
This wasn't fixed by #242.
I think `hierarchical_output_modules` is good.

edwardcwang:
An ILM is more than just a module name right? Isn't it more like a block?

colinschmidt:
I think of it as a hardened module. So this is in reference to all the collateral needed to use that hardened module somewhere else.
Do you have an alternative name? I do think anything suggested so far is really good yet just better than ilm.

edwardcwang:
maybe output_hierarchical_blocks? And `HierarchicalBlock` for the associated class

https://github.com/ucb-bar/hammer/blob/8997d176042e2ad3273e5d5bce37ccc75a687892/src/hammer-vlsi/hammer_vlsi/constraints.py#L23

colinschmidt:
Thats also fine by me.

PR #349, opened 2019-04-11
Add supply and cpf api

colinschmidt:
Metadata: rebase and merge

Add the ability to generate a CPF power spec.
This also includes a simple supply net API to enable the CPF generation.
This only supports single domain power specifications for now.

I also threw in a free refactor (for that negative line count despite adding a feature achievement) 

I tested this on a downstream template, it works as before not breaking SYN, PAR, or LVS.

### DIFF
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -108,10 +108,21 @@ cadence:
 # General VLSI inputs.
 # These will vary per run of hammer-vlsi.
 vlsi.inputs:
-  # Supply voltages.
+  # Supply voltages and names
+  # Supply Struct:
+  #   name (str) - The name of your supply's net
+  #   pin (Optional[str]) - The cell pin that this net should drive, no pins are connected when omitted
+  #   tie (Optional[str]) - Another supply this supply is tied to. Nets with this set will  not end up in your final layout.
+  # power (List(Supply)): A list of of all power net(s) in the design
+  # ground (List(Supply)): A list of all ground net(s) in the design
+  # VDD (str): The default voltage of the primary power net
+  # GND (str): The default voltage of the primary ground net
   supplies:
+    power: [{name: "VDD", pin: "VDD"}]
+    ground: [{name: "VSS", pin: "VSS"}]
     VDD: "0.85 V"
     GND: "0 V"
+
 
   hierarchical:
     # Hierarchical par mode. (str)
--- a/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
+++ b/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
@@ -317,6 +317,25 @@ class HammerTool(metaclass=ABCMeta):
         self._submit_command = value
 
     @property
+    def top_module(self) -> str:
+        """
+        Get the top-level module.
+
+        :return: The top-level module.
+        """
+        try:
+            return self.attr_getter("_top_module", None)
+        except AttributeError:
+            raise ValueError("Nothing set for the top-level module yet")
+
+    @top_module.setter
+    def top_module(self, value: str) -> None:
+        """Set the top-level module."""
+        if not (isinstance(value, str)):
+            raise TypeError("top_module must be a str")
+        self.attr_setter("_top_module", value)
+
+    @property
     def logger(self) -> HammerVLSILoggingContext:
         """Get the logger for this tool."""
         try:
--- a/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
+++ b/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
@@ -846,6 +846,30 @@ class HammerTool(metaclass=ABCMeta):
             output.append(clock)
         return output
 
+    def get_all_supplies(self, key: str) -> List[Supply]:
+        supplies = self.get_setting(key)
+        output = []  # type: List[Supply]
+        for raw_supply in supplies:
+            supply = Supply(name=raw_supply['name'],pin=None,tie=None)
+            if 'pin' in raw_supply:
+                supply = supply._replace(pin=raw_supply['pin'])
+            if 'tie' in raw_supply:
+                supply = supply._replace(tie=raw_supply['tie'])
+            output.append(supply)
+        return output
+
+    def get_all_power_nets(self) -> List[Supply]:
+        return self.get_all_supplies("vlsi.inputs.supplies.power")
+
+    def get_independent_power_nets(self) -> List[Supply]:
+        return list(filter(lambda x: x.tie is None, self.get_all_power_nets()))
+
+    def get_all_ground_nets(self) -> List[Supply]:
+        return self.get_all_supplies("vlsi.inputs.supplies.ground")
+
+    def get_independent_ground_nets(self) -> List[Supply]:
+        return list(filter(lambda x: x.tie is None, self.get_all_ground_nets()))
+
     def get_gds_map_file(self) -> Optional[str]:
         """
         Get a GDS map in accordance with settings in the Hammer IR.
--- a/src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
+++ b/src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
@@ -21,6 +21,7 @@ from hammer_utils import reverse_dict, deepdict, optional_map, get_or_else, add_
 from hammer_tech import Library, ExtraLibrary
 
 from .constraints import *
+from .units import VoltageValue
 
 
 class HierarchicalMode(Enum):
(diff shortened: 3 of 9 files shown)


src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -846,6 +865,32 @@ def get_clock_ports(self) -> List[ClockPort]:
             output.append(clock)
         return output
 
+    def get_all_supplies(self, key: str) -> List[Supply]:
+        supplies = self.get_setting(key)
+        output = [] # type: List[Supply]
+        for raw_supply in supplies:
+            print("raw_supply:")
+            print(raw_supply)
```
edwardcwang: debug lines?

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -846,6 +865,32 @@ def get_clock_ports(self) -> List[ClockPort]:
             output.append(clock)
         return output
 
+    def get_all_supplies(self, key: str) -> List[Supply]:
+        supplies = self.get_setting(key)
+        output = [] # type: List[Supply]
```
```suggestion
        output = []  # type: List[Supply]
```

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -846,6 +865,32 @@ def get_clock_ports(self) -> List[ClockPort]:
...
+            if 'tie' in raw_supply:
+                supply = supply._replace(tie=raw_supply['tie'])
+            output.append(supply)
+        return output
+
+    def get_all_power_nets(self) -> List[Supply]:
+        return self.get_all_supplies("vlsi.inputs.supplies.power")
+
+    def get_independent_power_nets(self) -> List[Supply]:
+        return list(filter(lambda x: x.tie == None, self.get_all_power_nets()))
```
```suggestion
        return list(filter(lambda x: x.tie is None, self.get_all_power_nets()))
```

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -846,6 +865,32 @@ def get_clock_ports(self) -> List[ClockPort]:
...
+
+    def get_all_ground_nets(self) -> List[Supply]:
+        return self.get_all_supplies("vlsi.inputs.supplies.ground")
+
+    def get_independent_ground_nets(self) -> List[Supply]:
+        return list(filter(lambda x: x.tie == None, self.get_all_ground_nets()))
```
```suggestion
        return list(filter(lambda x: x.tie is None, self.get_all_ground_nets()))
```

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -1201,7 +1062,23 @@ def generate_power_spec_commands(self) -> List[str]:
             self.logger.error(
                 "Invalid power specification type '{tpe}'; only 'cpf' or 'upf' supported".format(tpe=power_spec_type))
             return []
-        power_spec_contents = str(self.get_setting("vlsi.inputs.power_spec_contents"))  # type: str
+
+        power_spec_contents = "" # type: str
```
edwardcwang:
```suggestion
        power_spec_contents = ""  # type: str
```

colinschmidt:
Fixing style nits and force pushing since I want both commits to land as is.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
colinschmidt: removed

edwardcwang:
LGTM pending other comments

src/hammer-vlsi/defaults.yml
```diff
@@ -108,11 +108,22 @@ cadence:
...
+  #   pin (Optional[str]) - The cell pin that this net should drive, no pins are connected when omitted
+  #   tie (Optional[str]) - Another supply this supply is tied to. Nets with this set will  not end up in your final layout.
+  # power (List(Supply)): A list of of all power net(s) in the design
+  # ground (List(Supply)): A list of all ground net(s) in the design
+  # VDD (str): The default voltage of the primary power net
+  # GND (str): The default voltage of the primary ground net
   supplies:
+    power: [{name: "VDD", pin: "VDD"}]
+    ground: [{name: "VSS", pin: "VSS"}]
     VDD: "0.85 V"
```
jwright6323: I think this is beyond the scope of this PR, but we should file an issue to move the nominal voltages into the object you created above.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -862,6 +682,59 @@ def hcells_list(self, value: List[str]) -> None:
...
     ### END Generated interface HammerLVSTool ###
 
+class HasUPFSupport(HammerTool):
+    """Mix-in trait with functions useful for tools with UPF style power
+    constraints"""
+    @property
+    def upf_power_specification(self) -> str:
+        raise NotImplementedError("Automatic generation of UPF power specifications is not supported yet.")
+
+class HasCPFSupport(HammerTool):
```
jwright6323: No changes requested here, but adding this to #184 as a list of things we want to eventually move. CPF is Cadence-specific and IMO should live with the rest of the Cadence stuff.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -927,7 +800,7 @@ def post_synth_sdc(self) -> Optional[str]:
         pass
 
 
-class CadenceTool(HasSDCSupport, HammerTool):
+class CadenceTool(HasSDCSupport, HasCPFSupport, HasUPFSupport, HammerTool):
```
jwright6323: Why `HasUPFSupport`?

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -1201,7 +1062,23 @@ def generate_power_spec_commands(self) -> List[str]:
...
                 "Invalid power specification type '{tpe}'; only 'cpf' or 'upf' supported".format(tpe=power_spec_type))
             return []
-        power_spec_contents = str(self.get_setting("vlsi.inputs.power_spec_contents"))  # type: str
+
+        power_spec_contents = ""  # type: str
+        power_spec_mode = str(self.get_setting("vlsi.inputs.power_spec_mode"))  # type: str
+        if power_spec_mode == "empty":
+            return []
+        elif power_spec_mode == "auto":
+            if power_spec_type == "cpf":
```
jwright6323: I see a world where this is unnecessary, and the correct method is called based on what PNR tool you're using, with overrides if desired. I'll file an issue here too.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -1201,7 +1062,23 @@ def generate_power_spec_commands(self) -> List[str]:
             self.logger.error(
                 "Invalid power specification type '{tpe}'; only 'cpf' or 'upf' supported".format(tpe=power_spec_type))
             return []
-        power_spec_contents = str(self.get_setting("vlsi.inputs.power_spec_contents"))  # type: str
+
+        power_spec_contents = ""  # type: str
+        power_spec_mode = str(self.get_setting("vlsi.inputs.power_spec_mode"))  # type: str
+        if power_spec_mode == "empty":
+            return []
+        elif power_spec_mode == "auto":
```
jwright6323: Did we settle on the difference between `auto` and `generate`?

src/hammer-vlsi/defaults.yml
colinschmidt: Ok. An issue sounds good.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
colinschmidt: There is a spec that is available online. So I think its a bit different than the ILM stuff. https://projects.si2.org/?page=811

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
colinschmidt: The cadence tools do support UPF, so this is future proofing for when someone implements the generation for it.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
colinschmidt: You can use either cpf or upf for the tool I tested this on. So I thought it best to give the user freedom, if for example one generator produces better output. Perhaps we can have a default per tool for auto.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
colinschmidt: I went with auto because more `*_mode` variables use it. I do think this is closer to `auto` than `generate` assuming `auto` means Hammer does everything and `generate` means you need to setup Hammer to be able to generate.
The default supplies are sane, so its really about the `type` choice being populated automatically that would make it full-auto.

Issue #252, opened 2018-11-01
Bumps API


edwardcwang:
Related to I/O cell API

jwright6323:
I'll pick this one up when I get some free time.

jwright6323:
We'll probably want to work on this soon

edwardcwang:
Fixed thanks to @colinschmidt !

PR #355, opened 2019-04-15
Add an API for placing and assigning bumps

colinschmidt:
This is a new optional API that allows the user to specify
a rectangular set of bumps and assign them to signals,
power/ground, leave them unconnected, or remove them.

### DIFF
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -277,6 +277,22 @@ vlsi.inputs:
   # SRAM Parameters
   sram_parameters: []
 
+  # Bumps settings struct, Optional
+  # Currently only support rectangular, evenly-spaced grids
+  # Bumps are numbered starting with 1,1 in the lower left
+  # Unassigned bumps will be removed, special name "+NC", leaves bump unconnected
+  # setting top-level bummps to null corresponds to a no bump chip
+  bumps: null
+  # BumpsDefinition struct members
+  # - x (int) - number of bumps in the x dimension
+  # - y (int) - number of bumps in the y dimension
+  # - pitch (float) - pitch of bumps in microns
+  # - cell (str) - Name of bump cell
+  # assignments - List of BumpAssignment structs
+  #  - name (str) - The name of the net being assigned to a bump
+  #  - x (int) - The X coordinate of the bump assigned to this net
+  #  - y (int) - The Y coordinate of the bump assigned to this net
+
 vlsi.submit:
   # The submit command to use. "none", "local", or null will run on the current host. See hammer_submit_command.py for other options.
   command: "local"
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -67,6 +67,11 @@ vlsi.technology:
   # Required for some CAD tools to function properly.
   placement_site: null
 
+  # Top cut/via layer for blockage under bumps. (Optional[str])
+  # Only used if using vlsi.inputs.bumps
+  # TODO: remove this after stackup supports vias ucb-bar/hammer#354
+  bump_block_cut_layer: null
+
   # Extra semiconductor IP libraries/stdcell libs/hard macros. (List[ExtraLibrary])
   # Used for non-technology-vendor-provided custom IP blocks.
   # ExtraLibrary is a struct with two fields:
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -282,19 +282,26 @@ vlsi.inputs:
   # SRAM Parameters
   sram_parameters: []
 
-  # Bumps settings struct, Optional
+  # Bumps settings mode. (str)
+  # Specifies how to setup the bumps for your design.
   # Currently only support rectangular, evenly-spaced grids
   # Bumps are numbered starting with 1,1 in the lower left
-  # Unassigned bumps will be removed, special name "+NC", leaves bump unconnected
-  # setting top-level bummps to null corresponds to a no bump chip
+  # Unassigned bumps will be removed, bumps with no_connect set will be present but unconnected
+  # Valid options:
+  # manual - Use the bumps struct to create and assign bumps
+  # empty - Do not add any bumps
+  bumps_mode: empty
+  # Bumps settings struct only required if above key is set to manual
   bumps: null
   # BumpsDefinition struct members
   # - x (int) - number of bumps in the x dimension
   # - y (int) - number of bumps in the y dimension
   # - pitch (float) - pitch of bumps in microns
   # - cell (str) - Name of bump cell
-  # assignments - List of BumpAssignment structs
-  #  - name (str) - The name of the net being assigned to a bump
+  # assignments - List of BumpAssignment structs. You must specify one of name or no_connect.
+  # If both are specified the bump will be left unconnected
+  #  - name (Optional[str]) - The name of the net being assigned to a bump
+  #  - no_connect (Optional[bool]) - If True the bump will be present but unconnected
   #  - x (int) - The X coordinate of the bump assigned to this net
   #  - y (int) - The Y coordinate of the bump assigned to this net
 
--- a/src/hammer-vlsi/hammer_vlsi/constraints.py
+++ b/src/hammer-vlsi/hammer_vlsi/constraints.py
@@ -80,7 +80,8 @@ Supply = NamedTuple('Supply', [
 ])
 
 BumpAssignment = NamedTuple('BumpAssignment', [
-    ('name', str),
+    ('name', Optional[str]),
+    ('no_connect', Optional[bool]),
     ('x', int),
     ('y', int)
 ])
--- a/src/hammer-vlsi/hammer_vlsi/constraints.py
+++ b/src/hammer-vlsi/hammer_vlsi/constraints.py
@@ -14,7 +14,8 @@ from typing import Dict, NamedTuple, Optional, List, Any
 from hammer_utils import reverse_dict
 from .units import TimeValue, VoltageValue, TemperatureValue
 
-__all__ = ['ILMStruct', 'SRAMParameters', 'Supply', 'ClockPort', 'OutputLoadConstraint',
+__all__ = ['ILMStruct', 'SRAMParameters', 'Supply', 'BumpAssignment',
+           'BumpsDefinition', 'ClockPort', 'OutputLoadConstraint',
            'DelayConstraint', 'ObstructionType', 'PlacementConstraintType',
            'Margins', 'PlacementConstraint', 'MMMCCornerType', 'MMMCCorner']
 
@@ -76,6 +77,20 @@ Supply = NamedTuple('Supply', [
     ('name', str),
     ('pin', Optional[str]),
     ('tie', Optional[str])
+])
+
+BumpAssignment = NamedTuple('BumpAssignment', [
+    ('name', str),
+    ('x', int),
+    ('y', int)
+])
+
+BumpsDefinition = NamedTuple('BumpsDefinition', [
+    ('x', int),
+    ('y', int),
+    ('pitch', float),
+    ('cell', str),
+    ('assignments', List[BumpAssignment])
 ])
 
 ClockPort = NamedTuple('ClockPort', [
--- a/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
+++ b/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
@@ -896,6 +896,7 @@ class HammerTool(metaclass=ABCMeta):
         elif bumps_mode != "manual":
             self.logger.error("Invalid bumps_mode:{m}, only empty or manual supported. Assuming empty.".format(
                 m=bumps_mode))
+            return None
         assignments = []  # type: List[BumpAssignment]
         for raw_assign in self.get_setting("vlsi.inputs.bumps.assignments"):
             name = None if not "name" in raw_assign else raw_assign["name"]
--- a/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
+++ b/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
@@ -891,7 +891,7 @@ class HammerTool(metaclass=ABCMeta):
 
     def get_bumps(self) -> Optional[BumpsDefinition]:
         bumps = self.get_setting("vlsi.inputs.bumps")
-        if bumps is None or bumps == "null":
+        if bumps is None:
             return None
         assignments = []  # type: List[BumpAssignment]
         for raw_assign in bumps["assignments"]:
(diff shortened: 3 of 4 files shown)


colinschmidt:
This fixes #252

src/hammer-vlsi/defaults.yml
```diff
@@ -277,6 +277,22 @@ vlsi.inputs:
...
+  bumps: null
+  # BumpsDefinition struct members
+  # - x (int) - number of bumps in the x dimension
+  # - y (int) - number of bumps in the y dimension
+  # - pitch (float) - pitch of bumps in microns
+  # - cell (str) - Name of bump cell
+  # assignments - List of BumpAssignment structs
+  #  - name (str) - The name of the net being assigned to a bump
+  #  - x (int) - The X coordinate of the bump assigned to this net
+  #  - y (int) - The Y coordinate of the bump assigned to this net
```
edwardcwang: To avoid using `null` for a subtree of settings since it may be confusing/error-inducing, could we do something more like this instead?
```
  bumps:
    enabled: true # <-- may be inferrable if x*y = 0
    x: 0
    y: 0
    pitch: 0
    cell: ""
    assignments: []
```

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -889,6 +889,17 @@ def get_all_ground_nets(self) -> List[Supply]:
     def get_independent_ground_nets(self) -> List[Supply]:
         return list(filter(lambda x: x.tie is None, self.get_all_ground_nets()))
 
+    def get_bumps(self) -> Optional[BumpsDefinition]:
+        bumps = self.get_setting("vlsi.inputs.bumps")
+        if bumps is None or bumps == "null":
```
jwright6323: Is the string `"null"` possible? I thought we made it so that it is converted to `None`.

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -78,6 +79,20 @@ def from_setting(d: dict) -> "SRAMParameters":
...
 
+BumpAssignment = NamedTuple('BumpAssignment', [
+    ('name', str),
+    ('x', int),
+    ('y', int)
+])
+
+BumpsDefinition = NamedTuple('BumpsDefinition', [
+    ('x', int),
+    ('y', int),
```
jwright6323: Is this total `x` and `y` bump counts? Not clear to me how this is different from the `x` and `y` in `BumpAssignment`

src/hammer-vlsi/defaults.yml
jwright6323: I think I like `null` better than having an additional `enabled` field. Do you have an example of an error it might cause?

src/hammer-vlsi/defaults.yml
```diff
@@ -277,6 +277,22 @@ vlsi.inputs:
...
 
+  # Bumps settings struct, Optional
+  # Currently only support rectangular, evenly-spaced grids
+  # Bumps are numbered starting with 1,1 in the lower left
+  # Unassigned bumps will be removed, special name "+NC", leaves bump unconnected
+  # setting top-level bummps to null corresponds to a no bump chip
+  bumps: null
+  # BumpsDefinition struct members
+  # - x (int) - number of bumps in the x dimension
+  # - y (int) - number of bumps in the y dimension
```
jwright6323: See comment below about naming of `x` and `y` here.

src/hammer-vlsi/defaults.yml
edwardcwang: One case that comes to mind is that I'm not sure how hammer_config behaves when the parent yml has `null`s in its tree and you want to override subparts of it. In general, while currently hammer setting trees get serialized to maps/dicts, I would be hesistant to reply on it. It's also inconsistent with what we have in the rest of Hammer. It also interacts strangely with static checking of the IR in the future which will help prevent errors.

edwardcwang: Here's a more concrete example that yields a `KeyError`:

```python
hammer_config.load_config_from_string("""foobar: {'b': 1}""", is_yaml=True)['foobar']
```

`foobar` is an example of this kind of optional tree and if it is present as a dict, trying to grab it directly via `['foobar']` will cause an error as `hammer_config` will expand the dict.

colinschmidt: I don't quite understand the error that occurs. I used `get_setting` in both the null and populated case and had no issues.
I do think this is a slightly different case than any we have. IMO the solution most similar to other APIs would be to have a `bumps_mode: [auto,none]` key along with the current key, which is still `null` by default but would be gated by the mode key.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
colinschmidt: Yeah you are right I can remove that string check.

colinschmidt:
Maybe it would be helpful to show one example I used to test for the two different `x` and `y`. I think the current encoding is not confusing in that context (earlier I had a more verbose version but refactored to this one):
```
"vlsi.inputs.bumps": {
    "x": 14,
    "y": 14,
    "pitch": 200,
    "cell": "MY_REDACTED_BUMP_CELL",
    "assignments": [
        {"name": "reset", "x": 5, "y": 3},
        {"name": "+NC", "x": 5, "y": 4},
        {"name": "VDD", "x": 2, "y": 1},
        {"name": "VSS", "x": 2, "y": 2},
        {"name": "VDD", "x": 2, "y": 3},
        {"name": "VSS", "x": 2, "y": 4},
        {"name": "VDD", "x": 13, "y": 1},
        {"name": "VSS", "x": 13, "y": 2},
        {"name": "VDD", "x": 13, "y": 3},
        {"name": "VSS", "x": 13, "y": 4},
        {"name": "VDD", "x": 2, "y": 14},
        {"name": "VSS", "x": 2, "y": 13},
        {"name": "VDD", "x": 2, "y": 12},
        {"name": "VSS", "x": 2, "y": 11},
        {"name": "VDD", "x": 13, "y": 14},
        {"name": "VSS", "x": 13, "y": 13},
        {"name": "VDD", "x": 13, "y": 12},
        {"name": "VSS", "x": 13, "y": 11}
    ]
}
```

colinschmidt:
Another question raised by John was whether we should make the `+NC` and explicit API parameter. I guess something like:
```{"name": null, "x": 5, "y": 5, "no_connect": True}```
where `no_connect` is an `Optional[bool]` where None means false. (Note: this is another case of the Optional[bool] problem we had with clock constraints.)

File: src/hammer-vlsi/defaults.yml (hunk 1 above)
Comment (jwright6323) [reply src/hammer-vlsi/defaults.yml]: The `bumps_mode` solution seems fine to me and consistent with the way we've done everything else, and solves @edwardcwang concerns (I think).

File: src/hammer-vlsi/defaults.yml (hunk 4 above)
Comment (jwright6323) [reply src/hammer-vlsi/defaults.yml]: This is fine, no need to change.

**Comment by jwright6323** [comment]
I tend to like the `no_connect` field better than a magic string because it more clearly states to the user what the API is.

**Comment by colinschmidt** [comment]
> I tend to like the `no_connect` field better than a magic string because it more clearly states to the user what the API is.

I think we should then make the name field optional and define what happens if both are set. I propose no_connect overrides name.

**Comment by jwright6323** [comment]
> > I tend to like the `no_connect` field better than a magic string because it more clearly states to the user what the API is.
> 
> I think we should then make the name field optional and define what happens if both are set. I propose no_connect overrides name.

Yes, sounds good. I think `no_connect` overriding `name` is right. Maybe we should emit a hammer warning if both are defined.

Just out of curiosity, what do you "name" the unconnected bump nets right now, or is it even explicit at all? I suppose one option is that `name` could override this default, but I don't really think that's useful.

File: src/hammer-vlsi/hammer_vlsi/constraints.py (hunk 3 above)
Comment (edwardcwang) [reply src/hammer-vlsi/hammer_vlsi/constraints.py]: I think that is the case (`x (int) - number of bumps in the x dimension`).

Maybe we could rename it to `total_x` or something?

File: src/hammer-vlsi/defaults.yml (hunk 1 above)
Comment (edwardcwang) [reply src/hammer-vlsi/defaults.yml]: `bumps_mode` sounds good to me.

To elaborate on the error - I tried with Colin's specific example below and was able to reproduce this type of error:

```python
import hammer_config
mystr = """
"vlsi.inputs.bumps": {
    "x": 14,
    "y": 14,
    "pitch": 200,
    "cell": "MY_REDACTED_BUMP_CELL",
    "assignments": [
        {"name": "reset", "x": 5, "y": 3},
        {"name": "+NC", "x": 5, "y": 4},
        {"name": "VDD", "x": 2, "y": 1},
        {"name": "VSS", "x": 2, "y": 2},
        {"name": "VDD", "x": 2, "y": 3},
        {"name": "VSS", "x": 2, "y": 4},
        {"name": "VDD", "x": 13, "y": 1},
        {"name": "VSS", "x": 13, "y": 2},
        {"name": "VDD", "x": 13, "y": 3},
        {"name": "VSS", "x": 13, "y": 4},
        {"name": "VDD", "x": 2, "y": 14},
        {"name": "VSS", "x": 2, "y": 13},
        {"name": "VDD", "x": 2, "y": 12},
        {"name": "VSS", "x": 2, "y": 11},
        {"name": "VDD", "x": 13, "y": 14},
        {"name": "VSS", "x": 13, "y": 13},
        {"name": "VDD", "x": 13, "y": 12},
        {"name": "VSS", "x": 13, "y": 11}
    ]
}
"""
hammer_config.load_config_from_string(mystr, is_yaml=True)
# {'vlsi.inputs.bumps.x': 14, 'vlsi.inputs.bumps.y': 14, 'vlsi.inputs.bumps.pitch': 200, 'vlsi.inputs.bumps.cell': 'MY_REDACTED_BUMP_CELL', 'vlsi.inputs.bumps.assignments': [{'name': 'reset', 'x': 5, 'y': 3}, {'name': '+NC', 'x': 5, 'y': 4}, {'name': 'VDD', 'x': 2, 'y': 1}, {'name': 'VSS', 'x': 2, 'y': 2}, {'name': 'VDD', 'x': 2, 'y': 3}, {'name': 'VSS', 'x': 2, 'y': 4}, {'name': 'VDD', 'x': 13, 'y': 1}, {'name': 'VSS', 'x': 13, 'y': 2}, {'name': 'VDD', 'x': 13, 'y': 3}, {'name': 'VSS', 'x': 13, 'y': 4}, {'name': 'VDD', 'x': 2, 'y': 14}, {'name': 'VSS', 'x': 2, 'y': 13}, {'name': 'VDD', 'x': 2, 'y': 12}, {'name': 'VSS', 'x': 2, 'y': 11}, {'name': 'VDD', 'x': 13, 'y': 14}, {'name': 'VSS', 'x': 13, 'y': 13}, {'name': 'VDD', 'x': 13, 'y': 12}, {'name': 'VSS', 'x': 13, 'y': 11}], '_config_path': 'unspecified'}
```

As you can see, the dictionary has been parsed by Hammer into a subtree of settings. If I try to grab `vlsi.inputs.bumps` directly I get a KeyError since that key no longer exists:

```
>>> hammer_config.load_config_from_string(mystr, is_yaml=True)["vlsi.inputs.bumps"]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'vlsi.inputs.bumps'
```

src/hammer-vlsi/hammer_vlsi/constraints.py
jwright6323: I actually think `x` and `y` are fine, after thinking about it some more.

src/hammer-vlsi/defaults.yml
colinschmidt: But `get_setting` handles this case correctly so in what case do we explicitly grab the dictionary?
I don't really care about this but just want to understand when this error shows up.

colinschmidt:
Yes I'll add the `no_connect` field and issue a warning when they conflict.

src/hammer-vlsi/defaults.yml
edwardcwang: I'll dig into this a bit more. Not sure off the top of my head why `get_setting` is able to grab the dict correctly. e.g. `bumps["assignments"]` to me looks like it's grabbing `bumps` explicitly as a dict

On a related note, are there any (even simple) tests for this PR?

edwardcwang: Here's a simple test that reproduces the issue:

```python
    def test_bumps(self) -> None:
        """
        Test that HammerTool bump support works.
        """
        import hammer_config

        tech_dir, tech_dir_base = HammerToolTestHelpers.create_tech_dir("dummy28")
        tech_json_filename = os.path.join(tech_dir, "dummy28.tech.json")
        HammerToolTestHelpers.write_tech_json(tech_json_filename)
        tech = self.get_tech(hammer_tech.HammerTechnology.load_from_dir("dummy28", tech_dir))
        tech.cache_dir = tech_dir
        tech.logger = HammerVLSILogging.context("")

        test = DummyTool()
        test.logger = HammerVLSILogging.context("")
        test.run_dir = tempfile.mkdtemp()
        test.technology = tech
        database = hammer_config.HammerDatabase()
        settings = """
{
	"vlsi.inputs.bumps": {
					"x": 14,
					"y": 14,
					"pitch": 200,
					"cell": "MY_REDACTED_BUMP_CELL",
					"assignments": [
						{"name": "reset", "x": 5, "y": 3},
						{"name": "+NC", "x": 5, "y": 4},
						{"name": "VDD", "x": 2, "y": 1},
						{"name": "VSS", "x": 2, "y": 2},
						{"name": "VDD", "x": 2, "y": 3},
						{"name": "VSS", "x": 2, "y": 4},
						{"name": "VDD", "x": 13, "y": 1},
						{"name": "VSS", "x": 13, "y": 2},
						{"name": "VDD", "x": 13, "y": 3},
						{"name": "VSS", "x": 13, "y": 4},
						{"name": "VDD", "x": 2, "y": 14},
						{"name": "VSS", "x": 2, "y": 13},
						{"name": "VDD", "x": 2, "y": 12},
						{"name": "VSS", "x": 2, "y": 11},
						{"name": "VDD", "x": 13, "y": 14},
						{"name": "VSS", "x": 13, "y": 13},
						{"name": "VDD", "x": 13, "y": 12},
						{"name": "VSS", "x": 13, "y": 11}
					]
				}
}
"""
        database.update_project([hammer_config.load_config_from_string(settings, is_yaml=False)])
        test.set_database(database)
        tech.set_database(database)

        print(test.get_bumps())

        # Cleanup
        shutil.rmtree(tech_dir_base)
        shutil.rmtree(test.run_dir)
```

yields

```
======================================================================
ERROR: test_bumps (__main__.HammerToolTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "src/hammer-vlsi/test.py", line 327, in test_bumps
    print(test.get_bumps())
  File "/dev/hammer/src/hammer-vlsi/hammer_vlsi/hammer_tool.py", line 893, in get_bumps
    bumps = self.get_setting("vlsi.inputs.bumps")
  File "/dev/hammer/src/hammer-vlsi/hammer_vlsi/hammer_tool.py", line 725, in get_setting
    return self._database.get_setting(key, nullvalue)
  File "/dev/hammer/src/hammer_config/config_src.py", line 614, in get_setting
    raise KeyError("Key " + key + " is missing")
KeyError: 'Key vlsi.inputs.bumps is missing'

----------------------------------------------------------------------
```

colinschmidt: Thanks for explaining this! Coping this test into my updates.

colinschmidt:
Tests are added.

PR #361, opened 2019-04-17
Top layer placement api

colinschmidt:
This upstreams something I always end up adding in hooks. A lot of hardmacros will need keep-outs around them because their pins cannot handle power grids or standard cells directly adjacent.

This API is intended to be used to implement these keep-outs. The top-layer is optional and no keep-out should be created if it is not specified if it is specified the keep-out will be up and including the layer specified in the key.

### DIFF
--- a/src/hammer-tech/stackup.py
+++ b/src/hammer-tech/stackup.py
@@ -337,3 +337,15 @@ class Stackup(NamedTuple('Stackup', [
             if m.name == name:
                 return m
         raise ValueError("Metal named %s is not defined in stackup %s" % (name, self.name))
+
+    def get_metal_by_index(self, index: int) -> Metal:
+        """
+        Get a given metal layer by index.
+
+        :param index: Index of the metal layer
+        :return: A metal layer object
+        """
+        for m in self.metals:
+            if m.index == index:
+                return m
+        raise ValueError("Metal with index %d is not defined in stackup %s" % (index, self.name))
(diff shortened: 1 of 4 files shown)


edwardcwang:
What's the intended difference between this API and the obstructions/blockages API?

colinschmidt:
I think it will just be painful/error prone to require all hardmacros specify an obstruction that is exactly the right dimensions and position and layers.

edwardcwang:
I'm a little concerned `top_layer` isn't living in the most optimal location it could be. I'm ok with its location for now but we may want to move it to somewhere else to avoid bloating the hardmacro struct too much. (e.g. what to do if the Hammer IR is generated by something else and one wants to retroactively use this API)

Another question: isn't some of this information encoded in the LEF itself? Could it make sense to extract this information from the LEF?

colinschmidt:
I initially considered using the `layers` property to avoid adding a new property to the struct but
1) that seems to add some confusion to what that field meant and
2) would require changing the `layers` field to indexes or do lots of weird parsing in the stackup to find out which layer in the `layers` was the top layer.

You could extract this from a LEF, but you also might want to override it. Some of my hardmacros use upper layers but I only want a keep-out on lower layers.

jwright6323:
Yes, it does live in the LEF and ideally we'd get it from there. This has come up before though, and I think we generally agreed that parsing the LEF is a PITA so we've continued to punt and put the data we actually use in the IR. I think any time we do this we should provides backwards compatibility and disable features and/or error on their use if a new field is missing.

PR #362, opened 2019-04-22
Add simple clock gating api

### LINKED ISSUE #88: Clock-gating API
I think at first it can be a binary switch but sometimes you might need to name which cells you want.
Not confident in what the API would look like past the binary option.

colinschmidt:
Only supports auto, fixes #88

### DIFF
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -328,6 +328,16 @@ synthesis.inputs:
   # Set to null to not specify from the JSON.
   top_module: null
 
+# Syntheis tool settings
+synthesis:
+  # Clock gating mode.
+  # type: str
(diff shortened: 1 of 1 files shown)


src/hammer-vlsi/defaults.yml
```diff
@@ -328,6 +328,16 @@ synthesis.inputs:
   # Set to null to not specify from the JSON.
   top_module: null
 
+# Syntheis tool settings
+synthesis:
+  # Clock gating mode.
+  # type: str
+  # This tells the tool how to handle clock gating.
+  # Valid options:
+  # auto - Turn the tools automatic clock gating inference on.
```
```suggestion
  # auto - Turn on automatic clock gating inference in CAD tools.
```

colinschmidt:
We'll leave manual clock gating until we need/want it.

Issue #318, opened 2019-02-23
Move hammer_tech tests out of hammer-vlsi


jwright6323:
Per offline discussion this is just a historical relic.

edwardcwang:
#369 has made this problem worse as there are now more hammer_tech test support files floating in the `hammer-vlsi` dir.

Issue #231, opened 2018-10-24
DRY hammer-vlsi unit tests


jwright6323:
_Originally posted by @edwardcwang in https://github.com/ucb-bar/hammer/pull/219_

edwardcwang:
Original comment: HammerSubmitCommandTestContext shares duplicate code with HammerToolHooksTestContext. We don't necessarily need to fix it urgently, although it might be nice to consolidate at some point (large amounts of copy pasted code can get hairy as we all know).

colinschmidt:
Yeah I think we are probably gonna want to make several passes at this. I know last time I added tests I had to some refactoring to avoid too much copy pasting.

PR #369, opened 2019-04-26
Add pin assignment API

### LINKED ISSUE #112: pin placement API for floorplan
For each module you place you should be able to specify where the pins go.

Simple API would just be which side(s) but you some tools give you much more detail than that.

colinschmidt:
Fixes #112 and ucb-bar/hammer-cad-plugins#52

Relatively simple but enables user to have a decent amount of control.
The user can select pins and which layers/side they are placed on.
In addition pins associated with macros that should not be placed
can also be handled

There is a unit test and I also ran it on my favorite technology.

### DIFF
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -312,6 +312,25 @@ vlsi.inputs:
     #  - y (int) - The Y coordinate of the bump assigned to this net
     assignments: []
 
+  # Pin placement mode. (str)
+  # Specifies how to arrange pins for your design.
+  # Valid options:
+  # none - This mode lets the CAD tool do whatever it wants (providing no constraints).
+  #   Typically this is sane but unpredictable.
+  # generated - Use the assignments list below
+  # auto - is unsupported but should look at hierarchy and guess where pins should go
+  pin_mode: none
+  pin:
+    # List of PinAssignment Structs.
+    #   You must specify pins and one of either macro or layers and side.
+    # - pins (str) - The name(s) of pins that will be assigned.
+    # - side (Optional[str]) - The side of the chip/block these pins will be palced along.
+    #   One of "left", "right", "top", "bottom"
+    # - layers (Optional[List[str]]) - The metal layers that these pins will be placed on.
+    # - macro (Optional[bool]) - If true this pins come from an internal macro
+    #   and should just be noted as such without placing them.
+    assignments: []
+
 vlsi.submit:
   # The submit command to use. "none", "local", or null will run on the current host. See hammer_submit_command.py for other options.
   command: "local"
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -322,13 +322,16 @@ vlsi.inputs:
   pin_mode: none
   pin:
     # List of PinAssignment Structs.
-    #   You must specify pins and one of either macro or layers and side.
+    #   You must specify pins and one of either preplaced or layers and side.
     # - pins (str) - The name(s) of pins that will be assigned.
     # - side (Optional[str]) - The side of the chip/block these pins will be placed along.
     #   One of "left", "right", "top", "bottom"
     # - layers (Optional[List[str]]) - The metal layers that these pins will be placed on.
-    # - macro (Optional[bool]) - If true this pins come from an internal macro
-    #   and should just be noted as such without placing them.
+    # - preplaced (Optional[bool]) - If true, these pins are preplaced by some internal block or macro.
+    #   This is telling the tool that these pins are on an internal block and I want them
+    #   "placed" where they are currently on the internal block. This can be used for IP
+    #   that directly connects to bumps, or signals that will be connected by abutment,
+    #   and much more.
     assignments: []
 
 vlsi.submit:
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -323,7 +323,9 @@ vlsi.inputs:
   pin:
     # List of PinAssignment Structs.
     #   You must specify pins and one of either preplaced or layers and side.
-    # - pins (str) - The name(s) of pins that will be assigned.
+    # - pins (str) - The name(s) of pins that will be assigned. All current backends support
+    #   * as wildcards in the pin names. So "*" for all pins or "io_tl_a*" for all pins that
+    #   start with "io_tl_a".
     # - side (Optional[str]) - The side of the chip/block these pins will be placed along.
     #   One of "left", "right", "top", "bottom"
     # - layers (Optional[List[str]]) - The metal layers that these pins will be placed on.
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -318,7 +318,7 @@ vlsi.inputs:
   # none - This mode lets the CAD tool do whatever it wants (providing no constraints).
   #   Typically this is sane but unpredictable.
   # generated - Use the assignments list below
-  # auto - is unsupported but should look at hierarchy and guess where pins should go
+  # auto - (not implemented yet) looks at the hierarchy and guess where pins should go
   pin_mode: none
   pin:
     # List of PinAssignment Structs.
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -324,7 +324,7 @@ vlsi.inputs:
     # List of PinAssignment Structs.
     #   You must specify pins and one of either macro or layers and side.
     # - pins (str) - The name(s) of pins that will be assigned.
-    # - side (Optional[str]) - The side of the chip/block these pins will be palced along.
+    # - side (Optional[str]) - The side of the chip/block these pins will be placed along.
     #   One of "left", "right", "top", "bottom"
     # - layers (Optional[List[str]]) - The metal layers that these pins will be placed on.
     # - macro (Optional[bool]) - If true this pins come from an internal macro
--- a/src/hammer-vlsi/hammer_vlsi/constraints.py
+++ b/src/hammer-vlsi/hammer_vlsi/constraints.py
@@ -14,10 +14,11 @@ from typing import Dict, NamedTuple, Optional, List, Any
 from hammer_utils import reverse_dict
 from .units import TimeValue, VoltageValue, TemperatureValue
 
-__all__ = ['ILMStruct', 'SRAMParameters', 'Supply', 'BumpAssignment',
-           'BumpsDefinition', 'ClockPort', 'OutputLoadConstraint',
-           'DelayConstraint', 'ObstructionType', 'PlacementConstraintType',
-           'Margins', 'PlacementConstraint', 'MMMCCornerType', 'MMMCCorner']
+__all__ = ['ILMStruct', 'SRAMParameters', 'Supply', 'PinAssignment',
+           'BumpAssignment', 'BumpsDefinition', 'ClockPort',
+           'OutputLoadConstraint', 'DelayConstraint', 'ObstructionType',
+           'PlacementConstraintType', 'Margins', 'PlacementConstraint',
+           'MMMCCornerType', 'MMMCCorner']
 
 
 # Struct that holds various paths related to ILMs.
@@ -77,6 +78,13 @@ Supply = NamedTuple('Supply', [
     ('name', str),
     ('pin', Optional[str]),
     ('tie', Optional[str])
+])
+
+PinAssignment = NamedTuple('PinAssignment', [
+    ('pins', str),
+    ('side', Optional[str]),
+    ('layers', Optional[List[str]]),
+    ('macro', Optional[bool])
 ])
 
 BumpAssignment = NamedTuple('BumpAssignment', [
--- a/src/hammer-vlsi/hammer_vlsi/constraints.py
+++ b/src/hammer-vlsi/hammer_vlsi/constraints.py
@@ -84,7 +84,7 @@ PinAssignment = NamedTuple('PinAssignment', [
     ('pins', str),
     ('side', Optional[str]),
     ('layers', Optional[List[str]]),
-    ('macro', Optional[bool])
+    ('preplaced', Optional[bool])
 ])
 
 BumpAssignment = NamedTuple('BumpAssignment', [
--- a/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
+++ b/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
@@ -937,15 +937,15 @@ class HammerTool(metaclass=ABCMeta):
             if not (side is None or side == "top" or side == "bottom" or side == "right" or side == "left") :
                 self.logger.warning("Pins {p} have invalid side {s}. Assuming pins will be handled by CAD tool.".format(p=pins, s=side))
                 continue
-            preplaced = False if not "preplaced" in raw_assign else raw_assign["preplaced"]
+            preplaced = raw_assign.get("preplaced", False)
             layers = [] if not "layers" in raw_assign else raw_assign["layers"]
             if preplaced:
-                if len(layers) != 0 or side != None:
+                if len(layers) != 0 or side is not None:
                     self.logger.warning("Pins {p} assigned as a preplaced pin with layers or side. Assuming pins are preplaced pins and ignoring layers and side.".format(p=pins))
                     assigns.append(PinAssignment(pins=pins, side=None, layers=[], preplaced=preplaced))
                     continue
             else:
-                if len(layers) == 0 or side == None:
+                if len(layers) == 0 or side is None:
                     self.logger.warning("Pins {p} assigned without layers or side. Assuming pins will be handled by CAD tool.".format(p=pins))
                     # No pin appended
                     continue
--- a/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
+++ b/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
@@ -17,7 +17,7 @@ from typing import Any, Callable, Dict, Iterable, List, Optional, Set, Tuple, ca
 import hammer_config
 import hammer_tech
 from hammer_logging import HammerVLSILoggingContext
-from hammer_tech import LibraryFilter, Stackup
+from hammer_tech import LibraryFilter, Stackup, RoutingDirection
 from hammer_utils import (add_lists, assert_function_type, get_or_else,
                           optional_map)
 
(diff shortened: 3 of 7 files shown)


src/hammer-vlsi/defaults.yml
```diff
@@ -312,6 +312,25 @@ vlsi.inputs:
     #  - y (int) - The Y coordinate of the bump assigned to this net
     assignments: []
 
+  # Pin placement mode. (str)
+  # Specifies how to arrange pins for your design.
+  # Valid options:
+  # none - This mode lets the CAD tool do whatever it wants (providing no constraints).
+  #   Typically this is sane but unpredictable.
+  # generated - Use the assignments list below
+  # auto - is unsupported but should look at hierarchy and guess where pins should go
```
```suggestion
  # auto - (not implemented yet) looks at the hierarchy and guess where pins should go
```

src/hammer-vlsi/defaults.yml
```diff
@@ -312,6 +312,25 @@ vlsi.inputs:
...
+  pin_mode: none
+  pin:
```
edwardcwang: Should this be `vlsi.inputs.pin` or `vlsi.inputs.pins`? (modulo updating the `mode` above)

src/hammer-vlsi/defaults.yml
```diff
@@ -312,6 +312,25 @@ vlsi.inputs:
...
+    # List of PinAssignment Structs.
+    #   You must specify pins and one of either macro or layers and side.
+    # - pins (str) - The name(s) of pins that will be assigned.
```
edwardcwang: Maybe call this `name`? Not sure how this works if this is multiple pins

src/hammer-vlsi/defaults.yml
```diff
@@ -312,6 +312,25 @@ vlsi.inputs:
...
+    # - side (Optional[str]) - The side of the chip/block these pins will be palced along.
```
```suggestion
    # - side (Optional[str]) - The side of the chip/block these pins will be placed along.
```

src/hammer-vlsi/defaults.yml
```diff
@@ -312,6 +312,25 @@ vlsi.inputs:
...
+    #   One of "left", "right", "top", "bottom"
+    # - layers (Optional[List[str]]) - The metal layers that these pins will be placed on.
+    # - macro (Optional[bool]) - If true this pins come from an internal macro
```
```suggestion
    # - macro (Optional[bool]) - If true, this pin came from an internal macro
```

edwardcwang: I'm also not too sure how this flag works

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -914,6 +914,49 @@ def get_bumps(self) -> Optional[BumpsDefinition]:
...
+            self.logger.error(
+                "Invalid pin_mode {mode}. Using none pin mode.".format(mode=pin_mode))
+            return []
+
+        # Generated pin mode needs to ingest the assignments
+        assigns = []  # type: List[PinAssignment]
+        for raw_assign in self.get_setting("vlsi.inputs.pin.assignments"):
+            pins = str(raw_assign["pins"])  # type: str
+            side = None if not "side" in raw_assign else raw_assign["side"]
+            if not (side == None or side == "top" or side == "bottom" or side == "right" or side == "left") :
```
```suggestion
            if not (side is None or side == "top" or side == "bottom" or side == "right" or side == "left") :
```

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -914,6 +914,49 @@ def get_bumps(self) -> Optional[BumpsDefinition]:
...
+                    continue
+            else:
+                if len(layers) == 0 or side == None:
+                    self.logger.warning("Pins {p} assigned without layers or side. Assuming pins will be handled by CAD tool.".format(p=pins))
+                    # No pin appended
+                    continue
+            stackup = self.get_stackup()
+            for layer in layers:
+                direction = stackup.get_metal(layer).direction
+                if not((direction == RoutingDirection.Horizontal and (side == "left" or side == "right")) or
```
edwardcwang: style

src/hammer-vlsi/tech_test.py
```diff
@@ -11,20 +11,24 @@
 import unittest
 
 from hammer_vlsi import HammerVLSISettings
-from test import HammerToolTestHelpers, DummyTool, HasGetTech
-from typing import Any, Dict, List
+import test
```
edwardcwang: `test` is an internal file so I'm hesitant to import it this way. Did a wildcard import not suffice?

src/hammer-vlsi/tech_test.py
```diff
@@ -11,20 +11,24 @@
...
+from typing import Any, Dict, List, Optional
 
 from hammer_logging import HammerVLSILogging
 import hammer_tech
 from hammer_tech import LibraryFilter, Stackup, Metal, WidthSpacingTuple
 from hammer_utils import deepdict
 
 
-class HammerTechnologyTest(HasGetTech, unittest.TestCase):
+class HammerTechnologyTest(unittest.TestCase):
```
edwardcwang: What happened to mixing in `HasGetTech`?

src/hammer-vlsi/test.py
```diff
@@ -616,6 +619,80 @@ def test_bumps(self) -> None:
...
+         test.set_database(database)
+
+         with HammerLoggingCaptureContext() as c:
+             my_pins = test.get_pin_assignments()
+         self.assertTrue(c.log_contains("Pins bad_side assigned layers "))
+         self.assertTrue(c.log_contains("Pins bad_tx_n assigned as a macro pin with layers"))
+         self.assertTrue(c.log_contains("Pins no_layers assigned without layers"))
+         self.assertTrue(c.log_contains("Pins wrong_side have invalid side"))
+         assert my_pins is not None
+         # Only one of the assignments is invalid so the above 7 becomes 6
```
edwardcwang: Is there a test that captures what happens when all the values are good?

src/hammer-vlsi/defaults.yml
jwright6323: palced -> placed
```suggestion
    # - side (Optional[str]) - The side of the chip/block these pins will be placed along.
```

src/hammer-vlsi/test.py
```diff
@@ -616,6 +619,80 @@ def test_bumps(self) -> None:
          shutil.rmtree(tech_dir_base)
          shutil.rmtree(test.run_dir)
 
+    def test_pins(self) -> None:
```
edwardcwang: I appreciate this comprehensive test.

src/hammer-vlsi/defaults.yml
```diff
@@ -312,6 +312,25 @@ vlsi.inputs:
...
+    #   and should just be noted as such without placing them.
```
jwright6323: The name `macro` is confusing to me- I'm interpreting this as "don't touch because I know what I'm doing," e.g. a supply pin that ends up going out to a bump. Is that right?

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -14,10 +14,11 @@
...
 
-__all__ = ['ILMStruct', 'SRAMParameters', 'Supply', 'BumpAssignment',
-           'BumpsDefinition', 'ClockPort', 'OutputLoadConstraint',
-           'DelayConstraint', 'ObstructionType', 'PlacementConstraintType',
-           'Margins', 'PlacementConstraint', 'MMMCCornerType', 'MMMCCorner']
+__all__ = ['ILMStruct', 'SRAMParameters', 'Supply', 'PinAssignment',
+           'BumpAssignment', 'BumpsDefinition', 'ClockPort',
+           'OutputLoadConstraint', 'DelayConstraint', 'ObstructionType',
+           'PlacementConstraintType', 'Margins', 'PlacementConstraint',
+           'MMMCCornerType', 'MMMCCorner']
```
jwright6323: @edwardcwang Can we delete `__all__`? Is there any reason to hide some things? It's annoying to maintain.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -914,6 +914,49 @@ def get_bumps(self) -> Optional[BumpsDefinition]:
...
 
+    def get_pin_assignments(self) -> List[PinAssignment]:
+        """
+        Get a list of pin assignments in accordance with settings in the Hammer IR.
+        :return: A potentially empty list of PinAssigments.
+        """
+        pin_mode = str(self.get_setting("vlsi.inputs.pin_mode"))  # type: str
+        if pin_mode == "none":
+            return []
+        elif pin_mode != "generated":
```
jwright6323: Most other places in the code we do something like:
```suggestion
        elif pin_mode == "generated":
            pass
        else:
```
I'm not super picky though. Up to you if you want to change it.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -914,6 +914,49 @@ def get_bumps(self) -> Optional[BumpsDefinition]:
...
+                    self.logger.warning("Pins {p} assigned without layers or side. Assuming pins will be handled by CAD tool.".format(p=pins))
+                    # No pin appended
+                    continue
+            stackup = self.get_stackup()
+            for layer in layers:
+                direction = stackup.get_metal(layer).direction
+                if not((direction == RoutingDirection.Horizontal and (side == "left" or side == "right")) or
+                       (direction == RoutingDirection.Vertical and (side == "top" or side == "bottom")) or
+                       (direction == RoutingDirection.Redistribution)):
+                    self.logger.error("Pins {p} assigned layers {l} that do not match the direction of their side {s}. This is very likely to cause issues.".format(p=pins, l=layers, s=side))
```
jwright6323: Does throwing an error cause hammer to exit?

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
jwright6323: Do we want an enum for side? I'm leaning no, but worth bringing up.

src/hammer-vlsi/hammer_vlsi/constraints.py
edwardcwang: In this file pretty much everything in this file is exposed there anyway, so it's fine to do it in this file.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
edwardcwang: That would be `self.logger.fatal` though I think in this case I would favor `error` to allow it to continue in special cases.

src/hammer-vlsi/defaults.yml
jwright6323: Marking this resolved as it is redundant with @edwardcwang's comment

src/hammer-vlsi/defaults.yml
colinschmidt: It is intentionally plural because you can specify multiple pins at once see the test. I can call it names if you would like

src/hammer-vlsi/defaults.yml
colinschmidt: This particular field is singular because the "assignments" are plural

src/hammer-vlsi/defaults.yml
colinschmidt: I think the correct text is "If true, these pins come..."

Will answer the macro question as a full comment since both you and John have questions.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
colinschmidt: This is a complex boolean expression. I wrote it out verbosely so that it is easy to understand the intent.
What is the style convention I should follow?

src/hammer-vlsi/tech_test.py
colinschmidt: Because they import each other you can not use `from`.

src/hammer-vlsi/tech_test.py
colinschmidt: When changing imports it is unable to be used in this manner, despite it clearly being in scope based on the reimplemented `def get_tech` below.
I do not know enough about python multiple inheritance to fix is easily.

src/hammer-vlsi/test.py
colinschmidt: Several of the pins above are "all good". It didn't seem helpful to break out the test cases since no problems are fatal.

src/hammer-vlsi/test.py
colinschmidt: Yeah it took an annoying amount of time because of the stackup not being in a common helper method, causing the import change, breaking mysteriously because of weird python stuff. But it is done and now no one else can break it. :)

src/hammer-vlsi/defaults.yml
colinschmidt: Answering as full reply because Edward has the same question.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
colinschmidt: lol sure whatever floats your python style boat.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
colinschmidt: It seemed excessive to me but I don't much care either way.

colinschmidt:
The macro field is you telling the tool that there is a pin on an internal block that I want "placed" where it currently is in that internal block. I have used this for IP that directly connects to bumps, but you could also use it for things like signals connected by abutment and I'm sure many more.
The reason its called macro is because the pin will be defined on some inner macro/block. Open to naming suggestions.

src/hammer-vlsi/test.py
edwardcwang: Yeah I'm thinking from a future proofing perspective where we add additional IR checks or make the current checks more strict it may be good to have a testcase for "good pin spec".

src/hammer-vlsi/tech_test.py
edwardcwang: So a circular import situation is going on?

colinschmidt: Yes. This would be solved by #231

src/hammer-vlsi/defaults.yml
```diff
@@ -312,6 +312,25 @@ vlsi.inputs:
...
+  # none - This mode lets the CAD tool do whatever it wants (providing no constraints).
+  #   Typically this is sane but unpredictable.
+  # generated - Use the assignments list below
+  # auto - (not implemented yet) looks at the hierarchy and guess where pins should go
+  pin_mode: none
+  pin:
+    # List of PinAssignment Structs.
+    #   You must specify pins and one of either macro or layers and side.
+    # - pins (str) - The name(s) of pins that will be assigned.
+    # - side (Optional[str]) - The side of the chip/block these pins will be placed along.
```
edwardcwang: High level question: how do I control the width/height of the generated pins? And their x/y locations? e.g. let's say we use a floorplanning DSL or other method to floorplan and want to control the physical locations.

src/hammer-vlsi/defaults.yml
edwardcwang: Yeah that sounds good. Maybe name it `names` and say multiple pins can be specified via wildcards.

src/hammer-vlsi/defaults.yml
edwardcwang: Text wise "If true, these pins come..." sounds good

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
edwardcwang: Maybe break them out like

```python
is_horizontal = direction == RoutingDirection.Horizontal and side == "left" or side == "right"
is_vertical = direction == RoutingDirection.Vertical and side == "top" or side == "bottom"
is_redis = direction == RoutingDirection.Redistribution
if not (is_horizontal or is_vertical or is_redis):
  ...
```

edwardcwang:
re. `macro`:

1) How about the name `preplaced` or `internal`?

2) "The macro field is you telling the tool that there is a pin on an internal block that I want "placed" where it currently is in that internal block. I have used this for IP that directly connects to bumps, but you could also use it for things like signals connected by abutment and I'm sure many more." <-- can we add this explanation to the YAML comment/doc?

src/hammer-vlsi/defaults.yml
jwright6323: You don't and shouldn't control the width or the height. These are intended to be conneted by abutment, not via drop.

The x/y locations will come later (I can't recall if @colinschmidt included that in a comment but we discussed offline).

edwardcwang: What's the usecase for this then? A case I encounter is using P&R tools to build a macro that will be ingested at some higher level, and it is critical that we are able to control the width, height, x, and y of the pin. I think even the underlying TCL command support this...

jwright6323: Of course the TCL commands support it; I'm saying you don't need it for this level of API. For a bottom (vertical route) pin, for example, you don't care about the pin height at all. The width is the width of a single route. The actual x position does matter, but not much, so we've decided there are more important issues to tackle than to implement that now.

edwardcwang: How hard would it be to add support for "width" (maybe "depth" but it is less important) and positioning? I think if these two were present this API would become a lot more useful.

jwright6323: You can always write TCL if you need something very specific. I think that width is probably less useful than you seem to be implying. The only reasons I see for adding width are:
1. If you want to do analog routing, but IMO that should inherit its width from the non-default routing rule used to draw it.
2. If you want to do via-drop pins, but I question the usefulness of this outside of a few niches.

I don't think positioning matters enough to gate this PR, but I'll let @colinschmidt speak to that. If you have a specific use case that might help motivate this, please share it.

colinschmidt: It's not hard but I think we have different idea of what this API is for.
The primary use case for this for users like me is to be able to control where the pins of inner digital blocks are.
For example, if I have a block that will be placed adjacent to the chip boundary I can't have any pins going out on that side, so I would constrain the pins to be on other sides.
If the pins are so fixed I presume they would be defined by some IP's LEF and so you could use the `macro` portion of the API to force the inner pins to be outer pins.
In general, digital place and route does not require such specific placement. LEFs are very robust in terms of pin placement definitions so exporting you block as such should allow for easy ingestion in all tools I'm familiar with.

colinschmidt:
Ok I think I've addressed all comments. Thanks for the feedback everyone.

src/hammer-vlsi/defaults.yml
harrisonliew: Just for clarity for future users, a short comment that one can use `*` for wildcarding the pins would be nice.

colinschmidt: Done.

src/hammer-vlsi/defaults.yml
edwardcwang: Deferred to #371

src/hammer-vlsi/tech_test.py
edwardcwang: #318

src/hammer-vlsi/test.py
edwardcwang: Done

src/hammer-vlsi/tech_test.py
edwardcwang: Fixed

PR #370, opened 2019-04-26
Add Site objects and use Decimal for Site and Stackup

### LINKED ISSUE #319: Switch to integer (grid) or fixed-point-based metal calculations
Floats are nasty and constantly calling `self.snap()` in the stackup code is inefficient and annoying.

jwright6323:
Fixes #319. This adds site objects which will be used for hierarchical floorplanning checks in ucb-bar/hammer-cad-plugins#91

This will also finally switch the coordinate system over to integer multiples of the manufacturing grid instead of dealing with floats and rounding everywhere.

### DIFF
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -204,6 +204,56 @@ class ExtraLibrary(NamedTuple('ExtraLibrary', [
         lib_copied.extra_prefixes = extra_prefixes  # type: ignore
         return lib_copied
 
+class Site(NamedTuple('Site', [
+    ('name', str),
+    ('x', float),
+    ('y', float)
+])):
+    """
+    A standard cell site, which is the minimum unit of x and y dimenions a standard cell can have.
+
+    name: The name of this site (usually something like "core")
+    x: The x dimension
+    y: The y dimension
+    """
+    __slots__ = ()
+
+    @property
+    def grid_unit(self) -> float:
+        """
+        Return the manufacturing grid unit.
+
+        TODO: this assumes a manufacturing grid of 0.001
+        """
+        return 0.001
+
+    def snap(self, num: float) -> float:
+        """
+        Snap a number to the grid unit.
+
+        TODO: internally represent numbers as integers or fixed-point to
+        obviate the need for this (see #319).
+
+        :param num: Number to snap
+        :return: Number snapped to grid_unit
+        """
+        return float(round(num / self.grid_unit)) * self.grid_unit
+
+    @staticmethod
+    def from_setting(d: dict) -> "Site":
+        """
+        Return a new Site
+
+        :param d: A dictionary with the keys "name", "x", and "y"
+        :return: A Site
+        """
+        return Site(
+            name=str(d["name"]),
+            x=float(d["x"]),
+            y=float(d["y"])
+        )
+
+
 
 # Struct that holds information about the size of a macro.
 # See defaults.yml.
@@ -806,6 +856,21 @@ class HammerTechnology:
             raise ValueError("Stackup named %s is not defined in tech JSON" % name)
         else:
             raise ValueError("Tech JSON does not specify any stackups")
+
+    def get_site_by_name(self, name: str) -> Site:
+        """
+        Return the site for the given key.
+        """
+        if self.config.sites is not None:
+            for item in list(self.config.sites):
+                if item["name"] == name:
+                    return Site.from_setting(item)
+            raise ValueError("Site named %s is not defined in tech JSON" % name)
+        else:
+            raise ValueError("Tech JSON does not specify any sites")
+
+    def get_placement_site(self) -> Site:
+        return self.get_site_by_name(self.get_setting("vlsi.technology.placement_site"))
 
 
 class HammerTechnologyUtils:
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -230,8 +230,8 @@ class Site(NamedTuple('Site', [
         """
         return Site(
             name=str(d["name"]),
-            x=coerce_to_grid(grid_unit, d["x"]),
-            y=coerce_to_grid(grid_unit, d["y"])
+            x=coerce_to_grid(d["x"], grid_unit),
+            y=coerce_to_grid(d["y"], grid_unit)
         )
 
 
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -11,6 +11,7 @@ import os
 import subprocess
 from abc import ABCMeta, abstractmethod
 from typing import Any, Callable, Iterable, List, NamedTuple, Optional, Tuple
+from decimal import Decimal
 
 import hammer_config
 import python_jsonschema_objects  # type: ignore
@@ -206,8 +207,8 @@ class ExtraLibrary(NamedTuple('ExtraLibrary', [
 
 class Site(NamedTuple('Site', [
     ('name', str),
-    ('x', float),
-    ('y', float)
+    ('x', Decimal),
+    ('y', Decimal)
 ])):
     """
     A standard cell site, which is the minimum unit of x and y dimenions a standard cell can have.
@@ -218,39 +219,19 @@ class Site(NamedTuple('Site', [
     """
     __slots__ = ()
 
-    @property
-    def grid_unit(self) -> float:
-        """
-        Return the manufacturing grid unit.
-
-        TODO: this assumes a manufacturing grid of 0.001
-        """
-        return 0.001
-
-    def snap(self, num: float) -> float:
-        """
-        Snap a number to the grid unit.
-
-        TODO: internally represent numbers as integers or fixed-point to
-        obviate the need for this (see #319).
-
-        :param num: Number to snap
-        :return: Number snapped to grid_unit
-        """
-        return float(round(num / self.grid_unit)) * self.grid_unit
-
     @staticmethod
-    def from_setting(d: dict) -> "Site":
+    def from_setting(grid_unit: Decimal, d: dict) -> "Site":
         """
         Return a new Site
 
+        :param grid_unit: The manufacturing grid unit in nm
         :param d: A dictionary with the keys "name", "x", and "y"
         :return: A Site
         """
         return Site(
             name=str(d["name"]),
-            x=float(d["x"]),
-            y=float(d["y"])
+            x=coerce_to_grid(grid_unit, d["x"]),
+            y=coerce_to_grid(grid_unit, d["y"])
         )
 
 
@@ -852,10 +833,19 @@ class HammerTechnology:
         if self.config.stackups is not None:
             for item in list(self.config.stackups):
                 if item["name"] == name:
-                    return Stackup.from_setting(item)
+                    return Stackup.from_setting(self.get_grid_unit(), item)
             raise ValueError("Stackup named %s is not defined in tech JSON" % name)
         else:
             raise ValueError("Tech JSON does not specify any stackups")
+
+    def get_grid_unit(self) -> Decimal:
+        """
+        Return the manufacturing grid unit.
+        """
+        if self.config.grid_unit is not None:
+            return Decimal(self.config.grid_unit)
+        else:
+            raise ValueError("Tech JSON does not specify a manufacturing grid unit")
 
     def get_site_by_name(self, name: str) -> Site:
         """
@@ -864,7 +854,7 @@ class HammerTechnology:
         if self.config.sites is not None:
             for item in list(self.config.sites):
                 if item["name"] == name:
-                    return Site.from_setting(item)
+                    return Site.from_setting(self.get_grid_unit(), item)
             raise ValueError("Site named %s is not defined in tech JSON" % name)
         else:
             raise ValueError("Tech JSON does not specify any sites")
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -220,7 +220,7 @@ class Site(NamedTuple('Site', [
     __slots__ = ()
 
     @staticmethod
-    def from_setting(grid_unit: Decimal, d: dict) -> "Site":
+    def from_setting(grid_unit: Decimal, d: Dict[str, Any]) -> "Site":
         """
         Return a new Site
 
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -213,7 +213,7 @@ class Site(NamedTuple('Site', [
     """
     A standard cell site, which is the minimum unit of x and y dimensions a standard cell can have.
 
-    name: The name of this site (usually something like "core")
+    name: The name of this site (often something like "core") as defined in the tech and standard cell LEFs
     x: The x dimension
     y: The y dimension
     """
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -10,7 +10,7 @@ import json
 import os
 import subprocess
 from abc import ABCMeta, abstractmethod
-from typing import Any, Callable, Iterable, List, NamedTuple, Optional, Tuple
+from typing import Any, Callable, Iterable, List, NamedTuple, Optional, Tuple. Dict
 from decimal import Decimal
 
 import hammer_config
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -10,7 +10,7 @@ import json
 import os
 import subprocess
 from abc import ABCMeta, abstractmethod
-from typing import Any, Callable, Iterable, List, NamedTuple, Optional, Tuple. Dict
+from typing import Any, Callable, Iterable, List, NamedTuple, Optional, Tuple, Dict
 from decimal import Decimal
 
 import hammer_config
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -211,7 +211,7 @@ class Site(NamedTuple('Site', [
     ('y', Decimal)
 ])):
     """
-    A standard cell site, which is the minimum unit of x and y dimenions a standard cell can have.
+    A standard cell site, which is the minimum unit of x and y dimensions a standard cell can have.
 
     name: The name of this site (usually something like "core")
     x: The x dimension
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -860,6 +860,9 @@ class HammerTechnology:
             raise ValueError("Tech JSON does not specify any sites")
 
     def get_placement_site(self) -> Site:
+        """
+        Return the default placement site defined by the hammer setting "vlsi.technology.placement_site"
+        """
         return self.get_site_by_name(self.get_setting("vlsi.technology.placement_site"))
 
 
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -20,7 +20,7 @@ from hammer_config import load_yaml
 from hammer_logging import HammerVLSILoggingContext
 from hammer_utils import (LEFUtils, add_lists, deeplist, get_or_else,
                           in_place_unique, optional_map, reduce_list_str,
-                          reduce_named)
+                          reduce_named, coerce_to_grid)
 
 from library_filter import LibraryFilter
 from filters import LibraryFilterHolder
--- a/src/hammer-tech/schema.json
+++ b/src/hammer-tech/schema.json
@@ -182,6 +182,25 @@
         "additionalproperties" : false
       }
     },
+    "sites" : {
+      "type" : "array",
+      "items" : {
+        "title" : "Site",
+        "type" : "object",
+        "additionalproperties" : false,
+        "properties" : {
+          "name" : {
+            "type" : "string"
+          },
+          "x" : {
+            "type" : "number"
+          },
+          "y" : {
+            "type" : "number"
+          }
+        }
+      }
+    },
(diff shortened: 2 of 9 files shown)


src/hammer-tech/hammer_tech.py
```diff
@@ -204,6 +205,36 @@ def store_into_library(self) -> Library:
...
+    A standard cell site, which is the minimum unit of x and y dimenions a standard cell can have.
+
+    name: The name of this site (usually something like "core")
+    x: The x dimension
+    y: The y dimension
+    """
+    __slots__ = ()
+
+    @staticmethod
+    def from_setting(grid_unit: Decimal, d: dict) -> "Site":
```
edwardcwang:
```suggestion
    def from_setting(grid_unit: Decimal, d: Dict[str, Any]) -> "Site":
```

jwright6323:
Also fixes #319

src/hammer-tech/hammer_tech.py
```diff
@@ -802,11 +833,35 @@ def get_stackup_by_name(self, name: str) -> Stackup:
...
+        """
+        if self.config.sites is not None:
+            for item in list(self.config.sites):
+                if item["name"] == name:
+                    return Site.from_setting(self.get_grid_unit(), item)
+            raise ValueError("Site named %s is not defined in tech JSON" % name)
+        else:
+            raise ValueError("Tech JSON does not specify any sites")
+
+    def get_placement_site(self) -> Site:
```
edwardcwang: pydoc?

src/hammer-tech/hammer_tech.py
```diff
@@ -204,6 +205,36 @@ def store_into_library(self) -> Library:
         lib_copied.extra_prefixes = extra_prefixes  # type: ignore
         return lib_copied
 
+class Site(NamedTuple('Site', [
+    ('name', str),
+    ('x', Decimal),
+    ('y', Decimal)
+])):
+    """
+    A standard cell site, which is the minimum unit of x and y dimenions a standard cell can have.
```
edwardcwang:
```suggestion
    A standard cell site, which is the minimum unit of x and y dimensions a standard cell can have.
```

jwright6323:
This is ready for a review, I think. Wait to merge until I finish the cad-plugins partner to this and update the relevant tech plugins.

src/hammer-vlsi/hammer_utils/__init__.py
```diff
@@ -101,6 +97,54 @@ def in_place_unique(items: List[Any]) -> None:
             i += 1
 
 
+def coerce_to_grid(num: float, grid: Decimal) -> Decimal:
```
edwardcwang: pydoc?

src/hammer-vlsi/utils_test.py
```diff
@@ -49,6 +51,29 @@ def test_optional_map(self) -> None:
         self.assertNotEqual(optional_map("88", str_to_num), "880")
         self.assertEqual(optional_map("42", str_to_num), 420)
 
+    def test_gcd(self) -> None:
```
edwardcwang: Good to have tests :+1:

jwright6323:
Hmm, merging master broke the build. Let me look into this.

src/hammer-vlsi/hammer_utils/__init__.py
jwright6323: Added

src/hammer-tech/hammer_tech.py
jwright6323: added

src/hammer-tech/stackup.py
```diff
@@ -71,41 +73,44 @@ class WidthSpacingTuple(NamedTuple('WidthSpacingTuple', [
...
-        current_spacing = 0.0
+        current_spacing = Decimal(0)
         for wst in out:
             assert wst.min_spacing >= current_spacing
             current_spacing = wst.min_spacing
         return out
 
 
 class Metal(NamedTuple('Metal', [
+        ('grid_unit', Decimal),
```
edwardcwang: 1) Would the manufacturing grid ever vary per layer?

2) Maybe just to make it clear that the `grid_unit` is a higher-level parameter and not part of the data structure of  `Metal` itself, let's add a comment around it and say that this is not a part of the dict?

src/hammer-vlsi/defaults.yml
```diff
@@ -449,6 +454,11 @@ par:
       # Overrideable by appending _<layer name>
       track_width: 5
 
+      # The first track to contain a power stripe
+      # Overrideable by appending _<layer name>
+      # TODO(johnwright) include an auto-center option
```
```suggestion
      # TODO(johnwright): include an auto-center option
```

src/hammer-vlsi/utils_test.py
```diff
@@ -49,6 +51,43 @@ def test_optional_map(self) -> None:
         self.assertNotEqual(optional_map("88", str_to_num), "880")
         self.assertEqual(optional_map("42", str_to_num), 420)
 
+    def test_coerce_to_grid(self) -> None:
+        self.assertEqual(coerce_to_grid(1.23, Decimal("0.1")), Decimal("1.2"))
+        self.assertEqual(coerce_to_grid(1.23, Decimal("0.01")), Decimal("1.23"))
```
edwardcwang:
```suggestion
        self.assertEqual(coerce_to_grid(1.23, Decimal("0.01")), Decimal("1.23"))
        self.assertEqual(coerce_to_grid(1.227, Decimal("0.01")), Decimal("1.23"))
```

src/hammer-tech/stackup.py
```diff
@@ -121,48 +126,33 @@ class Metal(NamedTuple('Metal', [
...
         """
-        Snap a number to the grid unit.
-
-        TODO: internally represent numbers as integers or fixed-point to
-        obviate the need for this (see #319).
+        Return a Metal object from a dict with keys "name", "index", "direction", "min_width", "pitch", "offset", and "power_strap_widths_and_spacings"
 
-        :param num: Number to snap
-        :return: Number snapped to grid_unit
+        :param grid_unit: The manufacturing grid unit in nm
```
edwardcwang: Did we decide on `nm` or `um`? IIRC many of my CAD scripts units are in `um`. Maybe we should make use of `hammer_units`

src/hammer-tech/stackup.py
jwright6323: Yes, it's uncommon, but I have worked in a technology with different manufacturing grids per layer.

jwright6323: That said, the way most of this is currently written doesn't support that.

Re: 2, OK, I can add a comment.

src/hammer-tech/stackup.py
jwright6323: um, this is just a typo.

jwright6323: Hmm `hammer_units` is an interesting idea.

jwright6323: Sorry for the blast of comments on this one. Maybe we can convert to `hammer_units` in a future PR? I have filed #376 to convert all our internal units to decimals- maybe it makes sense to combine these two changes?

edwardcwang: Sounds good; let's file an issue?

src/hammer-tech/stackup.py
edwardcwang: re. technologies with different manufacturing grids - since we are storing a separate grid_unit in each `Metal` object, might be useful to support an optional `grid_unit` definition in the metal layer itself that gets used if present. This could be filed as an issue as well.

src/hammer-tech/stackup.py
jwright6323: Issue is #378

jwright6323:
This is ready to go alongside https://github.com/ucb-bar/hammer-cad-plugins/pull/91

src/hammer-tech/stackup.py
```diff
@@ -71,41 +73,45 @@ class WidthSpacingTuple(NamedTuple('WidthSpacingTuple', [
...
         # Check that spacings increase.
-        current_spacing = 0.0
+        current_spacing = Decimal(0)
         for wst in out:
             assert wst.min_spacing >= current_spacing
             current_spacing = wst.min_spacing
         return out
 
 
 class Metal(NamedTuple('Metal', [
```
edwardcwang: Not to nitpick on this too much: would like to see this written along something like this:

```python
class Metal(NamedTuple('Metal', [
        ('name', str),
        ('index', int),
        ('direction', RoutingDirection),
        ('min_width', Decimal),
        ('pitch', Decimal),
        ('offset', Decimal),
        ('power_strap_widths_and_spacings', List[WidthSpacingTuple]),
        # Note: grid_unit is not currently parsed as part of the Metal data structure!
        # See #379
        ('grid_unit', Decimal)
])):
```

edwardcwang:
Minor note that we may want to refactor the way we pass `grid_unit` into things in the future.

src/hammer-tech/stackup.py
jwright6323: OK, that's fine with me, I'll make the change.

jwright6323: done

src/hammer-vlsi/defaults.yml
```diff
@@ -434,11 +434,16 @@ par:
   # If power_straps_mode is 'generate', which method to use.
   # Currently, the valid options are:
   # - by_tracks - Specify the power strap plan per layer in terms of tracks
+  # - by_tracks_hier - Only valid for hierarchical blocks- uses the top-level by_tracks to determine parameters
```
colinschmidt: I thought this was going in a separate PR?

src/hammer-vlsi/hammer_utils/__init__.py
```diff
@@ -101,6 +97,75 @@ def in_place_unique(items: List[Any]) -> None:
...
+    :param num: The number to check
+    :param grid: The decimal grid value on which to check number
+    :return: True if num is on-grid, False otherwise
+    """
+    # TODO(johnwright) Remove float support in ucb-bar/hammer#376
+    num_dec = num if isinstance(num, Decimal) else coerce_to_grid(num, Decimal("0.001"))
+    assert isinstance(num_dec, Decimal)
+    return Decimal(int(num_dec / grid)) == (num_dec / grid)
+
+def gcd(*values: int) -> int:
```
colinschmidt: Is there really not a standard python function for this?

src/hammer-vlsi/hammer_utils/__init__.py
```diff
@@ -101,6 +97,75 @@ def in_place_unique(items: List[Any]) -> None:
...
+    """
+    Return the greatest common divisor of a series of ints
+
+    :param values: The values of which to compute the GCD
+    :return: The GCD
+    """
+    assert len(values) > 0
+    if len(values) == 1:
+        return values[0]
+    elif len(values) == 2:
```
colinschmidt: There is a math.gcd, so maybe use that in the base case here. It seems like a fold operation would work here for the whole thing but not familiar with the python functools stuff.

src/hammer-vlsi/hammer_utils/__init__.py
```diff
@@ -101,6 +97,75 @@ def in_place_unique(items: List[Any]) -> None:
...
+        a = values[0]
+        b = values[1]
+        while b > 0:
+            a, b =  b, a % b
+        return a
+    else:
+        return gcd(values[0], gcd(*values[1:]))
+
+def lcm(*values: int) -> int:
```
colinschmidt: ditto

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -960,6 +960,23 @@ def get_pin_assignments(self) -> List[PinAssignment]:
             assigns.append(PinAssignment(pins=pins, side=side, layers=layers, preplaced=preplaced))
         return assigns
 
+    def get_pin_metals(self) -> List[Metal]:
```
colinschmidt: I'm ok leaving this in, as its an internal function, but as we talked about it this isn't quite useful as is.

src/hammer-vlsi/tech_test.py
```diff
@@ -652,21 +648,21 @@ def test_twwt_wire(self) -> None:
...
 
 
     def test_min_spacing_for_width(self) -> None:
         # Generate multiple stackups, but we'll only use the largest for this test
         stackup = StackupTestHelper.create_test_stackup_list()[-1]
         for m in stackup.metals:
             # generate some widths:
-            for x in [1.0, 2.0, 3.4, 4.5, 5.25, 50.2]:
+            for x in ["1.0", "2.0", "3.4", "4.5", "5.25", "50.2"]:
                 # coerce to a manufacturing grid, this is just a test data point
```
colinschmidt: Why not use the helper function to "coerce"

src/hammer-vlsi/utils_test.py
```diff
@@ -49,6 +51,44 @@ def test_optional_map(self) -> None:
...
+        self.assertEqual(coerce_to_grid(1.227, Decimal("0.01")), Decimal("1.23"))
+        self.assertEqual(coerce_to_grid(200, Decimal("10")), Decimal("200"))
```
colinschmidt: I think we should be setting the decimal precision. Add then add a test for decimals out of precision i.e. "0.0001"

src/hammer-vlsi/utils_test.py
```diff
@@ -49,6 +51,44 @@ def test_optional_map(self) -> None:
         self.assertNotEqual(optional_map("88", str_to_num), "880")
         self.assertEqual(optional_map("42", str_to_num), 420)
 
+    def test_coerce_to_grid(self) -> None:
+        self.assertEqual(coerce_to_grid(1.23, Decimal("0.1")), Decimal("1.2"))
+        self.assertEqual(coerce_to_grid(1.23, Decimal("0.01")), Decimal("1.23"))
+        self.assertEqual(coerce_to_grid(1.227, Decimal("0.01")), Decimal("1.23"))
```
colinschmidt: Can we also check something like `coerce_to_grid(1/3, Decimal("0.001"))`

src/hammer-tech/schema.json
```diff
@@ -2,6 +2,9 @@
   "title" : "TechJSON",
   "type" : "object",
   "properties" : {
+    "grid_unit" : {
+      "type" : "string"
```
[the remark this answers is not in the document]
edwardcwang: Can Decimal parse from string? In the future we probably want to move to hammer_units

src/hammer-vlsi/hammer_utils/__init__.py
edwardcwang: The standard GCD only takes 2 arguments whereas this one takes an unlimited number of arguments. Maybe this one can use the standard gcd under the hood?

https://docs.python.org/3.7/library/fractions.html#fractions.gcd

src/hammer-vlsi/hammer_utils/__init__.py
edwardcwang: Ditto above. Python `reduce()` can double for Scala `reduce` and `fold`

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
edwardcwang: Is it because we don't capture all of the pins?

colinschmidt: So this function is not used in this PR so thats why it doesn't need to be here.
It's not useful as written because you want to be able to get the pin metals of an inner module via some `get_hierarchical_setting("MyInnerModule", "vlsi.inputs.pin.assignements")` that will be made in a forthcoming PR.

src/hammer-tech/hammer_tech.py
```diff
@@ -204,6 +205,36 @@ def store_into_library(self) -> Library:
...
 
+class Site(NamedTuple('Site', [
+    ('name', str),
+    ('x', Decimal),
+    ('y', Decimal)
+])):
+    """
+    A standard cell site, which is the minimum unit of x and y dimensions a standard cell can have.
+
+    name: The name of this site (usually something like "core")
```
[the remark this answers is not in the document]
jwright6323: Sure, will fix.

src/hammer-tech/schema.json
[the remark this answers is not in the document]
jwright6323: If you give `Decimal` a `float` without a precision you get a wonky number:
```
>>> from decimal import Decimal
>>> Decimal(0.001)
Decimal('0.001000000000000000020816681711721685132943093776702880859375')
```
The string is designed to prevent that, as it implies a precision from the string you give it.
```
>>> Decimal("0.001")
Decimal('0.001')
>>> Decimal("0.0010")
Decimal('0.0010')
```
I'm not opposed to manually specifying the precision, but it seems a bit clunky to me. I would propose just documenting why this is a string a bit better. Thoughts?

src/hammer-vlsi/defaults.yml
jwright6323: Oversight, will fix

src/hammer-vlsi/hammer_utils/__init__.py
jwright6323: I'll rework this to use the library `gcd`

src/hammer-vlsi/hammer_utils/__init__.py
jwright6323: OK, I can make it more functional

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
jwright6323: I'll stash this for a future PR.

src/hammer-tech/schema.json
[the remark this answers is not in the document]
edwardcwang: I would say `str` is better from a future-proofing perspective

src/hammer-vlsi/tech_test.py
jwright6323: It's just as clunky, if not more, IMO:
```
for x in ["1.0", "2.0", "3.4", "4.5", "5.25", "50.2"]:
  w = Decimal(x) * m.min_width
# vs.
for x in [1.0, 2.0, 3.4, 4.5, 5.25, 50.2]:
  w = Decimal(coerce_to_grid(x, StackupTestHelper.mfr_grid())) * m.min_width
```

src/hammer-vlsi/utils_test.py
jwright6323: Where would the precision live? Another variable? IMO the `grid_unit` _is_ the precision.

src/hammer-vlsi/utils_test.py
jwright6323: Sure, will add.

src/hammer-vlsi/utils_test.py
colinschmidt: So you want to set the precision to `len(grid_unit)` or something like that?

src/hammer-vlsi/tech_test.py
colinschmidt: ok thats fine then.

src/hammer-vlsi/hammer_utils/__init__.py
jwright6323: done

src/hammer-vlsi/hammer_utils/__init__.py
jwright6323: done

src/hammer-vlsi/hammer_utils/__init__.py
jwright6323: done

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
jwright6323: done

src/hammer-vlsi/utils_test.py
jwright6323: I don't think the precision is doing quite what you're (and I was) expecting. It's for decimal-to-decimal math, not float-to-decimal conversion. For example, it doesn't solve the Decimal(0.001) problem:
```
>>> getcontext().prec = 3
>>> Decimal(0.001)
Decimal('0.001000000000000000020816681711721685132943093776702880859375')
>>> Decimal("0.001") # 1 sig fig
Decimal('0.001')
>>> Decimal("1.00001") # 6 sig figs
Decimal('1.00001')
>>> Decimal("0.001") + Decimal("1.00001") # will result in prec=3 sig figs
Decimal('1.00')
```

src/hammer-tech/schema.json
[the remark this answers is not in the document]
jwright6323: See comment thread below. I think `str` is the way to go.

src/hammer-vlsi/hammer_utils/__init__.py
```diff
@@ -101,6 +99,71 @@ def in_place_unique(items: List[Any]) -> None:
...
+    Return the greatest common divisor of a series of ints
+
+    :param values: The values of which to compute the GCD
+    :return: The GCD
+    """
+    assert len(values) > 0
+    # 3.5 moves fractions.gcd to math.gcd
+    if sys.version_info.major == 3 and sys.version_info.minor < 5:
+        import fractions
+        return reduce(fractions.gcd, values)
```
edwardcwang: https://docs.python.org/3.1/library/fractions.html#fractions.gcd yeah was going to say exactly this

jwright6323:
I think this is ready to go if @colinschmidt is satisfied with the discussion above, but feel free to continue it otherwise.

src/hammer-vlsi/utils_test.py
colinschmidt: Ok well at least we tried.

colinschmidt:
LGTM

Issue #364, opened 2019-04-22
Altium PCB collateral


jwright6323:
This is somewhere in the nice-to-have/beginner-PR zone but it will be a bit of work. I think this is a good undergrad project.

With the new bumps API we should be able to generate library components for designs that are going directly on a PCB. Because we already use Altium, that's a good place to start. The footprint generation (or some components thereof) should live in a technology-specific plugin because it relates to the bump size and metallurgy, but the schematic generation can be fully tech-agnostic. The primary user input should be how to break apart the symbol pins into separate components (will show an EAGLE example of what the result should look like).

https://www.altium.com/documentation/18.0/display/ADES/((Creating+the+PCB+Footprint))_AD
https://techdocs.altium.com/display/SCRT/Script+Examples+Reference#DelphiScripts

edwardcwang:
In the long run it might be good to try to do this in a PCB CAD tool independent way as well...

jwright6323:
Agreed, I just didn't include that in the text of the issue but probably should have. I plan to do that in one combined effort. I think the "PCBTool" base class will end up being very simple at first.

jwright6323:
This is closed by #420

PR #372, opened 2019-04-26
Visualize bumps & macro placement

harrisonliew:
I generated this using a small config of my chip and it really helped me get my bump & macro placement script right:
image

Format tries to mimic what one would see in Innovus.

Also introduces the `pcb` namespace needed for #364


jwright6323:
Cool! A couple high-level requests before we start a detailed review:
- [ ] Can you make the type checker happy
- [ ] Can you merge master? (we just merged a PR so that's not really your fault)
- [ ] Can you show how this is used by the end user? Do you need to call this method in a hook, or is there an API I'm missing?

harrisonliew:
The user simply needs to set `pcb.generic.viz` to `bumps` if they only want to visualize bumps during a `par` action, or `all` to visualize bumps & macro placement. If the bumps API is not used (e.g. not a chip-level block or otherwise defined) it will not draw bumps. No fancy API call.

jwright6323:
IMO there should at least be a way to name the file, or tell it where to go. What is the default "rundir" right now? It'd be nice for it to go parallel with `syn-rundir`, `par-rundir`, etc. I suggest `pcb-rundir` ?

So something like `pcb-rundir/${TOP}.bumps.svg` ?

edwardcwang:
`pcb-rundir` almost suggests the existence of an action `pcb`.

jwright6323:
IMO there should be a `pcb` action which generates pcb deliverables and similar things. Copied from slack:
```pcb.deliverables: ["altium.symbol", "altium.footprint", "kicad.symbol", "kicad.footprint", "generic.bump_svg", "generic.verilog_stub", ...] # we would change / add to this if we ever supported package deliverables or flow
pcb.altium:
  # settings here
pcb.kicad:
  #settings here
pcb.generic:
  #settings here```

**Comment by harrisonliew** [comment]
Actually, I think this should be tied to the action. E.g. if I'm doing floorplanning and I want to visualize macros, bumps, and/or pins, I'd run in the `par` action and it'd appear in `par-rundir`. If only doing pcb deliverables, then a `pcb-rundir` would be appropriate, necessitating a new `pcb` action.

**Comment by jwright6323** [comment]
I'm ambivalent about where exactly the bump/floorplan SVG lives. If it makes more sense to you to have it in `par` that's fine. I do think it should be a separate (first) step, at the very least. My reasoning for that is that you may want to just run the SVG generator to take a look at your floorplan before invoking the CAD tools. I'm not sure how this plays into our current methodology, though, since all other steps generate TCL.

@edwardcwang @colinschmidt thoughts?

**Comment by edwardcwang** [comment]
Yeah given the two divergent usecases pointed out by Harrison, I'm wondering if we should try to expose the action itself in a new Hammer tool class that generically consumes Hammer IR and generates various types of collateral. Then ASIC flows can use the class as part of the flow and PCB actions/flows/whatnot can also take advantage of this feature.

**Comment by jwright6323** [comment]
`collateral.*` +
`HammerCollateralTool` ?

File: src/hammer-vlsi/hammer_vlsi/hammer_tool.py [hunk 1]
```diff
@@ -914,6 +914,129 @@ def get_bumps(self) -> Optional[BumpsDefinition]:
...
 
+    def generate_floorplan_viz(self) -> None:
+        viz = self.get_setting("pcb.generic.viz")
+        if viz in ["all", "bumps"]:
+            fsvg = open(os.path.join(self.run_dir, 'floorplan.svg'), 'w')
+            fsvg.write("""<?xml version="1.0"?>
+            <!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">\n""")
+
+            title_height = 100
+            fp_consts = self.get_placement_constraints()
```
colinschmidt: Don't you need to check that there is something here?
What happens if someone hasn't specified a floorplan but asks for a viz?

colinschmidt:
I think we should try to keep PCB similar to other tools, so rather than `deliverables` it just always produces a footprint and symbol, but based on your `pcb.tool` it may produce it specifically for one vendor or another.
I think we need to keep a fixe style here or we'll lose some of the benefits of the unified environment.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
harrisonliew: Updated to check for empty list and throwing a logging error. However, if no floorplans are specified, other things will break in a par run anyway (like bumps).

colinschmidt: You mean if you specify bumps and no floorplan? Or if you specify no bumps and no floorplan? Can you submit an issue on that

edwardcwang:
Discussion result on 2019-05-10: we should do this as a method that you can mix into any flow
1) mixed into custom flow as a step ← do this for now
2) standalone action
3) default steps in pcb/par
Think about how to do the latter ones after tapeout

jwright6323:
@harrisonliew #420 is in

harrisonliew:
Updating this PR with lots of new functionality:

* 4 modes now: all, floorplan, bumps, footprint (see `defaults.yml` for description)
* `generate_visualization()` is not used by default, but can be called from within any hook
* pcb action will have visualization mirrored about y-axis (top corner dot provides reference)
* Path shortening for better readability (e.g. `path/to/inst` -> `p/t/inst` if the shorten depth is 1)
* Grabs width/height info from LEFs if not specified for hardmacro/hierarchical constraints
* Gray no_connect bumps
* Small Decimal bugfix in PCB tool
* Logic cleanup

Here's an example of what comes out of par action with mode = "all":
image

Here's an example of what comes out of pcb action with mode = "all" (note only toplevel placement constraint is specified here and bump designators are printed above the bump signal names):
image

TODOs (asking for feedback):
* Making pcb action understand the top hierarchical placement constraint
* Translating fractional bump coordinates to integer for pcb footprint (maybe integer locations should be enforced instead?)
* Coloring P/G bumps properly if Hammer doesn't know they're P/G nets (e.g. specified only in CPF)

src/hammer-vlsi/defaults.yml
```diff
@@ -718,7 +741,7 @@ technology.core:
     # Can be a wildcard/glob
     # type: str
     tap_cell_rail_reference: null
-
+
```
colinschmidt: Delete stray space

src/hammer-vlsi/defaults.yml
```diff
@@ -449,6 +449,29 @@ vlsi.inputs:
...
+    mode: "all"
+    # Name of the desired output SVG file
+    # type: str
+    svg_file: "${vlsi.inputs.visualization.mode}.svg"
+    svg_file_meta: lazysubst
+    # Depth of paths to shorten for better readability in the SVG
+    # E.g. top/path/to/inst will shorten to t/p/t/inst for depth of 1 and t/p/top/inst for depth of 2
+    # A depth of 0 or less will perform no path shortening.
+    # type: int
+    shorten_path_depth: 1
```
colinschmidt: Is it really useful to have the `t/p/t` portions of the path?
I guess I'm imagining the hierarchical paths getting very long in submodules and maybe it would be better to have just `top/path/to/submodule/path/to/inst` become `s/p/t/inst`

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -1064,6 +1064,206 @@ def get_bumps(self) -> Optional[BumpsDefinition]:
             pitch=Decimal(str(self.get_setting("vlsi.inputs.bumps.pitch"))),
             cell=self.get_setting("vlsi.inputs.bumps.cell"), assignments=assignments)
 
+    def generate_visualization(self) -> None:
+        """
+        Generate an SVG visualizing the chip.
+        Depending on mode, it will display floorplan (placement constraints for the current hierarchy)
+        and/or bumps (overlaid on top of floorplan) and pad designators (if this tool is a PCBDeliverableTool).
+        Call this from any custom hook (not used by any default flow).
```
colinschmidt: Do we potentially want an action, that makes it easy to just do this visualization?

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -1064,6 +1064,206 @@ def get_bumps(self) -> Optional[BumpsDefinition]:
...
+        # Checks
+        if not viz_mode in ["all", "floorplan", "bumps", "footprint"]:
+            self.logger.error("Invalid type for 'vlsi.inputs.visualization.type'! No visualization generated.")
+            return
+
+        fp_consts = self.get_placement_constraints() #type: List[PlacementConstraint]
+        # TODO: make pcb action recognize toplevel constraint if par action had hierarchical constraints
+        try:
+            # Get toplevel constraint
+            top = next(filter(lambda x: x.type == PlacementConstraintType.TopLevel, fp_consts)) #type: PlacementConstraint
```
colinschmidt: I thought we added a helper method for this. If not we certainly should.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -1064,6 +1064,206 @@ def get_bumps(self) -> Optional[BumpsDefinition]:
...
+        except:
+            self.logger.error("You must at least have a 'toplevel' type placement constraint to generate a visualization!")
+            return
+
+        # TODO: will break if an alternate PCB tool is implemented without naming_scheme property.
```
colinschmidt: Shouldn't the HammerPCBTool just include this as an abstract property to enforce all tools will work?

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -1064,6 +1064,206 @@ def get_bumps(self) -> Optional[BumpsDefinition]:
...
+                return
+            elif viz_mode == "all":
+                self.logger.info("Visualizing in 'all' mode but not a PCBDeliverableTool. Skipping drawing pad designators.")
+
+        # Shorten path names to first letters except lowest "depth" levels in hierarchy
+        def shorten(path: str) -> str:
+            if shorten_path_depth > 0:
+                parent = path.split('/')[:-shorten_path_depth]
+                inst = path.split('/')[-shorten_path_depth:]
+                return os.path.join(*[p[0] for p in parent], *inst)
```
colinschmidt: I was expecting this to be a regular python join not path join so was confused why it didn't include `/`.
For reference I think the equivalent "normal" join code is:
```
"/".join([p[0] for p in parent] + inst)
```
I don't really care one way or the other it just seems like a strange use of the `os.path` stuff

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -1064,6 +1064,206 @@ def get_bumps(self) -> Optional[BumpsDefinition]:
...
+            else:
+                return path
+
+        # Get all bump & macro sizes from their masters when constraint width & height are not defined
+        def get_macro_wh(macro: Optional[str]) -> Tuple[Decimal, Decimal]:
+            # take 1st definition of macro
```
colinschmidt: Do we have to make this kind of decision anywhere else?
If so is it also first? I naively would have assumed it would be last, such that I could keep appending libraries and the most recent append would be picked up (like what happens when you have multiple hammer IR snippets appended).

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -1064,6 +1064,206 @@ def get_bumps(self) -> Optional[BumpsDefinition]:
...
+            try:
+                size = next(filter(lambda x: x.name == macro, macro_sizes))
+                return (size.width, size.height)
+            except:
+                self.logger.warning("Size of {} not found in any LEFs! Defaulting to 10um x 10um. Try manually specifying width & height.".format(macro))
+                return (Decimal("10"), Decimal("10"))
```
colinschmidt: Is this a common issue that we want to support?
I would have assumed you just want to error.

src/hammer-vlsi/defaults.yml
```diff
@@ -449,6 +449,29 @@ vlsi.inputs:
     #   Requires the semi_auto pin generation mode.
     assignments: []
 
+  # SVG visualization of placement constraints and bumps for the current hierarchical module.
```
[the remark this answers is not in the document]
harrisonliew: No. This would require a fundamental rewrite of hierarchical Hammer. When you call `get_placement_constraints`, Hammer goes through the graph and picks out only the constraints that apply to the current node.

src/hammer-vlsi/defaults.yml
harrisonliew: It already is `s/p/t/inst` for submodules because the first letter of the path will always be the current hierarchical module. So using your example:

Flat flow: `t/p/t/s/p/t/inst`
Hier flow, viz in `submodule`: `s/p/t/inst`
Hier flow, viz in top module: `t/p/t/submodule_inst`

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
harrisonliew: I think this was a philosophical question that we addressed earlier in this PR; "1) mixed into custom flow as a step" was what was decided to do for now.

We can address whether we want a standalone action and/or add it as default steps in par/pcb actions later.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
harrisonliew: There is (`get_toplevel` in the hierarchical helper of `HammerDriver`), but I'm pretty sure it's out of scope from the perspective of `HammerTool` because that is used before hierarchical constraints are processed.

src/hammer-vlsi/defaults.yml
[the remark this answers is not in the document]
colinschmidt: I think it would be cool to do both but we can just leave it as an issue.

src/hammer-vlsi/defaults.yml
colinschmidt: Ok I guess we should just think about this again when/if we do the nested viz

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
colinschmidt: Ok we'll discuss offline.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
colinschmidt: Hmm, seems like a flaw in that helper method, yet another issue 😢

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
harrisonliew: Yeah, the way I did it was definitely a case of trying to reduce some lines of code and not realizing I could do a simple string join...

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
harrisonliew: I'm quite sure that the `macro_sizes` here (`= self.technology.get_macro_sizes()`) has already done all the meta-processing for appending libraries because the extraction func will only accept 1 path (i.e. LEF files will be unique). But it's possible that multiple unique LEFs may define the same macro and provide multiple sizes? I'm not sure how we can decide which one is the best one there besides the first one... thoughts?

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
harrisonliew: No, I don't think so... not yet at least because we don't check for sizes anywhere else yet (will need to when Chisel floorplanning stuff matures). I don't want to error because it's more obvious that you probably left out an `extra_libraries` entry or something when seeing the viz with the 10x10 block rather than just seeing the text of the error, which gets lost in the flurry of other logger warnings during the `get_macro_sizes()` call.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
harrisonliew: Yeah, moved property to `HammerPCBDeliverableTool` from `GenericHammerPCTDeliverableTool` since it's required anyway (only property set by depends on a `vlsi.inputs` key).

It by default outputs a `BumpsPinNamingScheme`. Maybe this is limiting in the future if we support chips with wirebonding?

PR #373, opened 2019-04-26
Add travis support

colinschmidt:
This passes and replicates most of the build-bot (it should be able to support 3.3 and 3.4 but I don't know how the build-bot works around the typing issue).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+dist: xenial   # required for Python >= 3.7
+language: python
+python:
+  - "3.5"
+  - "3.6"
+  - "3.7"
+
+script:
+  - export HAMMER_HOME=$PWD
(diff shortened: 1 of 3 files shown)


edwardcwang:
Following up from offline discussion:

- buildbot no longer spews comments but instead leaves a "Details" link like Travis/etc
- Edward to clean up and release buildbot config so all core devs can push/modify it
- Edward to put some setup files into hammer itself to make test setups easier to reproduce

jwright6323:
Buildbot is down, so I'm going to go ahead and merge this because none of us here know how to get it back up. I will be happy to revert this and use buildbot when it's back up and accessible to other devs.

jwright6323:
Also, I tested the various failure cases in #374, then reverted it to the commit it was on.

PR #382, opened 2019-05-08
Fix typo and tweak type descriptions

edwardcwang:
No functional change. `std_cell_rail_layer` is actually the correct key


edwardcwang:
The eventual idea is for them to map to Python types and that's what I was trying to be consistent with.

PR #391, opened 2019-05-10
Reference vlsi.inputs.supplies in power_spec_mode

### LINKED ISSUE #387: Update power_spec_mode description now that auto mode is implemented
Description should mention that auto uses `vlsi.inputs.supplies`

colinschmidt:
This fixes #387

### DIFF
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -276,6 +276,7 @@ vlsi.inputs:
   # how many domains you have and what their relation is.
   # Valid options:
   # auto - Hammer will generate a power specification based on other settings.
+  #   Currently the other settings include at least vlsi.inputs.supplies.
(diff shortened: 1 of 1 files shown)


src/hammer-vlsi/defaults.yml
```diff
@@ -276,6 +276,7 @@ vlsi.inputs:
   # how many domains you have and what their relation is.
   # Valid options:
   # auto - Hammer will generate a power specification based on other settings.
+  #   Currently the other settings include at least vlsi.inputs.supplies.
```
edwardcwang: What do you think about something like "Hammer will generate a power specification based on other settings, including vlsi.inputs.supplies"? I had to think a bit about "Currently the other settings include" on first glance.

colinschmidt: Sure go for it.

PR #398, opened 2019-05-17
Implement semi-auto pin placement API

edwardcwang:
**Metadata: rebase-merge**

Companion PR forthcoming.

### DIFF
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -337,19 +337,48 @@ vlsi.inputs:
   # auto - (not implemented yet) looks at the hierarchy and guess where pins should go
   pin_mode: none
   pin:
+    # Pin generation modes.
+    # type: str
+    # Valid options are:
+    # - full_auto: auto-place minimum-width pins (intended for digital signals) along a side. This option is the default.
+    # - semi_auto: enables the auto-place features above but also optionally allows pin width, depth, and locations to
+    #   be manually specified.
+    generate_mode: full_auto
+
     # List of PinAssignment Structs.
     #   You must specify pins and one of either preplaced or layers and side.
     # - pins (str) - The name(s) of pins that will be assigned. All current backends support
     #   * as wildcards in the pin names. So "*" for all pins or "io_tl_a*" for all pins that
     #   start with "io_tl_a".
     # - side (Optional[str]) - The side of the chip/block these pins will be placed along.
-    #   One of "left", "right", "top", "bottom"
+    #   One of:
+    #   - left, right, top, bottom
+    #   - internal: creates an internal pin with `location` as its *center*.
+    #     Requires the semi_auto pin generation mode.
+    #   If left unspecified, the CAD tool may choose an arbitrary side to use (likely the closest side).
     # - layers (Optional[List[str]]) - The metal layers that these pins will be placed on.
     # - preplaced (Optional[bool]) - If true, these pins are preplaced by some internal block or macro.
     #   This is telling the tool that these pins are on an internal block and I want them
     #   "placed" where they are currently on the internal block. This can be used for IP
     #   that directly connects to bumps, or signals that will be connected by abutment,
     #   and much more.
+    # - location (Optional[Tuple[float, float]]) - specifies the *center* (x, y) location of the pin on the specified
+    #   edge.
+    #   If side is one of left/right/top/bottom, this specifies the point that lies on the edge that is used for
+    #   width/depth control. Below is an example of location if side is "top":
+    #
+    #   -------X------- (chip/block edge)
+    #   |             |
+    #   ---------------
+    #
+    #   If side is "internal", then this specifies the absolute *center* of the
+    #   Requires the semi_auto pin generation mode.
+    # - width (Optional[float]) - specifies a custom width for the pin. The width direction is orthogonal to the routing
+    #   direction.
+    #   Requires the semi_auto pin generation mode.
+    # - depth (Optional[float]) - specifies a custom depth for the pin. The depth direction is parallel to the routing
+    #   direction.
+    #   Requires the semi_auto pin generation mode.
     assignments: []
 
 vlsi.submit:
--- a/src/hammer-vlsi/hammer_vlsi/constraints.py
+++ b/src/hammer-vlsi/hammer_vlsi/constraints.py
@@ -20,7 +20,7 @@ __all__ = ['ILMStruct', 'SRAMParameters', 'Supply', 'PinAssignment',
            'BumpAssignment', 'BumpsDefinition', 'ClockPort',
            'OutputLoadConstraint', 'DelayConstraint', 'ObstructionType',
            'PlacementConstraintType', 'Margins', 'PlacementConstraint',
-           'MMMCCornerType', 'MMMCCorner']
+           'MMMCCornerType', 'MMMCCorner', 'PinAssignmentError', 'PinAssignmentPreplacedError']
 
 
 # Struct that holds various paths related to ILMs.
@@ -82,6 +82,7 @@ class SRAMParameters(NamedTuple('SRAMParameters', [
             mux=int(d["mux"])
         )
 
+
 Supply = NamedTuple('Supply', [
     ('name', str),
     ('pin', Optional[str]),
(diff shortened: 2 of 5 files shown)


src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -931,35 +931,43 @@ def get_pin_assignments(self) -> List[PinAssignment]:
                 "Invalid pin_mode {mode}. Using none pin mode.".format(mode=pin_mode))
             return []
 
+        generate_mode = str(self.get_setting("vlsi.inputs.pin.generate_mode"))
+        if generate_mode not in ("full_auto", "semi_auto"):
```
jwright6323: Hmm this is OK for now, but we may want to set a new standard for `generate_mode`s where the code that implements them defines the list of valid options. Maybe when/if we split `hammer_vlsi_impl.py` into multiple files (e.g. I could see a `pin_assignments.py` that contains a singleton `PinAssignmentModes` list).

edwardcwang: Do we actually envision more than these two? I feel that these two modes pretty much cover the major usecases we can think of for now.

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -80,19 +82,152 @@ def from_setting(d: dict) -> "SRAMParameters":
...
+            p=self.pin.pins)
+
+
+class PinAssignment(NamedTuple('PinAssignment', [
     ('pins', str),
     ('side', Optional[str]),
     ('layers', Optional[List[str]]),
-    ('preplaced', Optional[bool])
-])
+    ('preplaced', Optional[bool]),
```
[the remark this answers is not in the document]
edwardcwang: My understanding is that if `preplaced` no other options are valid (see the defaults.yml documentation)

Long term we may want to have subclasses of PinAssignment (e.g. `PreplacedPinAssignment`, etc)

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -80,19 +82,152 @@ def from_setting(d: dict) -> "SRAMParameters":
...
+                depth: Optional[float] = None) -> "PinAssignment":
+        return super().__new__(cls, pins, side, layers, preplaced, location, width, depth)
+
+    @staticmethod
+    def create(pins: str, side: Optional[str] = None, layers: Optional[List[str]] = None,
+                preplaced: Optional[bool] = None,
+                location: Optional[Tuple[float, float]] = None, width: Optional[float] = None,
+                depth: Optional[float] = None) -> "PinAssignment":
+        """
+        Static method that works around the fact that mypy gets very confused at the above.
```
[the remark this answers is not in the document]
edwardcwang: Fixed

src/hammer-vlsi/defaults.yml
```diff
@@ -341,15 +341,44 @@ vlsi.inputs:
...
+    #   Requires the semi_auto pin generation mode.
+    # - width (Optional[float]) - specifies a custom width for the pin. The width direction is orthogonal to the routing
+    #   direction.
+    #   Requires the semi_auto pin generation mode.
+    # - depth (Optional[float]) - specifies a custom depth for the pin. The depth direction is parallel to the routing
+    #   direction.
+    #   Requires the semi_auto pin generation mode.
     assignments: []
 
+    # Pin generation modes.
```
colinschmidt: Can we move this above the assignments struct so the references to it aren't backwards?

src/hammer-vlsi/defaults.yml
```diff
@@ -341,15 +341,44 @@ vlsi.inputs:
...
+    # type: str
+    # Valid options are:
+    # - full_auto: auto-place minimum-width pins (intended for digital signals) along a side. This option is the default.
+    # - semi_auto: enables the auto-place features above but also optionally allows pin width, depth, and positions to
```
colinschmidt: change positions -> locations so it matches the key in assignments struct.

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -80,19 +82,153 @@ def from_setting(d: dict) -> "SRAMParameters":
...
+    """Exception raised from parsing an invalid PinAssignment object."""
+
+
+class PinAssignmentSemiAutoError(PinAssignmentError):
+    """Exception raised from parsing a PinAssignment that requires semi-auto features without explicitly enabing
+    them."""
+
+
+class PinAssignmentPreplacedError(PinAssignmentError):
+    """Exception raised from parsing a replaced pin with extraneous information."""
```
colinschmidt: replace->preplaced

src/hammer-vlsi/hammer_vlsi/constraints.py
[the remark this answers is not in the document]
colinschmidt: Preplaced means you only need the `pins` part of the struct because the pins are already defined by some internal structure so all other parameters are ignored.

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -80,19 +82,153 @@ def from_setting(d: dict) -> "SRAMParameters":
...
+        else:
+            if len(get_or_else(layers, cast(List[str], []))) == 0 or side is None:
+                raise PinAssignmentError(
+                    "Pins {p} assigned without layers or side. Assuming pins will be handled by CAD tool.".format(
+                        p=pins))
+        return PinAssignment(
+            pins=pins,
+            side=side,
+            layers=layers,
+            preplaced=preplaced if "preplaced" in raw_assign else None,
```
colinschmidt: We want the default to be `Some(False)`.

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -80,19 +82,153 @@ def from_setting(d: dict) -> "SRAMParameters":
...
+
+    def to_dict(self) -> dict:
+        base = {
+            "pins": self.pins
+        }  # type: Dict[str, Any]
+        if self.side is not None:
+            base['side'] = self.side
+        if self.layers is not None:
+            base['layers'] = self.layers
+        if self.preplaced is not None:
```
colinschmidt: As above default should be `Some(False)`

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -931,35 +931,43 @@ def get_pin_assignments(self) -> List[PinAssignment]:
...
+                # Accept the pin assignment and ignore extra information
+                self.logger.warning(str(e))
+                assigns.append(e.pin)
+                continue
+            except PinAssignmentError as e:
+                # Ignore the invalid pin
+                self.logger.warning(str(e))
+                continue
+
+            if pin.layers is not None:
```
colinschmidt: Why is this one exception check left out of the actual exception match?

src/hammer-vlsi/hammer_vlsi/constraints.py
edwardcwang: Hmm, maybe it might be necessary to store an underlying preplaced value that's not the default. For example, if we just put False here, it breaks a round trip of JSON->object->JSON comparison as it injects the default instead of preserving the None.

src/hammer-vlsi/hammer_vlsi/constraints.py
edwardcwang: Ditto above

src/hammer-vlsi/hammer_vlsi/constraints.py
colinschmidt: I think thats kind of the point. We don't really want this to be optional(we're just letting it be so for the lazy users see: https://en.wikipedia.org/wiki/Robustness_principle), so when hammer emits it hammer will always use the default.

edwardcwang: Maybe in that case we should just set the internal type of `preplaced` to bool then. I can go ahead and do that

colinschmidt: I mean thats not what I want. I like not having to say preplaced in my input yamls. I just think we should be strict in what we output. Because the output is computer generated while the input is often human generated.

edwardcwang: I think we might be talking about different things here; I'd say to allow the dict parser to have an optional `preplaced` but have the internal Python type be just `bool` so that import/export is consistent.

colinschmidt: Ah ok. Yes that makes sense.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
jwright6323: Yeah I can't really think of anything else.

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -80,19 +82,153 @@ def from_setting(d: dict) -> "SRAMParameters":
...
+        return super().__new__(cls, pins, side, layers, preplaced, location, width, depth)
+
+    @staticmethod
+    def create(pins: str, side: Optional[str] = None, layers: Optional[List[str]] = None,
+                preplaced: Optional[bool] = None,
+                location: Optional[Tuple[float, float]] = None, width: Optional[float] = None,
+                depth: Optional[float] = None) -> "PinAssignment":
+        """
+        Static method that works around the fact that mypy gets very confused at
+        the custom constructor above that defines default arguments.
```
jwright6323: I'm still not sure I understand what's going on here- The comment is probably OK, but could you explain to me why mypy doesn't like the new constructor? Can you not make it happy with a type hint in `__new__` for some reason?

edwardcwang: I don't know the underlying technical reason, and I suspect it to be something internals to the workings of mypy. I think the upstream/desired fix is to move to Python3.6+ `NamedTuple`s or Python 3.7+ dataclasses, which we can't do (yet).

edwardcwang:
All changes addressed.

jwright6323:
Fine with me, should probably also get @colinschmidt to sign off

edwardcwang:
@colinschmidt Any further thoughts? Would like to merge EOD today or tomorrow.

PR #399, opened 2019-05-17
Add some assertions to sanity check the tech plugin

jwright6323:
We ran into issues with a malformed tech plugin that silently generated nonsensical power straps. Also funnily enough, our place and route tool of choice does NOT error out when you specify a negative power strap width...

### DIFF
--- a/src/hammer-tech/stackup.py
+++ b/src/hammer-tech/stackup.py
@@ -279,6 +279,7 @@ class Metal(NamedTuple('Metal', [
         """
         widths_and_spacings = self.power_strap_widths_and_spacings
         spacing = widths_and_spacings[0].min_spacing
+        assert self.pitch - self.min_width == spacing, "Tech plugin is malformed for metal {}, the minimum spacing in the width-spacing list must be the same as (pitch - min_width).".format(self.name)
         # the T W W T pattern contains two wires (W2) and 3 spaces (S3)
         s3w2 = ((2 * tracks) + 1) * self.pitch - self.min_width
         width = (s3w2 - spacing * 3) / 2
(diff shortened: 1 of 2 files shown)


src/hammer-tech/stackup.py
```diff
@@ -279,6 +279,7 @@ def get_width_spacing_start_twwt(self, tracks: int, force_even: bool = False) ->
         """
         widths_and_spacings = self.power_strap_widths_and_spacings
         spacing = widths_and_spacings[0].min_spacing
+        assert self.pitch - self.min_width == spacing, "Tech plugin is malformed for metal {}, the minimum spacing in the width-spacing list must be the same as (pitch - min_width).".format(self.name)
```
edwardcwang: Minor point, can we add unit tests for these kinds of bad cases?

jwright6323: IDK what you mean- there are no unit tests in the tech plugin.

edwardcwang: Well right now they live in `hammer-vlsi/tech_test.py` IIRC

jwright6323: but this isn't a problem with the tech code- it's catching a problem with a bad plugin. Are you saying you want a test that contains a bad plugin?

edwardcwang: Yeah, a test that purposely feeds it a bad plugin and check that it's caught

jwright6323: I think that's beyond the scope of this PR, but we can file an issue.

jwright6323: Filed #402

Issue #376, opened 2019-04-29
Convert all length units to Decimal


jwright6323:
Right now we use `float`s for most physical length units. There were real issues caused by this in the power straps code, so I wrote a function to snap things to a grid. However it's pretty easy to forget to use this, so I think going forward we should use the `Decimal` type, which will be much safer.

jwright6323:
- [ ] In https://github.com/ucb-bar/hammer-cad-plugins/pull/107/files @harrisonliew is using `Decimal(str(<float>))` which seems OK to me, but we might want to add some grid checks in there along with fixing this issue.

jwright6323:
Closed by #413

PR #403, opened 2019-05-19
Enhance bump API

harrisonliew:
Add `custom_cell` field in BumpAssignment constraint to override the default bump for any given bump.

For https://github.com/ucb-bar/hammer-cad-plugins/pull/107

### DIFF
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -324,7 +324,7 @@ vlsi.inputs:
     #  - no_connect (Optional[bool]) - If True the bump will be present but unconnected
     #  - x (Decimal) - The X coordinate of the bump assigned to this net
     #  - y (Decimal) - The Y coordinate of the bump assigned to this net
-    #  - custom_cell (Optional[str[) - Cell name (MACRO name) of a custom bump cell in place of the default for this bump. You must also specify the LEF and GDS for this bump in extra_libraries.
+    #  - custom_cell (Optional[str]) - Cell name (MACRO name) of a custom bump cell in place of the default for this bump. You must also specify the LEF and GDS for this bump in extra_libraries.
     assignments: []
 
   # Pin placement mode. (str)
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -322,8 +322,8 @@ vlsi.inputs:
     # If both are specified the bump will be left unconnected
     #  - name (Optional[str]) - The name of the net being assigned to a bump
     #  - no_connect (Optional[bool]) - If True the bump will be present but unconnected
-    #  - x (int) - The X coordinate of the bump assigned to this net
-    #  - y (int) - The Y coordinate of the bump assigned to this net
+    #  - x (float) - The X coordinate of the bump assigned to this net
+    #  - y (float) - The Y coordinate of the bump assigned to this net
     #  - custom_cell (Optional[str[) - Cell name (MACRO name) of a custom bump cell in place of the default for this bump. You must also specify the LEF and GDS for this bump in extra_libraries.
     assignments: []
 
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -316,7 +316,7 @@ vlsi.inputs:
     y: 0
     # pitch (float) - pitch of bumps in microns
     pitch: 0.0
-    # cell (str) - Name of bump cell
+    # cell (str) - Name of the default bump cell
     cell: ""
     # assignments - List of BumpAssignment structs. You must specify one of name or no_connect.
     # If both are specified the bump will be left unconnected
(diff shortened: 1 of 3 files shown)


src/hammer-vlsi/defaults.yml
```diff
@@ -316,14 +316,15 @@ vlsi.inputs:
...
     # pitch (float) - pitch of bumps in microns
     pitch: 0.0
-    # cell (str) - Name of bump cell
+    # cell (str) - Name of the default bump cell
     cell: ""
     # assignments - List of BumpAssignment structs. You must specify one of name or no_connect.
     # If both are specified the bump will be left unconnected
     #  - name (Optional[str]) - The name of the net being assigned to a bump
     #  - no_connect (Optional[bool]) - If True the bump will be present but unconnected
     #  - x (int) - The X coordinate of the bump assigned to this net
```
colinschmidt: Based on the other PR shouldn't this mention that these can be fractions now?

src/hammer-vlsi/defaults.yml
```diff
@@ -316,14 +316,15 @@ vlsi.inputs:
...
     #  - y (int) - The Y coordinate of the bump assigned to this net
+    #  - custom_cell (Optional[str[) - MACRO name of a custom bump cell in place of the default for this bump. You must also specify the LEF and GDS for this bump in extra_libraries.
```
```suggestion
    #  - custom_cell (Optional[str]) - Cell name (MACRO name) of a custom bump cell in place of the default for this bump. You must also specify the LEF and GDS for this bump in extra_libraries.
```

src/hammer-vlsi/defaults.yml
edwardcwang: Are x and y co-ordinates still intended to be integers?

src/hammer-vlsi/defaults.yml
colinschmidt: I think the second sentence is redundant but its worth being verbose :)

src/hammer-vlsi/defaults.yml
harrisonliew: no, updated to float for the BumpAssignment constraint also.

harrisonliew: see above

src/hammer-vlsi/defaults.yml
harrisonliew: well, I think it would be redundant if there was a way to cross-check this  with the provided LEFs before sending it off to the CAD tools.

src/hammer-vlsi/defaults.yml
jwright6323: unrelated question to @colinschmidt, does our tech plugin override this default, or do we specify this at the chip level?

src/hammer-vlsi/defaults.yml
```diff
@@ -316,14 +316,15 @@ vlsi.inputs:
...
+    # cell (str) - Name of the default bump cell
     cell: ""
     # assignments - List of BumpAssignment structs. You must specify one of name or no_connect.
     # If both are specified the bump will be left unconnected
     #  - name (Optional[str]) - The name of the net being assigned to a bump
     #  - no_connect (Optional[bool]) - If True the bump will be present but unconnected
-    #  - x (int) - The X coordinate of the bump assigned to this net
-    #  - y (int) - The Y coordinate of the bump assigned to this net
+    #  - x (float) - The X coordinate of the bump assigned to this net
+    #  - y (float) - The Y coordinate of the bump assigned to this net
```
jwright6323: `floats` are dangerous here. `Decimal` would be better.

src/hammer-vlsi/defaults.yml
colinschmidt: This is a chip level thing. There are many reasonable cells in our tech library we are just constrained to a certain one for this project.

src/hammer-vlsi/defaults.yml
jwright6323: (see #376)

src/hammer-vlsi/defaults.yml
jwright6323: You could still override it- I think it makes sense to have a sane default in the tech plugin, but at this point it would only help future projects.

src/hammer-vlsi/defaults.yml
harrisonliew: type hint changed to Decimal.

src/hammer-vlsi/defaults.yml
```diff
@@ -316,14 +316,15 @@ vlsi.inputs:
...
     cell: ""
     # assignments - List of BumpAssignment structs. You must specify one of name or no_connect.
     # If both are specified the bump will be left unconnected
     #  - name (Optional[str]) - The name of the net being assigned to a bump
     #  - no_connect (Optional[bool]) - If True the bump will be present but unconnected
-    #  - x (int) - The X coordinate of the bump assigned to this net
-    #  - y (int) - The Y coordinate of the bump assigned to this net
+    #  - x (Decimal) - The X coordinate of the bump assigned to this net
+    #  - y (Decimal) - The Y coordinate of the bump assigned to this net
+    #  - custom_cell (Optional[str[) - Cell name (MACRO name) of a custom bump cell in place of the default for this bump. You must also specify the LEF and GDS for this bump in extra_libraries.
```
```suggestion
    #  - custom_cell (Optional[str]) - Cell name (MACRO name) of a custom bump cell in place of the default for this bump. You must also specify the LEF and GDS for this bump in extra_libraries.
```

PR #404, opened 2019-05-20
Add input ILMs to the LVS interface to allow hierarchical LVS

harrisonliew:
Needed for https://github.com/ucb-bar/hammer-cad-plugins/pull/109

### DIFF
--- a/src/hammer-vlsi/hammer_vlsi/driver.py
+++ b/src/hammer-vlsi/hammer_vlsi/driver.py
@@ -385,12 +385,12 @@ class HammerDriver:
         lvs_tool.set_database(self.database)
         lvs_tool.submit_command = HammerSubmitCommand.get("lvs", self.database)
         lvs_tool.run_dir = run_dir
-        # TODO: support hierarchical
 
         lvs_tool.schematic_files = self.database.get_setting("lvs.inputs.schematic_files", nullvalue=[])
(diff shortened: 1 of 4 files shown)


src/hammer-vlsi/hammer_vlsi/driver.py
```diff
@@ -385,12 +385,12 @@ def load_lvs_tool(self, run_dir: str = "") -> bool:
         lvs_tool.set_database(self.database)
         lvs_tool.submit_command = HammerSubmitCommand.get("lvs", self.database)
         lvs_tool.run_dir = run_dir
-        # TODO: support hierarchical
 
         lvs_tool.schematic_files = self.database.get_setting("lvs.inputs.schematic_files", nullvalue=[])
         lvs_tool.layout_file = self.database.get_setting("lvs.inputs.layout_file", nullvalue="")
         lvs_tool.top_module = self.database.get_setting("lvs.inputs.top_module", nullvalue="")
         lvs_tool.hcells_list = self.database.get_setting("lvs.inputs.hcells_list", nullvalue=[])
+        lvs_tool.ilms = self.database.get_setting("lvs.inputs.ilms", nullvalue=[])
```
edwardcwang: 1) While Python may let you dynamically create new properties, we should avoid doing it this way  (see static type checker failure). We should create it as a proper "property" in `HammerLVSTool`.

2) When you do so, can you define the type as `List[ILMStruct]` and pre-process the dictionaries into structs?

Issue #400, opened 2019-05-17
Mode enum


edwardcwang:
We use `mode` often enough that we may as well create a proper enum for it rather than comparing against a string.

Possibly related to #347, #357, #358

Issue #397, opened 2019-05-16
Change power straps to be par.power_straps.generate_method


edwardcwang:
_Originally posted by @edwardcwang in https://github.com/ucb-bar/hammer/issues/371#issuecomment-491371189_

edwardcwang:
CC @jwright6323

jwright6323:
Also change the mode from `generated` to `generate`

PR #407, opened 2019-05-22
Mode enum

Jingyi99:
Issue: https://github.com/ucb-bar/hammer/issues/400
I was not entirely sure where to put the mode enum class so I put it in hammer_tool.py. I can move it if there is a better place to put it.

### DIFF
--- a/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
+++ b/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
@@ -28,6 +28,9 @@ from .hooks import (HammerStepFunction, HammerToolHookAction, HammerToolStep,
 from .submit_command import HammerSubmitCommand
 from .units import TemperatureValue, TimeValue, VoltageValue
 
+from enum import Enum
+from hammer_utils import reverse_dict
+
 __all__ = ['HammerTool']
 
 
@@ -1128,3 +1131,34 @@ class HammerTool(metaclass=ABCMeta):
         """
         output_buffer.append("""puts "{0}" """.format(cmd.replace('"', '\"')))
         output_buffer.append(cmd)
+class ModeType(Enum):
+    Auto = 1
+    Empty = 2
+    Manual = 3
+    Generate = 4
+    Generated = 5
+    Append = 6
+    Prepend = 7
+
+    @classmethod
+    def __mapping(cls) -> Dict[str, "ModeType"]:
+        return {
+            "auto": ModeType.Auto,
+            "empty": ModeType.Empty,
+            "manual": ModeType.Manual,
+            "generate": ModeType.Generated,
+            "generated": ModeType.Generate,
+            "append": ModeType.Append,
+            "prepend": ModeType.Prepend
+        }
+
+    @staticmethod
+    def from_str(input_str: str) -> "ModeType":
+        try:
+            return ModeType.__mapping()[input_str]
+        except KeyError:
+            raise ValueError("Invalid mode type: " + str(input_str))
+
+    def __str__(self) -> str:
+        return reverse_dict(ModeType.__mapping())[self]
+
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -437,10 +437,10 @@ par:
 
   # Power straps configuration.
   # Valid options are:
-  # - blank - Specify no power straps
+  # - empty - Specify no power straps
   # - manual - Specify the contents of a manual TCL script to use, specified in power_straps_script_contents.
   # - generate - Generate power straps from Hammer IR.
-  power_straps_mode: blank
+  power_straps_mode: empty
   power_straps_script_contents: null
 
   # Spacing around blocks and hardmacros in microns


src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -1128,3 +1131,34 @@ def verbose_tcl_append(cmd: str, output_buffer: List[str]) -> None:
         """
         output_buffer.append("""puts "{0}" """.format(cmd.replace('"', '\"')))
         output_buffer.append(cmd)
+class ModeType(Enum):
+    Auto = 1
+    Empty = 2
+    Manual = 3
+    Generate = 4
+    Generated = 5
```
edwardcwang: As far as I'm aware, the modes we use are:
```
    Auto = 1
    Empty = 2
    Manual = 3
    Generated = 4
```

`append`, `prepend` are part of another part of Hammer and not part of this mode enum.

Also, if we are going to create this enum, I think it may be good to also use them in some of the settings.

jwright6323: We should change power straps to `generated` from `generate` if we're going to use `generate`. I personally like `generate` better, but `generated` does keep all of the modes as adjectives, which also makes sense.

edwardcwang: Yeah, maybe we can do that as a separate issue then?

jwright6323: We can just tack it on to #397 .

Jingyi99: Hi Edward, thanks for your comment! I checked defaults.yml and it seems that pin place mode uses `generated`, and power straps mode uses `generate`. Just to make sure, they are the same and we'll stick with `generate`, right?
Also, I'm not sure what it means to use the enum in the settings, could you explain a bit more?

jwright6323: Hi Jingyi- let's use `generated` instead of `generate`. We'll change the power straps mode to `generate`.

edwardcwang: Yeah, so John's comment above is alluding to that problem. I think the desired action is to use `generated` and just ignore the power straps mode for now.

As for use, if you look in the code, right now it uses strings - the intention is to use this enum instead

https://github.com/ucb-bar/hammer/blob/8bcbc40d8428ad62a5f66df8817b02ce629aa048/src/hammer-vlsi/hammer_vlsi/hammer_tool.py#L896

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -970,7 +973,7 @@ def get_gds_map_file(self) -> Optional[str]:
         :return: Fully-resolved path to GDS map file or None.
         """
         # Mode can be auto, empty, or manual
-        gds_map_mode = str(self.get_setting("par.inputs.gds_map_mode"))  # type: str
+        gds_map_mode = ModeType.from_str(str(self.get_setting("par.inputs.gds_map_mode")))  # type: ModeType Enum
```
edwardcwang: ditto

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -892,10 +895,10 @@ def get_independent_ground_nets(self) -> List[Supply]:
         return list(filter(lambda x: x.tie is None, self.get_all_ground_nets()))
 
     def get_bumps(self) -> Optional[BumpsDefinition]:
-        bumps_mode = self.get_setting("vlsi.inputs.bumps_mode")
-        if bumps_mode == "empty":
+        bumps_mode = ModeType.from_str(self.get_setting("vlsi.inputs.bumps_mode")) #Type: ModeType Enum
```
```suggestion
        bumps_mode = ModeType.from_str(self.get_setting("vlsi.inputs.bumps_mode"))  # type: ModeType
```

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -1128,3 +1131,27 @@ def verbose_tcl_append(cmd: str, output_buffer: List[str]) -> None:
         """
         output_buffer.append("""puts "{0}" """.format(cmd.replace('"', '\"')))
         output_buffer.append(cmd)
+class ModeType(Enum):
```
```suggestion

class ModeType(Enum):
```

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -1323,15 +1323,15 @@ def generate_power_spec_commands(self) -> List[str]:
             return []
 
         power_spec_contents = ""  # type: str
-        power_spec_mode = str(self.get_setting("vlsi.inputs.power_spec_mode"))  # type: str
-        if power_spec_mode == "empty":
+        power_spec_mode = ModeType.from_str(str(self.get_setting("vlsi.inputs.power_spec_mode")))  # type: ModeType Enum
```
edwardcwang: ditto

Jingyi99:
It seems like I broke something...

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -970,7 +973,7 @@ def get_gds_map_file(self) -> Optional[str]:
         :return: Fully-resolved path to GDS map file or None.
         """
         # Mode can be auto, empty, or manual
-        gds_map_mode = str(self.get_setting("par.inputs.gds_map_mode"))  # type: str
+        gds_map_mode = ModeType.from_str(str(self.get_setting("par.inputs.gds_map_mode")))  # type: ModeType
```
jwright6323: We may want to create a static helper method called `ModeType.from_setting` that takes the `HammerTool` object and the setting string as args and returns a `ModeType`. e.g.

```
class ModeType:
  @staticmethod
  def from_setting(tool: HammerTool, setting: str) -> ModeType:
    return ModeType.from_str(str(tool.get_setting(setting)))
```
which would then allow the following change (no need for type hint, either):

```suggestion
        gds_map_mode = ModeType.from_setting(self, "par.inputs.gds_map_mode")
```

edwardcwang: Not to nitpick too much - at that point I would prefer almost `gds_map_mode = self.get_mode_setting("par.inputs.gds_map_mode")` to avoid a dependency from `ModeType` -> `HammerTool`

jwright6323: I also like that, perhaps more.

Jingyi99:
Sorry for introducing bugs. The errors in the unit tests are 
  ```
File "/buildbot/run_unittests/build/src/hammer-vlsi/hammer_vlsi/hammer_tool.py", line 1135, in HammerTool
    def get_mode_setting(self, setting: str) -> ModeType:
NameError: name 'ModeType' is not defined
```
Is it potentially because I was not importing properly?
Also, are there any units tests that I can run locally?
Thanks!

jwright6323:
Hi Jingyi- you can run the unittests in src/test/unittests.sh: https://github.com/ucb-bar/hammer/blob/master/src/test/unittests.sh
You cal also run the typechecker by running the mypy_with_exclusions.sh script

PR #410, opened 2019-05-23
Simple Filler API

### LINKED ISSUE #180: Filler & decap cell API
Like #178

bkoppelmann:
Hi,
This is a first step to fix #180. Since this is my first contribution, I'd like to keep it simple and get feedback before I invest too much time. 
Cheers,
Bastian

### DIFF
--- a/src/hammer-tech/specialcells.py
+++ b/src/hammer-tech/specialcells.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+#  stackup.py
+#  Data structures to represent technology's special cells.
+#
+#  See LICENSE for licence details.
+
+from enum import Enum
+from typing import List, NamedTuple, Tuple, Dict, Optional
+from decimal import Decimal
+
+class CellType(Enum):
+    TieCell = 1
+    EndCap = 2
+    IOFiller = 3
+    StdFiller = 4
+
+    @classmethod
+    def __mapping(cls) -> Dict[str, "RoutingDirection"]:
+        return {
+            "tiecell": CellType.TieCell,
+            "endcap": CellType.EndCap,
+            "iofiller": CellType.IOFiller,
+            "stdfiller": CellType.StdFiller,
+        }
+
+    @staticmethod
+    def from_str(input_str: str) -> "CellType":
+        # pylint: disable=missing-docstring
+        try:
+            return CellType.__mapping()[input_str]
+        except KeyError:
+            raise ValueError("Invalid CellType: " + str(input_str))
+
+    def __str__(self) -> str:
+        return reverse_dict(RoutingDirection.__mapping())[self]
+
+
+class SpecialCell(NamedTuple('SpecialCell', [
+        ('cell_type', CellType),
+        ('name', str),
+        ('size', Optional[Decimal])
+])):
+    """
+    A SpecialCell is a technology cell used for non logic task (e.g. Tiecells,
+    Endcap, filler, etc.
+
+    """
+    __slots__ = ()
+
+    @staticmethod
+    def from_setting(d: dict) -> "SpecialCell":
+        size = d['size']
+        if size is not None:
+            size = Decimal(d['size'])
+        # pylint: disable=missing-docstring
+        return SpecialCell(
+            cell_type=CellType.from_str(d["cell_type"]),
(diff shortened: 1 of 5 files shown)


src/hammer-vlsi/tech_test.py
```diff
@@ -531,6 +531,46 @@ def add_lib_with_lef(d: Dict[str, Any]) -> Dict[str, Any]:
...
+        tech.cache_dir = tech_dir
+
+        tool = DummyTool()
+        tool.technology = tech
+        database = hammer_config.HammerDatabase()
+        tool.set_database(database)
+
+        self.assertEqual(tool.technology.get_special_cell_by_type(CellType.TieCell),
+                [SpecialCell(name="cell1", cell_type=CellType.TieCell, size=None),
+                 SpecialCell(name="cell2", cell_type=CellType.TieCell, size=Decimal(1.5))
```
jwright6323: `Decimal("1.5")` is preferred over `Decimal(1.5)`. In this case you get no issues because 1.5 can be exactly represented in floating point, but for other floats, you can run into issues:
```
>>> Decimal(1.5)
Decimal('1.5')
>>> Decimal(1.1)
Decimal('1.100000000000000088817841970012523233890533447265625')
```

src/hammer-vlsi/tech_test.py
```diff
@@ -531,6 +531,46 @@ def add_lib_with_lef(d: Dict[str, Any]) -> Dict[str, Any]:
...
+                ])
+
+        self.assertEqual(tool.technology.get_special_cell_by_type(CellType.IOFiller),
+                [SpecialCell(name="cell3", cell_type=CellType.IOFiller, size=Decimal(0.5)),
```
jwright6323: See above

src/hammer-tech/specialcells.py
```diff
@@ -0,0 +1,63 @@
...
+    Endcap, filler, etc.
+
+    """
+    __slots__ = ()
+
+    @staticmethod
+    def from_setting(d: dict) -> "SpecialCell":
+        size = d['size']
+        if size is not None:
+            size = Decimal(d['size'])
```
jwright6323: For this, because the actual underlying `dict` entry is a `float`, you'll either want to cast this to a string, or coerce it to a grid. This will _probably_ do we want:

```suggestion
            size = Decimal(str(d['size']))
```

The reason is posted in one of the other review comments.

src/hammer-tech/specialcells.py
```diff
@@ -0,0 +1,63 @@
...
+from decimal import Decimal
+
+class CellType(Enum):
+    TieCell = 1
+    EndCap = 2
+    IOFiller = 3
+    StdFiller = 4
+
+    @classmethod
+    def __mapping(cls) -> Dict[str, "CellType"]:
```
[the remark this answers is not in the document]
edwardcwang: It's a classmethod since you can call it with a reference to the class, whereas IIRC static methods have no such capability.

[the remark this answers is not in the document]
jwright6323: Well, static methods have no sense of state, whereas class methods can access a global state that is associated with the class rather than its objects. This is stateless which makes me think it should be a static method.

bkoppelmann:
> I think the way we plan on solving #180 is through "technology hook overrides." Rather than simply having variables that get overridden by the tech plugin, this would allow the tech plugin to automatically override hooks with their own code (e.g. filler/endcap insertion), which would avoid trying to build a one-size-fits-all hook that would live in the main CAD plugin while still being portable from project-to-project.
> 
Do you have a very brief example how such an hook might look like?

> I've gone ahead and given some code feedback, but we'll probably want to solve this a different way. If you'd like to be involved, we'd be happy to have the help.

Sure, I'm interested to get involved. I was sure that my approach has drawbacks, that's why I went for early feedback

jwright6323:
Sorry for the slow response on this; been a busy few weeks. Haven't forgotten about it though- I'll send you an example hook this week.

bkoppelmann:
> Sorry for the slow response on this; been a busy few weeks. Haven't forgotten about it though- I'll send you an example hook this week.

No problem :). You don't have to come up with something too fancy. I'm just not familiar with the hammer terminology

bkoppelmann:
The filler API is already in Hammer so let's close this.

PR #411, opened 2019-05-23
Add floorplanning checks

jwright6323:
This PR will add floorplanning checks which will be run before PNR or optionall run as a separate action.


src/hammer-vlsi/hammer_vlsi/hammer_checks.py
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+#  hammer_vlsi_impl.py
```
edwardcwang: copy-paste?

jwright6323: Yessir

Issue #368, opened 2019-04-25
Get current modules floorplan easily


colinschmidt:
It would be nice to not have to parse through `get_placement_constraints()` to find the current modules width and height.

colinschmidt:
See https://github.com/ucb-bar/hammer-cad-plugins/blob/a714e6c1e106f8b2fb48cdfc0acd970834d92ef2/par/innovus/__init__.py#L235 for an example use case

edwardcwang:
Essentially a helper function that does `get_toplevel_placement_constraint`?

colinschmidt:
Yeah something like that would be helpful. I also thought maybe you might want `get_placement_for_inst("MyInstance")` where no argument or something produces toplevel constriant. However I don't currently have a use case for that.

PR #413, opened 2019-05-25
Floorplan Constraints Overhaul. Fixes #113 and #376

### LINKED ISSUE #113: Floorplan width should be optional for hardmacros
Its not used in the code path anyways

jwright6323:
Switch margins and constraints over to Decimal and make width/height for hard macros optional.

### DIFF
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -231,31 +231,45 @@ vlsi.inputs:
   # List of placement constraints for floorplanning.
   # Each item is a struct member:
   # path (str) - Path to the given instance
-  # type (str) - One of the following:
-  # - "dummy" (does nothing with this constraint)
-  # - "placement" (creates a placement constraint for an instance)
-  # - "toplevel" (top-level chip dimensions; may only occur once, for the top-level module)
-  # - "hardmacro" (places this hard macro at a particular spot)
-  # - "hierarchical" (marks this instance as part of hierarchical place and route)
-  # - "obstruction" (creates a blockage at a particular spot; see obs_types)
-  # TODO(edwardw): add rotation
+  # - Required for all types
+  # type (str) - The type of placement constraint
+  # - Required for all types
+  # - One of the following:
+  #   - "dummy" (does nothing with this constraint)
+  #   - "placement" (creates a placement constraint for an instance)
+  #   - "toplevel" (top-level chip dimensions; may only occur once, for the top-level module)
+  #   - "hardmacro" (places this hard macro at a particular spot)
+  #   - "hierarchical" (marks this instance as part of hierarchical place and route)
+  #   - "obstruction" (creates a blockage at a particular spot; see obs_types)
+  # orientation (Optional[str]) - The orientation (MX, MY, R0, R90, R180, R270)
+  # - Optional for all types
   # x (float) - x coordinate in um
+  # - Required for all types
   # y (float) - y coordinate in um
+  # - Required for all types
   # width (float) - width in um
+  # - Required for all types, but can be auto-filled for hierarchical and hardmacro if left blank
   # height (float) - height in um
-  # margins (Optional[Margins]) - margins for the top-level module (i.e. only only for a toplevel constraint). Contains the following keys:
-  # - "left" - margin from the left side of the core PNR area to the edge of the chip
-  # - "right" - margin from the right side of the core PNR area to the edge of the chip
-  # - "top" - margin from the top side of the core PNR area to the edge of the chip
-  # - "bottom" - margin from the bottom side of the core PNR area to the edge of the chip
-  # top_layer (str) - Optional: Specifies the highest layer used by this hardmacro
-  #   If specified will be used to keep other wires away from this hardmacro
-  # layers (List[str]) - Optional: only used for obstruction type. 
-  #   If specified, a list of strings that enumerates layer(s) blocked by the obstruction otherwise all layers are blocked
-  # obs_types (List[str]) - Required for obstruction type. A list of one or more of the following:
-  # - "place" - This obstruction type stops the placement of standard cells inside it
-  # - "route" - This obstruction type stops all routing inside it
-  # - "power" - This obstruction type stops only power routing/straps inside it
+  # - Required for all types, but can be auto-filled for hierarchical and hardmacro if left blank
+  # master (Optional[str]) - The master module name of the hierarchical or hardmacro constraint
+  # - Required for hierarchical and optional for hardmacro, disallowed otherwise
+  # margins (Optional[Margins]) - margins for the top-level module.
+  # - Required for toplevel, disallowed otherwise
+  # - Contains the following keys:
+  #   - "left" - margin from the left side of the core PNR area to the edge of the chip
+  #   - "right" - margin from the right side of the core PNR area to the edge of the chip
+  #   - "top" - margin from the top side of the core PNR area to the edge of the chip
+  #   - "bottom" - margin from the bottom side of the core PNR area to the edge of the chip
+  # top_layer (str) - Specifies the highest layer used by this hierarchical or hardmacro, used to keep other wires away
+  # - Optional for hierarchical and hardmacro
+  # layers (List[str]) - A list of strings that enumerates layer(s) blocked by the obstruction otherwise all layers are blocked
+  # - Required for the obstruction type, disallowed otherwise
+  # obs_types (List[str]) - A list of the types of obstructions for the given geometry
+  # - Required for the obstruction type, disallowed otherwise
+  # - A list of one or more of the following:
+  #   - "place" - This obstruction type stops the placement of standard cells inside it
+  #   - "route" - This obstruction type stops all routing inside it
+  #   - "power" - This obstruction type stops only power routing/straps inside it
   placement_constraints: []
 
   # Don't use list mode. (str)
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -241,8 +241,17 @@ vlsi.inputs:
   #   - "hardmacro" (places this hard macro at a particular spot)
   #   - "hierarchical" (marks this instance as part of hierarchical place and route)
   #   - "obstruction" (creates a blockage at a particular spot; see obs_types)
-  # orientation (Optional[str]) - The orientation (MX, MY, R0, R90, R180, R270)
+  # orientation (Optional[str]) - The orientation
   # - Optional for all types
+  # - Valid options:
+  #   - "r0"   (standard orientation)
+  #   - "r90"  (rotated 90 degrees clockwise)
+  #   - "r180" (rotated 180 degrees)
+  #   - "r270" (rotated 270 degrees clockwise; equivalent to -90 degrees counterclockwise)
+  #   - "mx"   (mirrored about the x-axis)
+  #   - "mx90" (mirrored about the x-axis, then rotated 90 degrees clockwise)
+  #   - "my"   (mirrored about the y-axis)
+  #   - "my90" (mirrored about the y-axis, then rotated 90 degrees clockwise)
   # x (float) - x coordinate in um
   # - Required for all types
   # y (float) - y coordinate in um
@@ -256,10 +265,10 @@ vlsi.inputs:
   # margins (Optional[Margins]) - margins for the top-level module.
   # - Required for toplevel, disallowed otherwise
   # - Contains the following keys:
-  #   - "left" - margin from the left side of the core PNR area to the edge of the chip
-  #   - "right" - margin from the right side of the core PNR area to the edge of the chip
-  #   - "top" - margin from the top side of the core PNR area to the edge of the chip
-  #   - "bottom" - margin from the bottom side of the core PNR area to the edge of the chip
+  #   - "left"   (margin from the left side of the core PNR area to the edge of the chip)
+  #   - "right"  (margin from the right side of the core PNR area to the edge of the chip)
+  #   - "top"    (margin from the top side of the core PNR area to the edge of the chip)
+  #   - "bottom" (margin from the bottom side of the core PNR area to the edge of the chip)
   # top_layer (str) - Specifies the highest layer used by this hierarchical or hardmacro, used to keep other wires away
   # - Optional for hierarchical and hardmacro
   # layers (List[str]) - A list of strings that enumerates layer(s) blocked by the obstruction otherwise all layers are blocked
--- a/src/hammer-vlsi/hammer_utils/lef_utils.py
+++ b/src/hammer-vlsi/hammer_utils/lef_utils.py
@@ -8,6 +8,7 @@
 #  See LICENSE for licence details.
 
 import re
+from decimal import Decimal
 from typing import List, Optional, Tuple
 
 __all__ = ['LEFUtils']
@@ -15,7 +16,7 @@ __all__ = ['LEFUtils']
 
 class LEFUtils:
     @staticmethod
-    def get_sizes(source: str) -> List[Tuple[str, float, float]]:
+    def get_sizes(source: str) -> List[Tuple[str, Decimal, Decimal]]:
         """
         Get the sizes of all macros in the given LEF source file.
 
@@ -24,7 +25,7 @@ class LEFUtils:
         """
         lines = source.split("\n")
 
-        output = []  # type: List[Tuple[str, float, float]]
+        output = []  # type: List[Tuple[str, Decimal, Decimal]]
 
         in_propertydefinitions = False  # type: bool
         in_macro = None  # type: Optional[str]
@@ -82,8 +83,8 @@ class LEFUtils:
             if regex_search:
                 if found_size:
                     raise ValueError("Found two SIZE statements in MACRO block for {m}".format(m=in_macro))
-                width = float(regex_search.group(1))
-                height = float(regex_search.group(2))
+                width = Decimal(regex_search.group(1))
+                height = Decimal(regex_search.group(2))
                 found_size = True
 
                 # Add size to output
--- a/src/hammer-vlsi/hammer_vlsi/constraints.py
+++ b/src/hammer-vlsi/hammer_vlsi/constraints.py
@@ -353,12 +353,38 @@ class PlacementConstraintType(Enum):
 
 
 # For the top-level chip size constraint, set the margin from core area to left/bottom/right/top.
-Margins = NamedTuple('Margins', [
+class Margins(NamedTuple('Margins', [
     ('left', Decimal),
     ('bottom', Decimal),
     ('right', Decimal),
     ('top', Decimal)
-])
+])):
+
+    @staticmethod
+    def from_dict(d: dict) -> "Margins":
+        return Margins(
+            left=Decimal(str(d["left"])),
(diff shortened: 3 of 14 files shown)


jwright6323:
Need to figure out a clean way to link `HardMacro` and `Hierarchical` placements to their masters, then this will be ready for review.

jwright6323:
@colinschmidt This is ready for review. I am running a job in the background to make sure it doesn't break EAGLE, hence the `DO NOT MERGE` tag, but I will remove and merge assuming it goes through OK.

jwright6323:
Oops this doesn't actually fix #368 yet.

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -242,42 +244,138 @@ class PlacementConstraint(NamedTuple('PlacementConstraint', [
...
         if "orientation" in constraint:
+            assert constraint_type != PlacementConstraintType.TopLevel, "Non-TopLevel constraint must not contain orientation: {}".format(constraint)
             orientation = str(constraint["orientation"])
+
+        ### Top layer ###
+        # This field is optional in HardMacro constraints
+        # This field is disallowed otherwise
+        top_layer = None  # type: Optional[str]
         if "top_layer" in constraint:
+            assert constraint_type == PlacementConstraintType.HardMacro, "Non-HardMacro constraint must not contain top_layer: {}".format(constraint)
```
colinschmidt: I also use this with hierarchical blocks. There is something fancier you could do to auto use pin layers but I think we might still want this control sometimes.

jwright6323:
@colinschmidt Hmm actually I want to do a few more things. Will put this on hold a little longer.

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -242,42 +244,138 @@ class PlacementConstraint(NamedTuple('PlacementConstraint', [
...
+            assert constraint_type != PlacementConstraintType.Obstruction, "Obstruction constraint must contain obs_types: {}".format(constraint)
+
+        ### Master ###
+        # This field is mandatory for Hierarchical constraints
+        # This field is optional for HardMacro constraints
+        # This field is disallowed otherwise
+        master = None  # type: Optional[str]
+        if "master" in constraint:
+            assert constraint_type in [PlacementConstraintType.Hierarchical, PlacementConstraintType.HardMacro], "Constraints other than Hierarchical and HardMacro must not contain master: {}".format(constraint)
+            master = constraint["master"]
```
colinschmidt: Missing str cast?

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -242,42 +244,138 @@ class PlacementConstraint(NamedTuple('PlacementConstraint', [
...
+            height = height_check
+
+        # Perform the check
+        if constraint_type in checked_types:
+            assert height == height_check, "Optional height value {} for constraint must equal the master value: {} for constraint {}".format(height, height_check, constraint)
+            assert width == width_check, "Optional width value {} for constraint must equal the master value: {} for constraint {}".format(width, width_check, constraint)
+
+        ### X & Y coordinates ###
+        # These fields are mandatory in all constraints
+        assert "x" in constraint, "Constraint must contain an x coordinate: {}".format(constraint)
```
colinschmidt: Why no y assert?

src/hammer-vlsi/hammer_vlsi/driver.py
```diff
@@ -806,12 +811,26 @@ def get_hierarchical_settings(self) -> List[Tuple[str, dict]]:
...
+
             list_of_placement_constraints = self.database.get_setting(
                 "vlsi.inputs.hierarchical.manual_placement_constraints")  # type: List[Dict]
             assert isinstance(list_of_placement_constraints, list)
-            hier_modules = reduce(add_dicts, list_of_hier_modules)
-            combined_raw_placement_dict = reduce(add_dicts, list_of_placement_constraints, {})  # type: Dict[str, dict]
-            hier_placement_constraints = {key: list(map(PlacementConstraint.from_dict, lst))
+            combined_raw_placement_dict = reduce(add_dicts, list_of_placement_constraints, {})  # type: Dict[str, List[Dict[str, Any]]]
+
+            def get_toplevel(d: List[Dict[str, Any]]) -> Optional[Dict[str, Any]]:
```
colinschmidt: Is this getting all top levels? Not just one? If so maybe pluralize the name.

src/hammer-vlsi/hammer_utils/lef_utils.py
```diff
@@ -82,8 +83,8 @@ def get_sizes(source: str) -> List[Tuple[str, float, float]]:
             if regex_search:
                 if found_size:
                     raise ValueError("Found two SIZE statements in MACRO block for {m}".format(m=in_macro))
-                width = float(regex_search.group(1))
-                height = float(regex_search.group(2))
+                width = Decimal(regex_search.group(1))
```
edwardcwang: Is `Decimal(x)` a superset of `float(x)`? I think this is true but just double checking

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -219,20 +220,21 @@ def __str__(self) -> str:
...
     ('type', PlacementConstraintType),
-    ('x', float),
-    ('y', float),
-    ('width', float),
-    ('height', float),
+    ('x', Decimal),
+    ('y', Decimal),
+    ('width', Decimal),
+    ('height', Decimal),
+    ('master', Optional[str]),
```
edwardcwang: Can you document this in defaults.yml?

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -242,42 +244,138 @@ class PlacementConstraint(NamedTuple('PlacementConstraint', [
...
 
     @staticmethod
-    def from_dict(constraint: dict) -> "PlacementConstraint":
+    def from_dict(masters: List[MacroSize], constraint: dict) -> "PlacementConstraint":
         constraint_type = PlacementConstraintType.from_str(
             str(constraint["type"]))
+
+        ### Margins ###
+        # This field is mandatory in TopLevel constraints
+        # This field is disallowed otherwise
```
edwardcwang: Can we move this to defaults.yml for DRY?

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -242,42 +244,138 @@ class PlacementConstraint(NamedTuple('PlacementConstraint', [
...
+        # This field is disallowed in TopLevel constraints
+        # This field is optional otherwise
+        orientation = None  # type: Optional[str]
         if "orientation" in constraint:
+            assert constraint_type != PlacementConstraintType.TopLevel, "Non-TopLevel constraint must not contain orientation: {}".format(constraint)
             orientation = str(constraint["orientation"])
+
+        ### Top layer ###
+        # This field is optional in HardMacro constraints
+        # This field is disallowed otherwise
```
edwardcwang: Ditto all of these, unless there's a reason for duplicating them here?

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -242,42 +244,138 @@ class PlacementConstraint(NamedTuple('PlacementConstraint', [
...
+        if "master" in constraint:
+            assert constraint_type in [PlacementConstraintType.Hierarchical, PlacementConstraintType.HardMacro], "Constraints other than Hierarchical and HardMacro must not contain master: {}".format(constraint)
+            master = constraint["master"]
+        else:
+            assert constraint_type != PlacementConstraintType.Hierarchical, "Hierarchical constraint must contain master: {}".format(constraint)
+
+        ### Width & height ###
+        # These fields are mandatory for Dummy, Placement, TopLevel, and Obstruction constraints
+        # These fields are optional for HardMacro and Hierarchical constraints
+        #   If omitted, they are copied from the master definition (HardMacro) or hierarchical top-level width and height (Hierarchical).
```
edwardcwang: Not sure I followed what you meant here by `master` (see above)

src/hammer-vlsi/hammer_utils/lef_utils.py
jwright6323: I believe so - I'll double check. I assume your intent is to ensure there's no string that will cast to `float` but will not cast to `Decimal`?

src/hammer-vlsi/hammer_vlsi/constraints.py
jwright6323: yes

src/hammer-vlsi/hammer_vlsi/constraints.py
jwright6323: I will also put it in defaults.yml, but it was very helpful in writing the code (and reading it after writing), so I would rather not delete it.

src/hammer-vlsi/hammer_vlsi/constraints.py
jwright6323: Same response as above

src/hammer-vlsi/hammer_vlsi/constraints.py
jwright6323: `master` is the module or macro name (that's the reason for the addition of the `master` field in the tuple). Once that is documented, I think it'll be clearer.

src/hammer-vlsi/hammer_vlsi/driver.py
jwright6323: This is just getting one `toplevel` (it gets fed to a map function below).

src/hammer-vlsi/hammer_vlsi/constraints.py
jwright6323: oversight, will fix

src/hammer-vlsi/hammer_vlsi/constraints.py
jwright6323: I'll add it.

src/hammer-vlsi/hammer_vlsi/constraints.py
jwright6323: OK, sounds good. I'll add it back to hierarchical.

src/hammer-vlsi/hammer_utils/lef_utils.py
edwardcwang: Yeah. I think it shouldn't be an issue, just wanted to check

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -242,42 +244,138 @@ class PlacementConstraint(NamedTuple('PlacementConstraint', [
...
+        # This field is mandatory in TopLevel constraints
+        # This field is disallowed otherwise
         margins = None  # type: Optional[Margins]
-        orientation = None  # type: Optional[str]
-        top_layer = None  # type: Optional[str]
-        layers = None  # type: Optional[List[str]]
-        obs_types = None  # type: Optional[List[ObstructionType]]
-        if constraint_type == PlacementConstraintType.TopLevel:
+        if "margins" in constraint:
+            assert constraint_type == PlacementConstraintType.TopLevel, "Non-TopLevel constraint must not contain margins: {}".format(constraint)
```
edwardcwang: Nitpick-ish: could we have these be raised as `ValueError`s? (since we have more precise knowledge than just some random assertion failed)

jwright6323: Sure, that's fine. I think at some point we want a more elegant way to handle these than either assertions or exceptions (e.g. you might want to let the tool continue a little bit longer to give you multiple errors before exiting)

src/hammer-vlsi/hammer_utils/lef_utils.py
jwright6323: Appears to be the case: https://docs.python.org/3/library/decimal.html#decimal.Decimal

jwright6323:
@edwardcwang @colinschmidt all changes addressed, re-review at your leisure. I did also enforce the width/height check on the HardMacro placements if the MacroSize data for the master exists in the database.

src/hammer-vlsi/defaults.yml
```diff
@@ -231,31 +231,45 @@ vlsi.inputs:
...
+  # type (str) - The type of placement constraint
+  # - Required for all types
+  # - One of the following:
+  #   - "dummy" (does nothing with this constraint)
+  #   - "placement" (creates a placement constraint for an instance)
+  #   - "toplevel" (top-level chip dimensions; may only occur once, for the top-level module)
+  #   - "hardmacro" (places this hard macro at a particular spot)
+  #   - "hierarchical" (marks this instance as part of hierarchical place and route)
+  #   - "obstruction" (creates a blockage at a particular spot; see obs_types)
+  # orientation (Optional[str]) - The orientation (MX, MY, R0, R90, R180, R270)
```
edwardcwang: minor nitpick, could we make these lowercase like below?
```suggestion
  # orientation (Optional[str]) - The orientation (mx, my, r0, r90, r180, r270)
```

src/hammer-vlsi/hammer_vlsi/constraints.py
```diff
@@ -242,42 +244,158 @@ class PlacementConstraint(NamedTuple('PlacementConstraint', [
     __slots__ = ()
 
     @staticmethod
-    def from_dict(constraint: dict) -> "PlacementConstraint":
+    def from_dict(masters: List[MacroSize], constraint: dict) -> "PlacementConstraint":
```
edwardcwang: Hmm, could we make the `masters` parsing into a separate function? Namely I'd like all Hammer objects to be serializable and deserializable from dict<->Python. Maybe it could be a second optional argument or second override or something. As for right now this will break some of the Hammer IR stuff.

src/hammer-vlsi/defaults.yml
```diff
@@ -231,31 +231,45 @@ vlsi.inputs:
...
-  # top_layer (str) - Optional: Specifies the highest layer used by this hardmacro
-  #   If specified will be used to keep other wires away from this hardmacro
-  # layers (List[str]) - Optional: only used for obstruction type. 
-  #   If specified, a list of strings that enumerates layer(s) blocked by the obstruction otherwise all layers are blocked
-  # obs_types (List[str]) - Required for obstruction type. A list of one or more of the following:
-  # - "place" - This obstruction type stops the placement of standard cells inside it
-  # - "route" - This obstruction type stops all routing inside it
-  # - "power" - This obstruction type stops only power routing/straps inside it
+  # - Required for all types, but can be auto-filled for hierarchical and hardmacro if left blank
+  # master (Optional[str]) - The master module name of the hierarchical or hardmacro constraint
```
edwardcwang: So is the idea something like this?

Somewhere in some master list (it's not clear to me how this list ends up in Hammer yet), there's a list "my_macro -> (400, 100)".
Then we create a hardmacro instantation like
```
path: "MyTop/amplifier_inst"
type: hardmacro
x: 100
y: 100
master: my_macro
```

I think I get what's going on, would be good to elaborate on this in the documentation and add a unit test.

src/hammer-vlsi/hammer_vlsi/driver.py
```diff
@@ -806,12 +811,26 @@ def get_hierarchical_settings(self) -> List[Tuple[str, dict]]:
...
+                results = list(filter(lambda x: x["type"] == "toplevel", d))
+                if len(results) == 0:
+                    return None
+                else:
+                    return results[-1]
+
+            toplevels_opt = {k: get_toplevel(v) for k, v in combined_raw_placement_dict.items()}  # type: Dict[str, Optional[Dict[str, Any]]]
+            toplevels = {k: v for k, v in toplevels_opt.items() if v is not None}  # type: Dict[str, Dict[str, Any]]
+            hier_macros = list(map(lambda x: MacroSize(library="", name=x[0], width=x[1]["width"], height=x[1]["height"]), toplevels.items()))
```
edwardcwang: Is this part of the code unit tested?

src/hammer-vlsi/defaults.yml
jwright6323: I have never seen a CAD tool use lower case for these rotations. I don't think they should change.

src/hammer-vlsi/hammer_vlsi/driver.py
jwright6323: No

src/hammer-vlsi/hammer_vlsi/constraints.py
jwright6323: Hmm that's kind of a major rewrite, but let me see what I can do. Basically, I would create a new function that has these same args, and do the `master` resolving there, then pass that to the `from_dict` method which would then make `width` and `height` mandatory again.

src/hammer-vlsi/defaults.yml
jwright6323: `master` is a pretty common name for the module an instance is an instance of; I don't really think it needs to be elaborated. I can go into more detail about where the info comes from, though. For `hierarchical` it comes from the `toplevel` placement constraint of that block, and for `hardmacro` it comes from `vlsi.technology.extra_macro_sizes` (hence the discussion yesterday).

jwright6323: (Forgot to add that I will put this in the documentation)

edwardcwang: Maybe if you can just put the example in the documentation that may be good enough. I had to do a double take to really understand what this example is doing, and I'd like for Hammer to be as easy to use as possible. If it's too much work for you I can help

src/hammer-vlsi/hammer_vlsi/driver.py
edwardcwang: Is it hard to add one? Again if it's too much work I can help

src/hammer-vlsi/hammer_vlsi/constraints.py
edwardcwang: Sounds good to me.

src/hammer-vlsi/defaults.yml
edwardcwang: Let's discuss offline.

colinschmidt:
It's actually about painting the bike shed

src/hammer-vlsi/hammer_vlsi/constraints.py
jwright6323: Resolved, please take a look.

src/hammer-vlsi/defaults.yml
jwright6323: Added some documentation, please take a look. Working on adding a unit test.

src/hammer-vlsi/hammer_vlsi/constraints.py
edwardcwang: Looks good, thanks!

src/hammer-vlsi/hammer_vlsi/driver.py
jwright6323: I'll work on this after my talk, then I think it's G2G.

jwright6323: OK pushed some unit tests.

edwardcwang:
Will take a look as soon as possible

PR #415, opened 2019-05-28
Add blockage_spacing to standard cell power strap placement

jwright6323:
This fixes power strap issues we ran into at the top level by passing blockage_spacing to the method that builds std cell rails.


jwright6323:
@dpgrubb13 Can you verify this

jwright6323:
Putting a link here for bookkeeping: ucb-bar/hammer-cad-plugins#111

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -556,10 +556,15 @@ def specify_all_power_straps_by_tracks(self, layer_names: List[str], ground_net:
         for l in pin_layers:
             assert l in layer_names, "Pin layer {} must be in power strap layers".format(l)
 
+        rail_layer_name = self.get_setting("technology.core.std_cell_rail_layer")
+        rail_layer = self.get_stackup().get_metal(rail_layer_name)
+        blockage_spacing = coerce_to_grid(float(self._get_by_tracks_metal_setting("blockage_spacing", rail_layer_name)), rail_layer.grid_unit)
```
colinschmidt: No defaults needed to change with this now?

jwright6323: Well the blockage spacing fields were already there, so this should work the same as what exists now.

Issue #378, opened 2019-04-30
Use hammer_units for physical lengths


jwright6323:
Sorry for the blast of comments on this one. Maybe we can convert to `hammer_units` in a future PR? I have filed #376 to convert all our internal units to decimals- maybe it makes sense to combine these two changes?

_Originally posted by @jwright6323 in https://github.com/ucb-bar/hammer/pull/370_

PR #421, opened 2019-06-04
Fix JSON serialization bug with Decimals

jwright6323:
One of the CLI driver `json.dumps` needed the special Hammer serializer to work with Decimals. I think we should get rid of this when we solve #378 by not serializing `Decimal`s directly.

### DIFF
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -452,7 +452,7 @@ class HammerTechnology:
             raise TypeError("lib must be a dict")
 
         # Convert the dict to JSON...
-        return Library.from_json(json.dumps(lib), cls=HammerJSONEncoder)
+        return Library.from_json(json.dumps(lib, cls=HammerJSONEncoder))
 
     @property
     def tech_defined_libraries(self) -> List[Library]:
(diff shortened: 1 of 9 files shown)


jwright6323:
we're actively debugging, so gunna slap a `DO NOT MERGE` on this

edwardcwang:
Thanks for catching this! I would be interested to know if there is a way to fix it less invasively than manually adding the cls parameter everywhere in the code?

jwright6323:
I don't like the `cls` thing either, which is why I suggested this:
> I think we should get rid of this when we solve #378 by not serializing Decimals directly.

Basically, if we use a wrapper class around our length units and/or use the hammer units classes (which I assume can be "taught" to serialize to strings), we won't need to serialize `Decimal`s.

Other than that, I don't know of a way other than using `simplejson` as recommended here:
https://stackoverflow.com/questions/1960516/python-json-serialize-a-decimal-object

edwardcwang:
That makes sense. Another option (e.g. if we ever need to do something like this for other types) is to create a wrapper in `hammer_utils` that wraps `json.dumps()` with our desired settings.

jwright6323:
Yeah, I also thought about that a bit, but I mostly wanted to fix the extant master bug first, then clean it up later. I'd be ok with a hammer wrapper around `json.dumps`; that seems like the best cleanliness-to-effort ratio at the current moment.

edwardcwang:
Sounds good :) Will file an issue then

PR #420, opened 2019-06-04
Add PCB deliverables, fixes #122 and #364

jwright6323:
**Metadata: rebase merge**

This PR will add a `HammerPCBDeliverableTool` and an implementation to create an open version of footprints and schematic symbols

PR #425, opened 2019-06-08
PCB bump pitch should use post-shrink units

jwright6323:
When merging the tech plugin after #420, I realized that I didn't scale the pitch in the PCB footprint.


jwright6323:
Thanks for the quick review.

PR #426, opened 2019-06-08
modify test framework and remove vending libraries.

sequencer:
1. use `pytest` as test framework to give a pretty test output.
2. change `src/hammer-vlsi/test.py ` to `src/hammer-vlsi/vlsi_test.py ` to meet `pytest` name requirement.
3. use monkey patch patching upstream `python_jsonschema_objects`, avoid vending old libraries.
4. use `pip install --user` for 3rd party libraries installation, being able to run in non-root privilege.


sequencer:
Seem to CI has different python in the `/opt` path. `pip` might have problem in this case.

sourceme.sh
```diff
@@ -1,10 +1,17 @@
 #!/bin/sh
 # Convenience script for setting variables for in-tree use of hammer.
+echo "Installing dependency"
+python3 -m pip install --user typing six enum34 python-jsonschema-objects pyaml pytest
```
edwardcwang: Could you move this to a separate script maybe named `setup.sh` instead? This script is sourced pretty frequently and it would be good to avoid a potentially expensive run of pip each time.

sourceme.sh
```diff
@@ -1,10 +1,17 @@
...
 if [ -z "$HAMMER_HOME" ]; then
     >&2 echo "Must set HAMMER_HOME to root. Try export HAMMER_HOME="\$PWD" for the current dir."
 else
+echo "Setting PYTHONPATH and PATH"
```
edwardcwang: Could you avoid these `echo`es in the script?

src/hammer-tech/hammer_tech.py
```diff
@@ -30,6 +31,7 @@
 # Access it like hammer_tech.filters.lef_filter
 filters = LibraryFilterHolder()
 
+
```
edwardcwang: Unnecessary extra line

src/test/unittests.sh
```diff
@@ -5,14 +5,14 @@
...
-python3 ../hammer-vlsi/test.py
-python3 ../hammer-vlsi/tech_test.py
-python3 ../hammer-vlsi/constraints_test.py
-python3 ../hammer-vlsi/cli_driver_test.py
-python3 ../hammer-vlsi/utils_test.py
-python3 ../hammer-vlsi/units_test.py
-python3 ../hammer-vlsi/verilog_utils_test.py
-python3 ../hammer-vlsi/lef_utils_test.py
-python3 ../hammer_config_test/test.py
+python3 -m pytest ../hammer-vlsi/vlsi_test.py
```
edwardcwang: Minor nitpick, can you set up the test files so they can still be run via `python3 ../hammer-vlsi/vlsi_test.py`?

sourceme.sh
edwardcwang: Also, could you pin these to version ranges e.g. like in https://stackoverflow.com/questions/8795617/how-to-pip-install-a-package-with-min-and-max-version-range

.gitmodules
```diff
@@ -1,13 +1,3 @@
 [submodule "src/tools/firrtl"]
 	path = src/tools/firrtl
 	url = https://github.com/ucb-bar/firrtl.git
-[submodule "src/tools/pyyaml"]
```
edwardcwang: Hmm. While I'm not opposed to using pip and the other standard tools, I do want to make sure that this can continue to work out of the box on rusty CAD machines since that is the type of environment where Hammer will run. Maybe we can move these repos to `contrib/` (preferred) or `vending/` or some other prefix which we can use if Hammer is cloned onto an old CAD machine?

edwardcwang:
I think I see what the issue with the CI is and will fix within the next few days.

.gitmodules
sequencer: Oh, I understand. I have test this on CentOS 6 and CentOS 7 to make it work in the old environment.
If you would like to use `contrib/` to vend these libraries, how about I give 2 flows, one for vending, another for `pip` in 2 different `sourceme.sh`?

edwardcwang: That sounds reasonable - maybe let's try that and see what happens.

src/test/unittests.sh
sequencer: OK, I'll fix this, and `pytest` will be still work in single command `pytest` too.

src/hammer-tech/hammer_tech.py
sequencer: Sorry for that, I didn't check it carefully.

sourceme.sh
sequencer: OK!

sourceme.sh
sequencer: OK I will use `requirements.txt` to handle the 3rd party package list, and this will introduce 2 more files. :)

.gitmodules
sequencer: OK, working on it now :)

sequencer:
Emmm, I'm fighting with CI, please ignore the annoying force commits, I'll squash all commits finally.

sequencer:
Emmm, I thought I did no wrong, maybe is CI's problem.

jwright6323:
At a brief glance this looks like it would mess up our flow, so I'd like to review this whenever it's ready.

sequencer:
> At a brief glance this looks like it would mess up our flow, so I'd like to review this whenever it's ready.

I'm so sorry about that. I wanna make hammer also work fine in some modern distributions like CentOS7, Debian 8+, even Arch Linux, let community can participate in this project. So I'm trying to port a standard python work flow to hammer. I will also try my best not to mess your flow up.
And maybe the most dangerous thing is that monkey patch(but I have already to make this clean, and pass all test).
As for the dependencies management, I'll also keep the 2 flows both work fine in `pip` and `git submodule` flow, the only difference is `git submodule` remote path, and libraries version. But because of I have already finished the monkey patch of `python-jsonschema-objects`, pass all test after updating 3rd party libraries to latest version, so I think it won't introduce more influence.
Maybe it might introduce some potential bugs, I hope you can test this patch with full hammer flow to avoid that.
and hunk below:
```
-export PYTHONPATH="$HAMMER_HOME/src:$HAMMER_HOME/src/jsonschema:$HAMMER_HOME/src/python-jsonschema-objects:$HAMMER_HOME/src/hammer-tech:$HAMMER_HOME/src/hammer-vlsi:$HAMMER_HOME/src/tools/pyyaml/lib3:$PYTHONPATH"
+export PYTHONPATH="$HAMMER_HOME/src:$HAMMER_HOME/src/hammer-tech:$HAMMER_HOME/src/hammer-vlsi:$PYTHONPATH"
+# if import failed, then use vending version
+python3 -c "import jsonschema, python_jsonschema_objects, yaml" > /dev/null 2>&1
+if [[ $? == 1 ]]; then
+    export PYTHONPATH="$HAMMER_HOME/src/contrib/jsonschema:$HAMMER_HOME/src/contrib/python-jsonschema-objects:$HAMMER_HOME/src/contrib/pyyaml:$PYTHONPATH"
+fi
```
can make sure if system doesn't provide these libraries, user can directly use `git submodule` version.

jwright6323:
well, I don't know for sure that it breaks yet, but whenever CI is passing I will give this a shot on our environment and let you know.

sequencer:
Thank you!
If there are any bugs introduced by this patch, I’ll fix it immediately.

sequencer:
Use branch `origin/pip_pytest_support` to support this, will reopen this in another PR after ready to merge.

PR #428, opened 2019-06-10
Add physical only cell list API

colinschmidt:
To be used when creating an LVS netlist

### DIFF
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -407,7 +407,7 @@ class HammerTechnology:
     @property
     def physical_only_cells_list(self) -> Optional[List[str]]:
         """
-        Get the list of physical only  cells.
+        Get the list of physical only cells.
         :return: List of physical only cells, or None if the technology does not define such a list.
         """
         physical_only_cells_list_raw = self.config.physical_only_cells_list  # type: Optional[List[str]]
@@ -415,7 +415,7 @@ class HammerTechnology:
             return None
         else:
             # Work around the weird objects implemented by the jsonschema generator.
-            physical_only_cells_list = list(map(lambda x: str(x), list(physical_only_cells_list_raw)))
+            physical_only_cells_list = [str(x) for x in physical_only_cells_list_raw]
             return physical_only_cells_list
 
     @property
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -405,6 +405,20 @@ class HammerTechnology:
             return dont_use_list
 
     @property
+    def physical_only_cells_list(self) -> Optional[List[str]]:
+        """
+        Get the list of physical only  cells.
+        :return: List of physical only cells, or None if the technology does not define such a list.
+        """
+        physical_only_cells_list_raw = self.config.physical_only_cells_list  # type: Optional[List[str]]
+        if physical_only_cells_list_raw is None:
+            return None
+        else:
+            # Work around the weird objects implemented by the jsonschema generator.
+            physical_only_cells_list = list(map(lambda x: str(x), list(physical_only_cells_list_raw)))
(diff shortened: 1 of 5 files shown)


colinschmidt:
I'm currently using this in my favorite technology, and it solves my use case.

src/hammer-tech/hammer_tech.py
```diff
@@ -404,6 +404,20 @@ def dont_use_list(self) -> Optional[List[str]]:
...
+        """
+        Get the list of physical only  cells.
+        :return: List of physical only cells, or None if the technology does not define such a list.
+        """
+        physical_only_cells_list_raw = self.config.physical_only_cells_list  # type: Optional[List[str]]
+        if physical_only_cells_list_raw is None:
+            return None
+        else:
+            # Work around the weird objects implemented by the jsonschema generator.
+            physical_only_cells_list = list(map(lambda x: str(x), list(physical_only_cells_list_raw)))
```
edwardcwang:
```suggestion
            physical_only_cells_list = list(map(str, list(physical_only_cells_list_raw)))
```

src/hammer-tech/hammer_tech.py
```diff
@@ -404,6 +404,20 @@ def dont_use_list(self) -> Optional[List[str]]:
             dont_use_list = list(map(lambda x: str(x), list(dont_use_list_raw)))
             return dont_use_list
 
+    @property
+    def physical_only_cells_list(self) -> Optional[List[str]]:
+        """
+        Get the list of physical only  cells.
```
edwardcwang:
```suggestion
        Get the list of physical only cells.
```

src/hammer-vlsi/defaults.yml
```diff
@@ -454,6 +454,21 @@ par.inputs:
   # type: bool
   gds_merge: false
 
+  # Physical only cells mode.
+  # type: str
+  # The list of physical only cells are cells which don't have a logic function.
+  # They won't have a spice circuit for instance.
```
```suggestion
  # They won't have a SPICE circuit for instance.
```

src/hammer-vlsi/defaults.yml
```diff
@@ -454,6 +454,21 @@ par.inputs:
...
+  # This is currently only used for what cells to exclude from the LVS netlist, which
+  # requires spice circuits for all cells.
```
```suggestion
  # requires SPICE circuits for all cells.
```

src/hammer-vlsi/defaults.yml
```diff
@@ -454,6 +454,21 @@ par.inputs:
   # type: bool
   gds_merge: false
 
+  # Physical only cells mode.
+  # type: str
+  # The list of physical only cells are cells which don't have a logic function.
+  # They won't have a spice circuit for instance.
+  # This is currently only used for what cells to exclude from the LVS netlist, which
```
edwardcwang: Does this include filler/decap cells? May be related to #180 

Also, how does it affect LVS? Does it mean that if a physical-only cell causes a short, it won't be detected?

src/hammer-tech/hammer_tech.py
jwright6323: Getting into the holy war a bit:
```suggestion
            physical_only_cells_list = [str(x) for x in physical_only_cells_list_raw]
```

src/hammer-vlsi/defaults.yml
jwright6323: Well, filler and decap cells can and should be treated differently here. Filler cells we remove- they have no LVS-able devices in them. Decap cells do. LVS should still detect a short.

src/hammer-tech/hammer_tech.py
edwardcwang: Just FYI that `map(func, ...)` is faster than its list comprehension equivalent as opposed to `map(lambda x: <custom logic>, ...)` so would be good to favour `map(str, ...)` in this case

PR #434, opened 2019-06-13
Add an additional PCB deliverable that we need for our PCB flow

jwright6323:
Also makes a few code structure changes to be less repetitive

### DIFF
--- a/src/hammer-vlsi/constraints_test.py
+++ b/src/hammer-vlsi/constraints_test.py
@@ -208,7 +208,7 @@ class BumpsTest(unittest.TestCase):
         self.assertEqual(sorted_assignments, [assignments[x] for x in sorted(idxs)])
 
         sorted_assignments = BumpsPinNamingScheme.A0.sort_by_name(definition, assignments)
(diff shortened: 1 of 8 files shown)


jwright6323:
I think this is good to go (tests should pass after a typo fix), but I'm going to hold onto it until I get the green light from our board expert that the views look good

jwright6323:
This appears to be correct, now. Ready to merge after a review.

jwright6323:
I unfortunately don't have a lot of answers for this- I would have liked to include a bit more documentation but I'm not familiar with what this file format is. Our PCB expert send me a few examples and "documentation" in the form of a commented version of this file, which I mostly translated into inline python comments here.

As for how you use it, it just gets spit out when you run `pcb.generic`. There are no inputs or configs, and it should not assert if other deliverables are also not asserting. There is a unit tests, but it is just an existence check (like the other PCB deliverables at this point). I don't have the bandwidth to write rigorous tests for the PCB deliverables right now.

jwright6323:
Thanks, issues are good.

PR #436, opened 2019-06-13
Add a create field

jwright6323:
Add a 'create' field to PlacementConstraint that will allow the PNR tool to create instances that are not in the source netlist (e.g. physical-only cells)

### DIFF
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -273,6 +273,8 @@ vlsi.inputs:
   #                          For example, if this hardmacro is an instance of "my_amplifier", this
   #                          field should be set to "my_amplifier".
   # - Required for hierarchical and optional for hardmacro, disallowed otherwise
+  # create (Optional[bool]) - Create an instance that does not exist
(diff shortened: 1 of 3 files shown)


jwright6323:
@dpgrubb13 to test and sign off on this and its companion

src/hammer-vlsi/defaults.yml
```diff
@@ -273,6 +273,8 @@ vlsi.inputs:
   #                          For example, if this hardmacro is an instance of "my_amplifier", this
   #                          field should be set to "my_amplifier".
   # - Required for hierarchical and optional for hardmacro, disallowed otherwise
+  # create (Optional[bool]) - Create an instance that does not exist
```
colinschmidt: Add "In addition to placing the instance".

jwright6323: fixed

jwright6323:
@dpgrubb13 can you test this once again?

dpgrubb13:
I retested after pulling and I got the same result.

PR #438, opened 2019-06-13
Add a TCLTool mixin

jwright6323:
Proposed in ucb-bar/hammer-cad-plugins#116, this moves the `append` and `verbose_append` methods to a mixin trait that is then used by `CadenceTool` and could be used in the future by other tool types.

### DIFF
--- a/src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
+++ b/src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
@@ -1064,8 +1064,25 @@ class HasSDCSupport(HammerTool):
         """
         pass
 
+class TCLTool(HammerTool):
+    """Mix-in trait for tools which consume a flat TCL file as input"""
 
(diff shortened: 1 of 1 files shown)


edwardcwang:
@jwright6323 Follow-up thought: we might want to change the hook types to be `x: TCLTool` rather than `x: HammerTool` to allow type checker to run properly on hook files as well

jwright6323:
That was actually the motivation for this :)

jwright6323:
@edwardcwang unfortunately I hit an unexpected roadblock- `make_*_hook` expects a new hook with type signature `func(x: HammerTool) -> bool`, and passing it a function taking `x: TCLTool` makes it unhappy. We can't change the type signature of hooks because you may want a non-`TCLTool` hook, so we'll need to spend a bit more thought energy into this. For now, my favorite_technology_hooks.py file just has some crude `assert isinstance(x, ...)` commands at the beginning of the function.

Issue #309, opened 2019-02-14
ASAP PDK plugin for Hammer


edwardcwang:
Related: #177

http://asap.asu.edu/asap/

donggyukim:
This looks very interesting

harrisonliew:
Closed by #467, #499, #512, and #511.

PR #443, opened 2019-06-15
experimentally add asap tech.json

sequencer:
Trying to resolve https://github.com/ucb-bar/hammer/issues/309 to avoid some NDA problems in `sade32`.
Here remains some questions to be resolved. I not sure the path is right or not. Hope it works :p


jwright6323:
Nice- this has been a desired feature for us for a while now. We are in the middle of a tapeout right now; when that is done in early July, we can give this a more thorough review.

sequencer:
re-open at https://github.com/ucb-bar/hammer/pull/467

PR #445, opened 2019-06-15
recursively extract and chmod

sequencer:
1. use python internal `tarfile` and `os.chmod` to get rid of external command.
2. `asap7` https://github.com/ucb-bar/hammer/pull/443 contains recursive tarball, this patch make `extract_tarballs` extract tarball recursively.

### DIFF
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -8,7 +8,7 @@
 
 import json
 import os
-import subprocess
+import tarfile
 from abc import ABCMeta, abstractmethod
 from typing import Any, Callable, Iterable, List, NamedTuple, Optional, Tuple, Dict
 from decimal import Decimal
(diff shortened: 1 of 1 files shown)


sequencer:
Oh I forget that `fstring` was introduced to python since 3.6.
My fault.

jwright6323:
> Oh I forget that `fstring` was introduced to python since 3.6.
> My fault.

We actually run the unit tests on 3.3-3.7, so we can't use `fstring`s, although we do prefer `.format(...)` to `%` formatting

sequencer:
OK, I'll use `.format` for next PR

jwright6323:
I mean, I personally don't care a whole lot, but that's just what we seem to have settled on. As a side note, @edwardcwang are we using python3.3-3.5 anywhere? I know at least we are using 3.6- are you using something below that? Does it make sense to remove 3.3-3.5 support to enable `fstring`s? Much of what hammer does is string interpolation, so it would be a lot cleaner to use `fstring`s instead of having a bunch of `"...".format(foo=foo, bar=bar)` lying around.

edwardcwang:
I personally recall using Hammer on at least on 3.4 machine. Normally I would say let's remove support for older versions, but since running on antiquated CAD machines is important for Hammer's usecases (and it's not always easy to compile Python3.6 on antiquated CAD machines) I'd say let's keep it for now :-(

Maybe we could write a compiler pass later to remove fstrings and put in `.format` calls later or something if it's getting really annoying...

sequencer:
if you really wanna use `fstring` for simple string formatting. How about giving a try to a [3rd party backport](https://github.com/asottile/future-fstrings), I didn't use it before, so I'm not guarantee that will work fine in hammer usecase.

sequencer:
any updates?
:)

PR #446, opened 2019-06-16
Select DRC checks

jwright6323:
Add a field to specify specific checks to run (blank = all)

### DIFF
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -587,16 +587,25 @@ mentor:
 # DRC settings
 drc.inputs:
   # Top RTL module.
+  # type: str
   top_module: null
 
   # Input layout file
   # Typically a GDSII file from the place and route tool
+  # type: str
   layout_file: null
 
   # Valid modes are auto, manual, prepend, and append
+  # type: str
   additional_drc_text_mode: "auto"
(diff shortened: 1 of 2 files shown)


jwright6323:
Bumping this one too.

src/hammer-vlsi/defaults.yml
```diff
@@ -602,17 +602,26 @@ mentor:
...
 
   # Valid modes are auto, manual, prepend, and append
+  # type: str
   additional_drc_text_mode: "auto"
   # Custom DRC command text to add after the boilerplate commands at the top of the run file
+  # type: str
   additional_drc_text: ""
 
+  # DRC rules to run
+  # Leaving this empty will run all rules- otherwise the rules listed will be run
```
edwardcwang: @jwright6323 Is the idea that every tool/technology combination will define its own naming schema for rules? Also, can there be an option for certain rules `not` to run?

jwright6323: I don't think selectively turning off rules is likely to be very useful. It's far more common to do it the other way around. Hammer is not and should not do any defining of rule names- this is in the DRM, and this list is just a list of rules to run from the DRM. It's tool agnostic- when you get multiple DRC decks for different tool vendors, the rules generally still have the same names. There's no "schema" either- the vendor may have some pattern they use, but this is not intended to understand that.

PR #447, opened 2019-06-16
DRC/LVS additional text refactor

jwright6323:
This will be an attempt to make flows with multiple DRC decks work a little better. I have a solution for the current tapeout, so this will just sit here until that's done.


harrisonliew:
Basic way of specifying rule decks supplied in PDK (with relative paths). Tested with ASAP7 and can name decks for the tool plugins to use later.

edwardcwang:
@harrisonliew Could we add a unit test?

PR #448, opened 2019-06-17
First pass at a Makefile generator, fixes #234 and #453

### LINKED ISSUE #234: Figure out dependency tracking/integration with build tools
We should have some way of generating Makefiles or includes to improve build management (we don't want to reinvent the wheel here)

jwright6323:
This is still very beta, but it seems to work.

### DIFF
--- a/src/hammer-vlsi/cli_driver_test.py
+++ b/src/hammer-vlsi/cli_driver_test.py
@@ -9,8 +9,9 @@ import json
 import os
 import shutil
 import tempfile
+import re
 from decimal import Decimal
-from typing import Any, Callable, Dict, Optional
+from typing import Any, Callable, Dict, List, Optional
 
 import hammer_config
 from hammer_config import HammerJSONEncoder
@@ -534,6 +535,19 @@ class CLIDriverTest(unittest.TestCase):
 
 class HammerBuildSystemsTest(unittest.TestCase):
 
+    def _read_makefile_lines(self, lines: List[str]) -> Dict[str, List[str]]:
+        targets = {}  # type: Dict[str, List[str]]
+
+        for line in lines:
+            m = re.match(r"^([^.][^\s:]+)\s*:(.*)$", line)
+            if m:
+                t = m.group(1)
+                p = re.split(r"\s+", m.group(2))
+                self.assertFalse(t in targets)
+                targets[t] = p
+
+        return targets
+
     def test_flat_makefile(self) -> None:
         tmpdir = tempfile.mkdtemp()
         proj_config = os.path.join(tmpdir, "config.json")
@@ -564,7 +578,16 @@ class HammerBuildSystemsTest(unittest.TestCase):
 
         contents = open(d_file).readlines()
 
-        # TODO: Actually check the contents of hammer.d
+        targets = self._read_makefile_lines(contents)
+
+        tasks = {"pcb", "syn", "par", "drc", "lvs"}
+        expected_targets = tasks.copy()
+        expected_targets.update({os.path.join(tmpdir, x + "-rundir", x + "-output-full.json") for x in tasks})
+        expected_targets.update({os.path.join(tmpdir, x + "-input.json") for x in tasks if x not in {"syn", "pcb"}})
+
+        self.assertEqual(set(targets.keys()), set(expected_targets))
+
+        # TODO at some point we should add more tests
 
         # Cleanup
         shutil.rmtree(tmpdir)
@@ -613,7 +636,25 @@ class HammerBuildSystemsTest(unittest.TestCase):
 
         contents = open(d_file).readlines()
 
-        # TODO: Actually check the contents of hammer.d
+        targets = self._read_makefile_lines(contents)
+
+        mods = {"TopMod", "SubModA", "SubModB"}
+        expected_targets = {"pcb", os.path.join(tmpdir, "pcb-rundir", "pcb-output-full.json")}
+        expected_targets.update({"syn-" + x for x in mods})
+        expected_targets.update({"par-" + x for x in mods})
+        expected_targets.update({"lvs-" + x for x in mods})
+        expected_targets.update({"drc-" + x for x in mods})
+        expected_targets.update({os.path.join(tmpdir, "syn-" + x, "syn-output-full.json") for x in mods})
+        expected_targets.update({os.path.join(tmpdir, "par-" + x, "par-output-full.json") for x in mods})
+        expected_targets.update({os.path.join(tmpdir, "lvs-" + x, "lvs-output-full.json") for x in mods})
+        expected_targets.update({os.path.join(tmpdir, "drc-" + x, "drc-output-full.json") for x in mods})
+        expected_targets.update({os.path.join(tmpdir, "par-" + x + "-input.json") for x in mods})
+        expected_targets.update({os.path.join(tmpdir, "lvs-" + x + "-input.json") for x in mods})
+        expected_targets.update({os.path.join(tmpdir, "drc-" + x + "-input.json") for x in mods})
+
+        self.assertEqual(set(targets.keys()), expected_targets)
+
+        # TODO at some point we should add more tests
 
         # Cleanup
         shutil.rmtree(tmpdir)
--- a/src/hammer-vlsi/cli_driver_test.py
+++ b/src/hammer-vlsi/cli_driver_test.py
@@ -582,7 +582,8 @@ class HammerBuildSystemsTest(unittest.TestCase):
 
         tasks = {"pcb", "syn", "par", "drc", "lvs"}
         expected_targets = tasks.copy()
-        expected_targets.update({os.path.join(tmpdir, x + "-rundir", x + "-output-full.json") for x in tasks})
+        expected_targets.update({"redo-par", "redo-drc", "redo-lvs"})
+        expected_targets.update({os.path.join(tmpdir, x + "-rundir", x + "-output.json") for x in tasks})
         expected_targets.update({os.path.join(tmpdir, x + "-input.json") for x in tasks if x not in {"syn", "pcb"}})
 
         self.assertEqual(set(targets.keys()), set(expected_targets))
@@ -639,15 +640,18 @@ class HammerBuildSystemsTest(unittest.TestCase):
         targets = self._read_makefile_lines(contents)
 
         mods = {"TopMod", "SubModA", "SubModB"}
-        expected_targets = {"pcb", os.path.join(tmpdir, "pcb-rundir", "pcb-output-full.json")}
+        expected_targets = {"pcb", os.path.join(tmpdir, "pcb-rundir", "pcb-output.json")}
         expected_targets.update({"syn-" + x for x in mods})
         expected_targets.update({"par-" + x for x in mods})
         expected_targets.update({"lvs-" + x for x in mods})
         expected_targets.update({"drc-" + x for x in mods})
-        expected_targets.update({os.path.join(tmpdir, "syn-" + x, "syn-output-full.json") for x in mods})
-        expected_targets.update({os.path.join(tmpdir, "par-" + x, "par-output-full.json") for x in mods})
-        expected_targets.update({os.path.join(tmpdir, "lvs-" + x, "lvs-output-full.json") for x in mods})
-        expected_targets.update({os.path.join(tmpdir, "drc-" + x, "drc-output-full.json") for x in mods})
+        expected_targets.update({"redo-par-" + x for x in mods})
+        expected_targets.update({"redo-lvs-" + x for x in mods})
+        expected_targets.update({"redo-drc-" + x for x in mods})
+        expected_targets.update({os.path.join(tmpdir, "syn-" + x, "syn-output.json") for x in mods})
+        expected_targets.update({os.path.join(tmpdir, "par-" + x, "par-output.json") for x in mods})
+        expected_targets.update({os.path.join(tmpdir, "lvs-" + x, "lvs-output.json") for x in mods})
+        expected_targets.update({os.path.join(tmpdir, "drc-" + x, "drc-output.json") for x in mods})
         expected_targets.update({os.path.join(tmpdir, "par-" + x + "-input.json") for x in mods})
         expected_targets.update({os.path.join(tmpdir, "lvs-" + x + "-input.json") for x in mods})
         expected_targets.update({os.path.join(tmpdir, "drc-" + x + "-input.json") for x in mods})
--- a/src/hammer-vlsi/cli_driver_test.py
+++ b/src/hammer-vlsi/cli_driver_test.py
@@ -16,7 +16,7 @@ import hammer_config
 from hammer_config import HammerJSONEncoder
 from hammer_logging.test import HammerLoggingCaptureContext
 from hammer_tech import MacroSize
-from hammer_vlsi import CLIDriver, HammerDriver, PlacementConstraint, PlacementConstraintType
+from hammer_vlsi import CLIDriver, HammerDriver, HammerDriverOptions, HammerVLSISettings, PlacementConstraint, PlacementConstraintType
 from hammer_utils import deepdict
 
 import unittest
@@ -530,6 +530,98 @@ class CLIDriverTest(unittest.TestCase):
                 def synthesis_action(self, bad: int) -> dict:  # type: ignore
                     return {bad: "bad"}
             BadOverride()
+
+
+class HammerBuildSystemsTest(unittest.TestCase):
+
+    def test_flat_makefile(self) -> None:
+        tmpdir = tempfile.mkdtemp()
+        proj_config = os.path.join(tmpdir, "config.json")
+
+        config_dict = {}
+
+        settings = {
+                "vlsi.core.technology": "nop",
+                "vlsi.core.build_system": "make",
+                "synthesis.inputs.top_module": "TopMod"
+            }
+        with open(proj_config, "w") as f:
+            f.write(json.dumps(settings, cls=HammerJSONEncoder, indent=4))
+
+        options = HammerDriverOptions(
+            environment_configs=[],
+            project_configs=[proj_config],
+            log_file=os.path.join(tmpdir, "log.txt"),
+            obj_dir=tmpdir
+        )
+
+        self.assertTrue(HammerVLSISettings.set_hammer_vlsi_path_from_environment(),
+            "hammer_vlsi_path must exist")
+        driver = HammerDriver(options)
+
+        CLIDriver.generate_build_inputs(driver, lambda x: None)
+
+        d_file = os.path.join(driver.obj_dir, "hammer.d")
+        self.assertTrue(os.path.exists(d_file))
+
+        contents = open(d_file).readlines()
+
+        # TODO: Actually check the contents of hammer.d
+
+        # Cleanup
+        shutil.rmtree(tmpdir)
+
+    def test_hier_makefile(self) -> None:
+        tmpdir = tempfile.mkdtemp()
+        proj_config = os.path.join(tmpdir, "config.json")
+
+        config_dict = {}
+
+        settings = {
+                "vlsi.core.technology": "nop",
+                "vlsi.core.build_system": "make",
+                "vlsi.inputs.hierarchical.mode": "hierarchical",
+                "vlsi.inputs.hierarchical.top_module": "TopMod",
+                "vlsi.inputs.hierarchical.config_source": "manual",
+                "vlsi.inputs.hierarchical.manual_modules": [{"TopMod": ["SubModA", "SubModB"]}],
+                "vlsi.inputs.hierarchical.manual_placement_constraints": [
+                    {"TopMod": [
+                        {"path": "top", "type": "toplevel", "x": 0, "y": 0, "width": 1234, "height": 7890, "margins": {"left": 1, "top": 2, "right": 3, "bottom": 4}},
+                        {"path": "top/C", "type": "placement", "x": 2, "y": 102, "width": 30, "height": 40},
+                        {"path": "top/B", "type": "hierarchical", "x": 10, "y": 30, "master": "SubModB"},
+                        {"path": "top/A", "type": "hierarchical", "x": 200, "y": 120, "master": "SubModA"}]},
+                    {"SubModA": [
+                        {"path": "a", "type": "toplevel", "x": 0, "y": 0, "width": 100, "height": 200, "margins": {"left": 0, "top": 0, "right": 0, "bottom": 0}}]},
+                    {"SubModB": [
+                        {"path": "b", "type": "toplevel", "x": 0, "y": 0, "width": 340, "height": 160, "margins": {"left": 0, "top": 0, "right": 0, "bottom": 0}}]}
+                ]
+            }
+        with open(proj_config, "w") as f:
+            f.write(json.dumps(settings, cls=HammerJSONEncoder, indent=4))
+
+        options = HammerDriverOptions(
+            environment_configs=[],
+            project_configs=[proj_config],
+            log_file=os.path.join(tmpdir, "log.txt"),
+            obj_dir=tmpdir
+        )
+
+        self.assertTrue(HammerVLSISettings.set_hammer_vlsi_path_from_environment(),
+            "hammer_vlsi_path must exist")
+        driver = HammerDriver(options)
+
+        CLIDriver.generate_build_inputs(driver, lambda x: None)
+
+        d_file = os.path.join(driver.obj_dir, "hammer.d")
+        self.assertTrue(os.path.exists(d_file))
+
+        contents = open(d_file).readlines()
+
+        # TODO: Actually check the contents of hammer.d
+
+        # Cleanup
+        shutil.rmtree(tmpdir)
+
 
 
 if __name__ == '__main__':
--- a/src/hammer-vlsi/cli_driver_test.py
+++ b/src/hammer-vlsi/cli_driver_test.py
@@ -592,7 +592,7 @@ class HammerBuildSystemsTest(unittest.TestCase):
         tasks = {"pcb", "syn", "par", "drc", "lvs"}
         expected_targets = tasks.copy()
         expected_targets.update({"redo-" + x for x in tasks if x is not "pcb"})
-        expected_targets.update({os.path.join(tmpdir, x + "-rundir", x + "-output.json") for x in tasks})
+        expected_targets.update({os.path.join(tmpdir, x + "-rundir", x + "-output-full.json") for x in tasks})
         expected_targets.update({os.path.join(tmpdir, x + "-input.json") for x in tasks if x not in {"syn", "pcb"}})
 
         self.assertEqual(set(targets.keys()), set(expected_targets))
@@ -653,7 +653,7 @@ class HammerBuildSystemsTest(unittest.TestCase):
         targets = self._read_targets_from_makefile(contents)
 
         mods = {"TopMod", "SubModA", "SubModB"}
-        expected_targets = {"pcb", os.path.join(tmpdir, "pcb-rundir", "pcb-output.json")}
+        expected_targets = {"pcb", os.path.join(tmpdir, "pcb-rundir", "pcb-output-full.json")}
         expected_targets.update({"syn-" + x for x in mods})
         expected_targets.update({"par-" + x for x in mods})
         expected_targets.update({"lvs-" + x for x in mods})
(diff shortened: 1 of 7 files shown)


jwright6323:
Looking for a preliminary review before I invest any time in unit tests or documentation, but basically this renames the old `get_hierarchical_settings` method to `_hierarchical_helper` and also outputs the  dependency graph. This is partly out of laziness and partly out of DRY philosophy. There are now two methods to get the individual components of the tuple that `_hierarchical_helper` returns. 

There are two new CLI actions `build` and `build-inputs` (not sold on these names) that will make the build configuration for the tool pointed to by `vlsi.core.build_tool`. This currently only supports `make` and `none`. The `make` tool will just traverse the graph and spit out syn-par segments stitched together. There are some default make variables it creates, but they are overridable. You'd basically use this by creating a make target like so:

```
$(OBJ_DIR)/hammer.d:
    $(HAMMER_EXEC) -e env.yml -p config1.yml -p config2.yml -p config3.yml --obj_dir $(OBJ_DIR) build

include $(OBJ_DIR)/hammer.d
```

And, if you wanted, you could override `$(HAMMER_EXEC)` to be whatever you want (by default it is set to `sys.argv[0]` in the context of the `build` action, which is equal to whatever `$(HAMMER_EXEC)` is here). The tool stitches the make targets into your flow with the `$(HAMMER_DEPENDENCIES)` make variable, which is by default set to all input env and proj configs + all input verilog files. This is also overridable by setting it yourself.

src/hammer-vlsi/hammer_vlsi/cli_driver.py
```diff
@@ -856,6 +859,77 @@ def auto_action(driver: HammerDriver, append_error_func: Callable[[str], None])
 
         return driver, errors
 
+    @staticmethod
+    def make_build_inputs(driver: HammerDriver, append_error_func: Callable[[str], None]) -> Optional[dict]:
+        """
+        TODO
+        """
+        build_system_key = "vlsi.core.build_system"
```
edwardcwang: Just curious as to how come this is its own variable

src/hammer-vlsi/hammer_vlsi/cli_driver.py
```diff
@@ -856,6 +859,77 @@ def auto_action(driver: HammerDriver, append_error_func: Callable[[str], None])
...
+                        deps = " ".join(["par-" + x for x in out_edges])
+
+                    # TODO make this DRY
+                    obj_dir = os.path.realpath(driver.obj_dir)
+                    syn_run_dir = os.path.join(obj_dir, "syn-" + node)
+                    par_run_dir = os.path.join(obj_dir, "par-" + node)
+                    syn_timestamp = os.path.join(syn_run_dir, "timestamp")
+                    par_timestamp = os.path.join(par_run_dir, "timestamp")
+
+                    syn_out = os.path.join(syn_run_dir, "syn-output.json")
```
edwardcwang: Would be good if we can standardize these names somewhere (right now they're just scattered over the code manually)

src/hammer-vlsi/hammer_vlsi/cli_driver.py
```diff
@@ -856,6 +859,77 @@ def auto_action(driver: HammerDriver, append_error_func: Callable[[str], None])
...
+    def make_build_inputs(driver: HammerDriver, append_error_func: Callable[[str], None]) -> Optional[dict]:
+        """
+        TODO
+        """
+        build_system_key = "vlsi.core.build_system"
+        build_system = str(driver.database.get_setting(build_system_key, "none"))
+        dependency_graph = driver.get_hierarchical_dependency_graph()
+        if build_system == "none":
+            pass
+        elif build_system == "make":
```
edwardcwang: We might want to make these functions then index into it like `build_systems[system]()`

colinschmidt:
Can you add drc and lvs?

How does this interact with manual usage of hammer?
Suppose I use the makefile and something breaks in the middle, what are the settings/jsons/etc I need to pass to hammer to reproduce the step that failed?

jwright6323:
Sure I can add DRC and LVS. Right now this is merging the configs but we could change that pretty easily. The problem would then be if you changed one of configs, it would retrigger a rebuild of the whole flow. Someday we could have a smarter build system, but that day is a long way away. It's fairly easy to re-touch the timestamps, though.

src/hammer-vlsi/hammer_vlsi/cli_driver.py
jwright6323: Just easier to read- there are other methods that use this pattern, and I modeled after them. I can change it; it's not _that_ much shorter.

src/hammer-vlsi/hammer_vlsi/cli_driver.py
jwright6323: Agree. I tried calling `par_tool.run_dir` but this all happens before the driver sets up the syn and par tools, so those are `None`. Not sure the best way to do it. I'm inclined to leave it as a TODO for this PR, but if you can think of a clever, low-effort way to get it in I'm all ears.

src/hammer-vlsi/hammer_vlsi/cli_driver.py
jwright6323: IMO we should make that change whenever another build system is supported, which I imagine will be a while unless you guys are getting bazel set up.

src/hammer-vlsi/hammer_vlsi/cli_driver.py
edwardcwang: TODO is fine for now. Maybe file a `DeveloperSupport` issue?

src/hammer-vlsi/hammer_vlsi/cli_driver.py
edwardcwang: Are there strong interdependencies between the tools? My thought was that it would make future PRs to add other `build_system`s less invasive.

jwright6323: Maybe I'm misunderstanding what you're proposing. I was suggesting something like

```
elif build_system == "bazel":
    # Do stuff here
```

That's more or less the pattern we've used for other mode-type things before. I am interpreting what you want as:

```
build_systems[build_system](dependency_graph)
```

which doesn't seem any more or less invasive, but slightly less easy to read. If you want something different can you write a quick snippet?

edwardcwang: We might imagine something like

```python
def makefile(...): ...

def bazel(...): ...

def scons(...): ...

build_systems = {
  "makefile": makefile,
  "bazel": bazel,
  "scons": scons,
  ...
} # <-- this may be even programmatically modified
```

jwright6323: So the value proposition is that you can dynamically edit `build_systems` (and not paint yourself into a corner like we did with `submit_command`)?

edwardcwang: It seems so. I think it also helps since adding a new system now just involves

```python
def my_new_system(...): ...

build_systems = { 
  # ...
  "my_new_system": my_new_system
}
```

I think for modes it makes sense to just do `if/elif` since it's not often that we add new modes

jwright6323: Lemme give it a shot

jwright6323:
@edwardcwang have a look

src/hammer-vlsi/hammer_vlsi/cli_driver.py
jwright6323: I think this is resolved, but let me know if you have other feedback

src/hammer-vlsi/hammer_vlsi/cli_driver.py
jwright6323: OK, filed #450

jwright6323:
Also I believe this should close #234.

jwright6323:
@colinschmidt @edwardcwang I think this is ready for a real review. I'll push the button after my currently running job finishes, but I would consider this ready to go.

src/hammer-vlsi/cli_driver_test.py
```diff
@@ -532,5 +533,138 @@ def synthesis_action(self, bad: int) -> dict:  # type: ignore
             BadOverride()
 
 
+class HammerBuildSystemsTest(unittest.TestCase):
+
+    def _read_makefile_lines(self, lines: List[str]) -> Dict[str, List[str]]:
```
edwardcwang:
```suggestion
    def _read_targets_from_makefile(self, lines: List[str]) -> Dict[str, List[str]]:
        """
        Helper method to read information about targets from lines of a Makefile.
        """
```

src/hammer-vlsi/cli_driver_test.py
```diff
@@ -532,5 +533,138 @@ def synthesis_action(self, bad: int) -> dict:  # type: ignore
...
+        targets = {}  # type: Dict[str, List[str]]
+
+        for line in lines:
+            m = re.match(r"^([^.][^\s:]+)\s*:(.*)$", line)
```
edwardcwang: Can we add a comment here explaining what this regex is searching for?

src/hammer-vlsi/cli_driver_test.py
```diff
@@ -532,5 +533,138 @@ def synthesis_action(self, bad: int) -> dict:  # type: ignore
...
+            if m:
+                t = m.group(1)
+                p = re.split(r"\s+", m.group(2))
+                self.assertFalse(t in targets)
+                targets[t] = p
+
+        return targets
+
+    def test_flat_makefile(self) -> None:
```
edwardcwang:
```suggestion
    def test_flat_makefile(self) -> None:
        """
        Test that a Makefile for a flat design is generated correctly.
        """
```

src/hammer-vlsi/cli_driver_test.py
```diff
@@ -532,5 +533,138 @@ def synthesis_action(self, bad: int) -> dict:  # type: ignore
...
+
+    def _read_makefile_lines(self, lines: List[str]) -> Dict[str, List[str]]:
+        targets = {}  # type: Dict[str, List[str]]
+
+        for line in lines:
+            m = re.match(r"^([^.][^\s:]+)\s*:(.*)$", line)
+            if m:
+                t = m.group(1)
+                p = re.split(r"\s+", m.group(2))
+                self.assertFalse(t in targets)
```
edwardcwang:
```suggestion
                self.assertFalse(t in targets, "Found duplicate target {}".format(t))
```

jwright6323:
@colinschmidt do we need "hier_par_to_syn_action"s ?

src/hammer-vlsi/cli_driver_test.py
```diff
@@ -532,5 +533,138 @@ def synthesis_action(self, bad: int) -> dict:  # type: ignore
...
+        self.assertTrue(HammerVLSISettings.set_hammer_vlsi_path_from_environment(),
+            "hammer_vlsi_path must exist")
+        driver = HammerDriver(options)
+
+        CLIDriver.generate_build_inputs(driver, lambda x: None)
+
+        d_file = os.path.join(driver.obj_dir, "hammer.d")
+        self.assertTrue(os.path.exists(d_file))
+
+        contents = open(d_file).readlines()
```
edwardcwang: This leaks a fd
```suggestion
        with open(d_file, "r") as f:
            contents = f.readlines()
```

src/hammer-vlsi/cli_driver_test.py
```diff
@@ -532,5 +533,138 @@ def synthesis_action(self, bad: int) -> dict:  # type: ignore
...
+        expected_targets.update({os.path.join(tmpdir, x + "-input.json") for x in tasks if x not in {"syn", "pcb"}})
+
+        self.assertEqual(set(targets.keys()), set(expected_targets))
+
+        # TODO at some point we should add more tests
+
+        # Cleanup
+        shutil.rmtree(tmpdir)
+
+    def test_hier_makefile(self) -> None:
```
edwardcwang:
```suggestion
    def test_hier_makefile(self) -> None:
        """
        Test that a Makefile can be correctly generated for an hierarchical design.
        """
```

src/hammer-vlsi/hammer_vlsi/cli_driver.py
```diff
@@ -159,6 +160,8 @@ def __init__(self) -> None:
     def action_map(self) -> Dict[str, CLIActionType]:
         """Return the mapping of valid actions -> functions for each action of the command-line driver."""
         return add_dicts({
+            "build": self.generate_build_inputs,
+            "build-inputs": self.generate_build_inputs,
```
edwardcwang:
```suggestion
            "build-inputs": self.generate_build_inputs,
            "build_inputs": self.generate_build_inputs,
```

src/hammer-vlsi/hammer_vlsi/cli_driver.py
```diff
@@ -856,6 +859,21 @@ def auto_action(driver: HammerDriver, append_error_func: Callable[[str], None])
...
         return driver, errors
 
+    @staticmethod
+    def generate_build_inputs(driver: HammerDriver, append_error_func: Callable[[str], None]) -> Optional[dict]:
+        """
+        Generate the build tool artifacts for this flow, specified by the "vlsi.core.build_system" key.
+
+        :param driver: The HammerDriver object which has parsed the configs specified by -p
+        :param append_error_func: The function to use to append an error (unused).
+        :return: A build-system-specific dictionary
```
edwardcwang: Currently doesn't it return the dependency graph?

src/hammer-vlsi/hammer_vlsi/cli_driver.py
```diff
@@ -856,6 +859,21 @@ def auto_action(driver: HammerDriver, append_error_func: Callable[[str], None])
 
         return driver, errors
 
+    @staticmethod
+    def generate_build_inputs(driver: HammerDriver, append_error_func: Callable[[str], None]) -> Optional[dict]:
+        """
+        Generate the build tool artifacts for this flow, specified by the "vlsi.core.build_system" key.
+
+        :param driver: The HammerDriver object which has parsed the configs specified by -p
+        :param append_error_func: The function to use to append an error (unused).
```
edwardcwang: Shouldn't we use this to report errors in the build system methods?

src/hammer-vlsi/hammer_vlsi/hammer_build_systems.py
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+#  hammer_build_systems.py
+#  Class containing all the methods to create VLSI flow build system infrastructure
```
edwardcwang:
```suggestion
#  Class containing all the methods to create VLSI flow build system infrastructure.
```

src/hammer-vlsi/hammer_vlsi/driver.py
```diff
@@ -75,6 +75,9 @@ def __init__(self, options: HammerDriverOptions, extra_project_config: dict = {}
         # Store the run dir.
         self.obj_dir = options.obj_dir  # type: str
 
+        # Also store the options
+        self.options = options
```
edwardcwang: What's the context for this change?

src/hammer-vlsi/hammer_vlsi/driver.py
```diff
@@ -872,6 +883,17 @@ def get_hierarchical_settings(self) -> List[Tuple[str, dict]]:
...
+        return self._hierarchical_helper()[0]
+
+    def _hierarchical_helper(self) -> Tuple[List[Tuple[str, dict]], Dict[str, Tuple[List[str], List[str]]]]:
+        """
+        Read settings from the database, determine leaf/hierarchical modules, an order of execution, and return an
+        ordered list (from leaf to top) of modules and associated config snippets needed to run syn+par for that module
+        hierarchically and the dependency graph. Do not call this method directly- use get_hierarchical_settings or
+        get_hierarchial_dependency_graph instead.
+
+        :return: Tuple of 1. List of tuples of (module name, config snippet) and 2. The dependency graph
```
edwardcwang:
```suggestion
        :return: Tuple of (list of tuples of (module name, config snippet), dependency graph)
```

src/hammer-vlsi/hammer_vlsi/hammer_build_systems.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    dependency_graph = driver.get_hierarchical_dependency_graph()
+    return dependency_graph
+
+
+def build_makefile(driver: HammerDriver) -> dict:
+    """
+    Build a Makefile include in the obj_dir called hammer.d. This is intended to be dynamically
+    created and included into a top-level Makefile.
+
```
edwardcwang: Could you describe a bit about the Makefile that is generated at a high level? Looking at the unit tests I see for example `redo-` targets but couldn't find a high level description of what they are or what they do (don't see it in dependency graph either)

jwright6323: Sure, I agree it needs some more meat here.

src/hammer-vlsi/hammer_vlsi/hammer_build_systems.py
```diff
@@ -0,0 +1,172 @@
...
+    :param driver: The HammerDriver
+    :return: The dependency graph
+    """
+    dependency_graph = driver.get_hierarchical_dependency_graph()
+    makefile = os.path.join(driver.obj_dir, "hammer.d")
+    default_dependencies = driver.options.project_configs + driver.options.environment_configs
+    default_dependencies.extend(list(driver.database.get_setting("synthesis.inputs.input_files", [])))
```
colinschmidt: Is there some way to get add to this dynamically? I'm thinking of transcluded files for instance.

colinschmidt:
Yes you need hier-par-to-syn for the hierarchical deps

src/hammer-vlsi/hammer_vlsi/driver.py
jwright6323: fixed

src/hammer-vlsi/hammer_vlsi/driver.py
jwright6323: We need to know the actual paths of the input confs, so that they can be passed along to the build tool.

src/hammer-vlsi/hammer_vlsi/cli_driver.py
jwright6323: Sure, I passed it down. I originally hadn't used it before we split it out to a separate Dict and missed it.

src/hammer-vlsi/cli_driver_test.py
jwright6323: fixed

src/hammer-vlsi/cli_driver_test.py
jwright6323: fixed

src/hammer-vlsi/cli_driver_test.py
jwright6323: fixed

src/hammer-vlsi/cli_driver_test.py
jwright6323: Yep, it's always hard to understand a regex even 5 minutes after writing it...

src/hammer-vlsi/cli_driver_test.py
jwright6323: fixed

jwright6323:
Slapping `DO NOT MERGE` back on because I need to do this par-to-syn stuff. I'll also fix the par-outputs.json bug we discussed offline.

src/hammer-vlsi/cli_driver_test.py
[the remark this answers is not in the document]
jwright6323: Oops. No, it's broken.

src/hammer-vlsi/cli_driver_test.py
```diff
@@ -532,5 +533,138 @@ def synthesis_action(self, bad: int) -> dict:  # type: ignore
...
+        tmpdir = tempfile.mkdtemp()
+        proj_config = os.path.join(tmpdir, "config.json")
+
+        settings = {
+                "vlsi.core.technology": "nop",
+                "vlsi.core.build_system": "make",
+                "vlsi.inputs.hierarchical.mode": "hierarchical",
+                "vlsi.inputs.hierarchical.top_module": "TopMod",
+                "vlsi.inputs.hierarchical.config_source": "manual",
+                "vlsi.inputs.hierarchical.manual_modules": [{"TopMod": ["SubModA", "SubModB"]}],
```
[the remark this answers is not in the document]
jwright6323: What do you mean?

src/hammer-vlsi/hammer_vlsi/hammer_build_systems.py
jwright6323: Yikes, probably, but that sounds really hard.

src/hammer-vlsi/cli_driver_test.py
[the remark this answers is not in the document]
jwright6323: easy fix, though

src/hammer-vlsi/hammer_vlsi/cli_driver.py
```diff
@@ -856,6 +859,21 @@ def auto_action(driver: HammerDriver, append_error_func: Callable[[str], None])
 
         return driver, errors
 
+    @staticmethod
+    def generate_build_inputs(driver: HammerDriver, append_error_func: Callable[[str], None]) -> Optional[dict]:
+        """
+        Generate the build tool artifacts for this flow, specified by the "vlsi.core.build_system" key.
```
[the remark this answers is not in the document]
jwright6323: Filed #454

src/hammer-vlsi/cli_driver_test.py
```diff
@@ -532,5 +533,138 @@ def synthesis_action(self, bad: int) -> dict:  # type: ignore
...
+
+        tasks = {"pcb", "syn", "par", "drc", "lvs"}
+        expected_targets = tasks.copy()
+        expected_targets.update({"redo-" + x for x in tasks if x is not "pcb"})
+        expected_targets.update({os.path.join(tmpdir, x + "-rundir", x + "-output.json") for x in tasks})
+        expected_targets.update({os.path.join(tmpdir, x + "-input.json") for x in tasks if x not in {"syn", "pcb"}})
+
+        self.assertEqual(set(targets.keys()), set(expected_targets))
+
+        # TODO at some point we should add more tests
```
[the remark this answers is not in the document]
jwright6323: Filed #454

src/hammer-vlsi/hammer_vlsi/cli_driver.py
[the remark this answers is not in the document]
jwright6323: Yep, done

src/hammer-vlsi/hammer_vlsi/driver.py
```diff
@@ -864,6 +867,14 @@ def run_pcb(self, hook_actions: Optional[List[HammerToolHookAction]] = None, for
 
         return run_succeeded, output_config
 
+    def get_hierarchical_dependency_graph(self) -> Dict[str, Tuple[List[str], List[str]]]:
+        """
+        Return the dependency graph for this build if it is hierarchical, or an empty dict if it is flat.
```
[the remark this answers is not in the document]
jwright6323: Went with flow

src/hammer-vlsi/hammer_vlsi/hammer_build_systems.py
jwright6323: Filed #455

src/hammer-vlsi/cli_driver_test.py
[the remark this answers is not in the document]
colinschmidt: An issue on Hammer that is something like:
"Add unit tests for hierarchical designs with more than 2 levels of modules"

src/hammer-vlsi/hammer_vlsi/hammer_build_systems.py
jwright6323: Done

src/hammer-vlsi/cli_driver_test.py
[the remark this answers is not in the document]
jwright6323: #456

jwright6323:
I think I fixed all extant issues. I'm going to try this on a real design and see what happens. Can you guys glance over this once more while that runs? I do not anticipate any future changes, but who knows.

src/hammer-vlsi/cli_driver_test.py
```diff
@@ -532,5 +533,153 @@ def synthesis_action(self, bad: int) -> dict:  # type: ignore
             BadOverride()
 
 
+class HammerBuildSystemsTest(unittest.TestCase):
+
+    def _read_targets_from_makefile(self, lines: List[str]) -> Dict[str, List[str]]:
+        """
+        Helper method to read information about targets from lines of a Makefile
```
edwardcwang: nit
```suggestion
        Helper method to read information about targets from lines of a Makefile.
```

src/hammer-vlsi/cli_driver_test.py
```diff
@@ -532,5 +533,153 @@ def synthesis_action(self, bad: int) -> dict:  # type: ignore
...
+        """
+        targets = {}  # type: Dict[str, List[str]]
+
+        for line in lines:
+            # This regex is looking for all non-special targets (i.e. those that aren't .PHONY, .INTERMEDIATE, .SECONDARY, ...)
```
edwardcwang: +1

src/hammer-vlsi/defaults.yml
```diff
@@ -15,6 +15,11 @@ vlsi.core:
   # TODO: move this to vlsi.technology (or technology.core) ucb-bar/hammer#317
   node: 32
 
+  # Build system to use, if any.
+  # This is used to auto-generate a build file, if desired. It is not required to run hammer.
+  # Valid options are "none" or "make"
+  build_system: "make"
```
edwardcwang: Since it's not required maybe we could leave the default as `none`?

edwardcwang:
(some of them might be collapsed by GitHub since they are in the middle)

src/hammer-vlsi/defaults.yml
jwright6323: I can do that.

src/hammer-vlsi/hammer_vlsi/cli_driver.py
jwright6323: Yes, but it doesn't have to. I'll go ahead and change it- I doubt we'll ever return anything esle.

jwright6323:
Please ask me before committing to my branch; some of these commits need to be reverted because I already fixed them and they'd be easy to miss. I missed a few of your feedback comments because they didn't show up on the front page for some reason, but I just addressed them.

Also, I know we had an offline discussion about this but I'd like to reiterate that a period at the end of a sentence fragment is syntactically incorrect English. If you want period in the pydoc, we should make them full sentences. It doesn't make sense to have a period after
```
Helper method to read information about targets from lines of a Makefile
```

edwardcwang:
I see, sorry about that. Was trying to save some time and I'm sorry that I caused you extra work/bother in the process.

Re. docstrings/etc: my impression is that in software it's standard to put a period in descriptions even when it's not a complete sentence e.g. https://docs.python.org/3/library/typing.html#typing.Deque

jwright6323:
I see. Well, then let's put the periods in docstrings.

Also, can you re-review the last few commits I made? It's feature creeping a bit, but I noticed the DRC/LVS stuff was missing outputs.json.

src/hammer-vlsi/hammer_vlsi/hammer_build_systems.py
```diff
@@ -0,0 +1,244 @@
...
+    of the hierarchical modules (e.g. syn-Top, syn-SubModA, par-SubModA, etc.). The appropriate
+    dependencies and bridge actions are automatically generated from the hierarchy provided in the
+    Hammer IR.
+
+    Additionally, "redo" steps are created (e.g. redo-par for flat designs or redo-par-Top for
+    hierarchical), which allow the user to bypass the normal Makefile dependencies and force a
+    rerun of a particular task. This is useful when the user wants to change an input Hammer IR file
+    knowing it will not affect intermediate steps in the design.
+
+    An example use case for integrating this file into a top flow is provided below. Be sure to use
```
edwardcwang: Question, how does the snippet below include `$(OBJ_DIR)/hammer.d`?

src/hammer-vlsi/hammer_vlsi/cli_driver.py
```diff
@@ -386,13 +386,25 @@ def action(driver: HammerDriver, append_error_func: Callable[[str], None]) -> Op
                 else:
                     post_load_func_checked(driver)
                 success, output = driver.run_drc(extra_hooks)
+                if not success:
```
edwardcwang: Long-term we should probably DRY this stuff

edwardcwang: Also SRAM generator is missing these checks but it's probably not that important.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -706,6 +706,11 @@ def signoff_results(self) -> int:
 
 class HammerDRCTool(HammerSignoffTool):
 
+    def export_config_outputs(self) -> Dict[str, Any]:
```
edwardcwang: These are fine though at the moment we don't really use things like `drc-output.json` unless I missed a memo...?

src/hammer-vlsi/hammer_vlsi/hammer_build_systems.py
jwright6323: lol, it doesn't. I spent too much effort trying to make the example clean and forgot the most important line.

jwright6323: Fixed.

src/hammer-vlsi/hammer_vlsi/cli_driver.py
jwright6323: Filed #457

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
jwright6323: They aren't, but on the off chance they are consumed by something it'll be less confusing to have things be consistent. Also, it makes the Makefile a lot more uniform looking (the output jsons are the targets in the Makefile)

jwright6323:
Thanks, I'll hit the button when this job finishes.

jwright6323:
OK, looks like it worked. Made a small change to use ONLY the syn-*-input.yml for non-leafs, but it is otherwise as it was last night.

PR #449, opened 2019-06-18
Add absolute track_offset

harrisonliew:
This is needed to add an absolute offset to the power strap start position relative to the design's bounding box.

This is very useful to align power straps between blocks in hierarchical P&R.


src/hammer-vlsi/test.py
```diff
@@ -1674,6 +1678,9 @@ def test_simple_by_tracks_power_straps(self) -> None:
                     self.assertEqual(entry["nets"], nets)
                     # Check that the requested tracks equals the used tracks
                     requested_tracks = track_width * 2 + track_spacing
+                    print(metal.pitch)
```
edwardcwang: leftover print statements?

PR #463, opened 2019-06-23
Canonicalize the input file paths

jwright6323:
Calling hammer with `-p ../../foo.yml` does not give the correct behavior when using the `prependlocal` meta or `local` deepsubst metas.


jwright6323:
Buildbot seems to be down. This passes tests and type checking.

PR #467, opened 2019-06-24
add asap7 support

sequencer:


sequencer:
recreate https://github.com/ucb-bar/hammer/pull/443, and be able to synthesis with genus.
https://github.com/ucb-bar/hammer/pull/445 is also needed, because original tar is nested.

sequencer:
> I don't think ASAP7 should be the default. In fact, I think the current default is also wrong- it should be `nop`, IMO.

Done.

sequencer:
any updates?

edwardcwang:
Hey @sequencer - to follow up, the plan for this PR specifically is to review/test/merge #445 and then we will review/test this PR and get it in.

sequencer:
OK, thx!!

PR #374, opened 2019-04-26
Test PR; do not merge

edwardcwang:

PR #480, opened 2019-07-08
Bring primary CI back up

edwardcwang:


._travis.yml
```diff
@@ -10,6 +10,7 @@ script:
   - source sourceme.sh
   - cd src/test
   - pip install mypy
-  - ./mypy_with_exclusions.sh
+# Travis' mypy install causes some errors
```
colinschmidt: For the record I believe I see these errors as well thats why I made the other changes. So its some version issue that would appear for others.

edwardcwang:
Also tested with #374 and it seems to be responding now.

._travis.yml
edwardcwang: Ah good point, added the version number for mypy.

edwardcwang:
Will plan to this tomorrow unless other major issues crop up

edwardcwang:
Thanks!

colinschmidt:
I think we should leave travis checks on until the buildbot code is available.

PR #481, opened 2019-07-24
Add basic support for clock groups

### LINKED ISSUE #59: Support clock groups in SDC
Specify which clocks are asynchronous, etc to each other.

colinschmidt:
Fixes #59

### DIFF
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -215,6 +215,7 @@ vlsi.inputs:
   # source_path (str) - Required if generated. The path of the clock that this clock is generated from.
   # divisor (int) - Required if generated. The amount this generated clock is slowed from the source clock e.g. 2 => this clock will be two times slower than the source clock.
   #   We are constrained to integers by SDC.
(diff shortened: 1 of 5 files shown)


src/hammer-vlsi/defaults.yml
```diff
@@ -215,6 +215,7 @@ vlsi.inputs:
   # source_path (str) - Required if generated. The path of the clock that this clock is generated from.
   # divisor (int) - Required if generated. The amount this generated clock is slowed from the source clock e.g. 2 => this clock will be two times slower than the source clock.
   #   We are constrained to integers by SDC.
+  # group (str) - Optional. The name of the clock group this clock belongs to. Clocks in the same group will not be marked as asynchronous.
```
jwright6323: if this is optional and not of type `Optional[str]` can we document what the default value is?

src/hammer-vlsi/defaults.yml
```diff
@@ -215,6 +215,8 @@ vlsi.inputs:
...
+  #   Clocks without a group specified will all be placed in separate groups and thus marked as asynchronous to each other.
```
```suggestion
  #   Clocks with no group specified will all be placed in separate groups and thus marked as asynchronous to each other and all other groups.
```

PR #493, opened 2019-08-16
Fix clock groups

jerryz123:
- No need to `set_clock_groups` if there are none


src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -1047,10 +1047,11 @@ def sdc_clock_constraints(self) -> str:
                     groups[clock.group] = [clock.name]
             else:
                 ungrouped_clocks.append(clock.name)
-        output.append("set_clock_groups -asynchronous {grouped} {ungrouped}".format(
-            grouped = " ".join(["{{ {c} }}".format(c=" ".join(clks)) for clks in groups.values()]),
-            ungrouped = " ".join(["{{ {c} }}".format(c=clk) for clk in ungrouped_clocks])
-        ))
+        if len(groups):
```
edwardcwang: This could be a separate PR

PR #495, opened 2019-08-22
Allow max_width option for layers

jerryz123:
- `max_width` is an optional field which specifies the maximum wire width on that layer
- The power straps algorithms will not generate a wire width > `max_width`

### DIFF
--- a/src/hammer-tech/stackup.py
+++ b/src/hammer-tech/stackup.py
@@ -102,6 +102,7 @@ class Metal(NamedTuple('Metal', [
         ('index', int),
         ('direction', RoutingDirection),
         ('min_width', Decimal),
+        ('max_width', Decimal),
         ('pitch', Decimal),
         ('offset', Decimal),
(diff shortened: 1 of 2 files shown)


edwardcwang:
Would a hypothetical stackup dumping script need to be updated for this PR?

jerryz123:
If you are talking about the one in cad-plugins, I can look into updating that.

src/hammer-tech/stackup.py
```diff
@@ -194,6 +197,8 @@ def min_spacing_and_max_width_from_pitch(self, pitch: Decimal) -> Tuple[Decimal,
                 spacing = pitch - width
 
         width = pitch - spacing
+        if self.max_width > 0.0 and width > self.max_width:
```
jwright6323: I think some python versions complain about comparing floats (`0.0`) and decimals, but if this passes CI I'm cool with it.

Issue #503, opened 2019-09-03
Remove references to SAED32


jwright6323:
We should scrub all the references to SAED32- we don't necessarily need to do a history rewrite, but ASAP7 should be what we reference going forward.

edwardcwang:
Yep, once ASAP7 is up and stable this should be pretty straightforward. Props to @sequencer for driving this.

jwright6323:
Closed by #504

PR #504, opened 2019-09-04
remove saed32

sequencer:
based on the #503,
after lvs/drc for asap7 success, this could be merged.


jwright6323:
LGTM, @harrisonliew can you review?

harrisonliew:
LGTM too, thanks @sequencer!

farzadfch:
I am new to hammer and I was considering using saed32. I wonder why saed32 is removed. Is there something particularly wrong about saed32 or it's just the matter of giving up on supporting it in hammer?

jwright6323:
SAED32 is proprietary intellectual property of Synopsys, and we can't support it publicly, so we are opting to use ASAP7 instead, as it has a less restrictive EULA.

PR #510, opened 2019-09-07
Special cells

dpgrubb13:
API for special physical cells built on top of work from @bkoppelmann.  Looking for feedback on the handling of the "special fields" input in the tech plugin and how the key not found cases are handled when the requested kind of special cell is not defined in the tech plugin.

This also fixes the verilog sim files in the asap7 tech plugin such that HammerSim will complete successfully.


harrisonliew:
Two questions:

Are we supposed to remove all the FF/SS verilog files from all corners, or supposed to replace them with the TT version?

I think all of these types of cells can be automatically regarded as physical-only cells, which need to be excluded from the netlist that is written out after P&R. Can we have the physical-only cells tap into this struct instead of looking for a `physical only cells list` in the tech JSON?

dpgrubb13:
1. They should be removed, otherwise it will end up including duplicated files in the sim, and the single verilog file will cover for all corners.

2. Will do

3. That could be a good suggestion, I'll look into it.

PR #500, opened 2019-08-31
asap7 (take 2)

sequencer:

PR #511, opened 2019-09-07
ASAP7 (mostly DRC & LVS)

harrisonliew:
Contains commits from #447 (to be rebased) & shall likely supercede #500. May also change depending on #510.

- Contains a method to hack the ASAP7 PDK. Needed to resolve duplicate files & missing GDS's.
- Included a copy of the python-gdsii package (https://pypi.org/project/python-gdsii/). This is needed to copy the BOUNDARY layer as the threshold marking layers.
- Updated the tech plugin with rule decks & layer maps.

Thank you @sequencer!


sequencer:
I have a question that should we include `python-gdsii` by `git submodule` or with `pip`?
It is based on GPL, I'm afraid this will introduce some GPL problems for Hammer.
Can you verify this PR in your environment? It contains the `pip` and `git submodule` package management for hammer, comparing to vend code directly, which can control the package more easily. 
It failed in CI but work for the Arch Linux and CentOS 7.

sequencer:
btw, `python-gdsii` seems stall now(original upstream is already dead), I'm considering port `gds_vts.py` to https://github.com/heitzmann/gdspy, which is in a active development.

sequencer:
I noticed that `ASAP7_PDKandLIB.tar/ASAP7_PDKandLIB_v1p5/asap7libs_24.tar.bz2/asap7libs_24/cdl/pex` contains more similar problems that multi-VT is not matched to `SUBCKT`.
for example in `asap7sc7p5t_24_AO_SLVT.sp`, `AOI331xp33_ASAP7_75t_SL` is using `rvt` cell.

harrisonliew:
Given that the GDSII specification isn't changing and we are not editing them interactively, I found gdspy and related projects to be overly bloated for this purpose. As for including the package, I have no preference. I do appreciate your new PR though!

The PEX CDLs shouldn't affect LVS, but that is a bug that we might as well fix in the scripts.

sequencer:
> Given that the GDSII specification isn't changing and we are not editing them interactively, I found gdspy and related projects to be overly bloated for this purpose. 

Sure let's keep `gdspy` as you expected, if we need intricate gds manipulation functions further, we may change this later. 

> As for including the package, I have no preference but I have not tried #426 yet. I do appreciate your new PR though!

I'm working on some paperwork with cadence agency in School, after this, I can contribute a reproducible and stable EDA tool environment maintaining plugin for Hammer, which I have applied it for ICC in Wuhan.(the pip package manager is the the basement for this)

> The PEX CDLs shouldn't affect LVS, but that is a bug that we might as well fix in the scripts.

OK, I'll look further into these bugs and try to fix them.

edwardcwang:
@harrisonliew Could we check in gdsii as a submodule? This would help avoid the licence contamination issue.

sequencer:
How do you solve the LVS/DRC problem:
```
```

harrisonliew:
ucb-bar/hammer-cadence-plugins#30 does it (after re-writing the GDS), and you need to make sure you keep the `LAYOUT PRECISION 4000` parameter in the rules file.

harrisonliew:
It turns out that ASAP7 PNR operates on the 4x upscaled versions of the standard cells, but the final merged GDS pulls in the unscaled standard cell GDSs. This means the standard cells appear 4x too small at the correct origin location. I contacted Vinay and he showed me a Virtuoso method of scaling the GDS upon streaming in, but this is incompatible with our flow.

It appears we need to insert a `post_pnr_script` step of some sort in the PNR steps and actually switch over to `gdspy` because it appears to have ways of scaling geometries. We sould apply this script onto the post-PNR GDS.

sequencer:
How about scale precision 4000 -> 1000 in the `post_install_script`? Make PaR LVS DRC working in the precision.

sequencer:
And as much as I know, the Multi-VT gds file shouldn't be same since the transistor are supposed to be sputtered differently by marking them with special gds layer.
So I'm not sure if our approach(load, sed names, save) is a right way?

harrisonliew:
Changing the layout precision (in Innovus and DRC/LVS rule decks) does not perform any scaling, and may cause rounding errors when we do scaling after the GDS is created from Innovus. So we should not change the precision but instead scale down the top cell after the GDS is exported using `gdspy` (I can't get it to build yet...) while leaving the standard cell geometries alone, which is what the ASU guys did.

The script that I made to run with the `python-gdsii` copies the `BOUNDARY` layer as the needed VT layers for `LVT`, `SLVT`, and `SRAM`. This is what Vinay told me to do, so I think this is the correct way. LVS (in my test cases) does not have trouble recognizing the transistor VTs after these hacks.

sequencer:
I saw the `layermap` in the script, that's the layer mark for mosfet, I believe this is the right way.
I'm currently working on writing a `gdspy` script to do the same thing as `python-gdsii` did.

harrisonliew:
Ok, I look forward to seeing how you got `gdspy` to work correctly. I assume this will be rolled into some of your other PRs.

sequencer:
Emmm, `gdspy` will include `numpy`, this might be a `disaster` to old machines :(
I suggest you can have a try on `pip`, since there are a lot of C binding and FORTRAN binding to it, and it will be a nightmare to compile by hand.
I'm not sure if `pip install numpy` will include C binding or not, If that cannot work, You may need use `yum` to install the `numpy` with the `epel` repo for CentOS.

sequencer:
My EDA environment is based on Arch Linux, so I prefer use `pacman` to manage the python library correctly. However this might be impossible to porting this method to Company and Big School's machine farm. That the reason I suggest you can use pip to get it work.
After the Cadence guy finish the paper work, I can try to contribute a container-based env building script to easily build back-competitive environment to avoid such problems.

sequencer:
If you wanna use `pip` method, you can try #426. This should be work well in CentOS7(I used to test it on the pure CentOS7 machine).

sequencer:
I do believe although original gds name is `SLVT`, but it contains no `SLVT` layer, and it represents `RVT` since `SL` layer was added by that python script LOL. :)

harrisonliew:
I have been struggling with installing `gdspy` using #426. There's an ascii vs. utf-8 encoding problem when I SSH into our lab machines vs. running a terminal in an X desktop session. And on our lab machines it seems like it cannot find a `Python.h` file. Doing installs for packages is definitely messy and will break compatibility across environments...

Yes, you are correct, the original GDS was named for `SLVT` but contained only `RVT` cells (confirmed by Vinay).

sequencer:
> There's an ascii vs. utf-8 encoding problem when I SSH into our lab machines vs. running a terminal in an X desktop session.

I guess you are using macOS xquartz, you can get what locale in your lab machine by `localectl status`, and check your ssh by `locale`, you may need to get these locale same to solve the `utf-8` issues.

> And on our lab machines it seems like it cannot find a Python.h file

If your machine complain about the `Python.h`, it seems your environment didn't install Python with the package manager, so that `Python.h` is not installed correctly like `/usr/include/python3.7m/Python.h`.

I suggest that you can contact your system admin to install python 3.6 in the epel repo for you, otherwise, this will exactly be a nightmare.

If you wanna work around. This is my idea.
You need meet the requirement of
```
https://docs.scipy.org/doc/numpy/user/building.html
```
Python and C compiler is of cause on your machine, you may need compile Fortran 77 compiler by yourself(hopefully gcc can bootstrap that), and you may need compile [cython](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html), and then hack the path by `CFLAG=-I/generated/library/path` to compile numpy.
then hack the `PYTHONPATH` to exposed  `numpy` to your python.

So I still suggest install `numpy` by `yum`.

Another work around is use `docker` or `systemd-nspawn` to build your container, binding mount the EDA tools folder to the container, and you are the admin to the container that you can install any package.

sequencer:
or if you already have `pip`, try `pip install xxx --user`

sequencer:
> Changing the layout precision (in Innovus and DRC/LVS rule decks) does not perform any scaling, and may cause rounding errors when we do scaling after the GDS is created from Innovus.

> It turns out that ASAP7 PNR operates on the 4x upscaled versions of the standard cells, but the final merged GDS pulls in the unscaled standard cell GDSs

So If we change the precision of input gds file? downscale the standard cell before feeding to innovus?

sequencer:
Emmm, I just use `check_drc` in the innovus, which throws too much errors.
I'm not sure scaling down the innovus generated gds is good, since original library including `gds`, `lef` and `DRC/LVS decks` are still in 4x, every place use these informations might complain for the shrink `gds`.
If we continue to use 4x library, we may need do everything in 28nm, including the `DRC/LVS`, and other modifications to `gds`. After finish all executions, It is the time to provide a `gds` shrinking script for the final gds stream out.
Even in this method, I still have question in the RC extraction flow for post PaR simulation that how to scale down the extracted RC value? and for the essence of which, how did the timming lib changed during the scaling down and up?
Maybe the final solution is use everything in 7nm to eliminate the influence of scaling. I'll consult Vinay for further information for the scaling problem.

harrisonliew:
No, I have confirmed that the standard cell GDS and DRC/LVS rule decks are encoded in 1x and Vinay did scale down the top-level GDS prior to DRC/LVS. Only the LEFs are in 4x to get around Cadence licensing.

We definitely have not addressed the RC scaling. Let us know what Vinay says. We can do it similarly to #514 for now.

I have minor issues to fix for DRC/LVS now, will keep pushing as I go.

jwright6323:
The easiest thing to do is to give innovus a techfile with post-scaled parasitics so that if innovus draws something 4L, it would extract RC for something of length L. I would have assumed the ASAP7 kit already did that.

harrisonliew:
It's looking like the tech LEF and the DRM don't fully match up. For example, M4, M6, and M7 routing is flagged as off-grid (seems to be wrong offset) and M8/9 spacing table is not gradated as in the DRM (fixed in stackup object of tech JSON). I presume the DRM should be the authoritative reference?

sequencer:
I make the default node to 28 since the scale problem, which may need your review in origin/asap7_node_to_28

sequencer:
Have a try on this Module for PaR,
```
module RippleCarrySum2(
  input        clock,
  input        reset,
  input  [1:0] io_a,
  input  [1:0] io_b,
  output [2:0] io_z
);
  wire  as_0;
  wire  as_1;
  wire  bs_0;
  wire  bs_1;
  wire  ps_0;
  wire  cs_1;
  wire  ps_1;
  wire  _T;
  wire  _T_2;
  wire  cs_2;
  wire  sum_1;
  wire [1:0] _T_4;
  assign as_0 = io_a[0];
  assign as_1 = io_a[1];
  assign bs_0 = io_b[0];
  assign bs_1 = io_b[1];
  assign ps_0 = as_0 ^ bs_0;
  assign cs_1 = as_0 & bs_0;
  assign ps_1 = as_1 ^ bs_1;
  assign _T = as_1 & bs_1;
  assign _T_2 = cs_1 & ps_1;
  assign cs_2 = _T_2 | _T;
  assign sum_1 = ps_1 ^ cs_1;
  assign _T_4 = {cs_2,sum_1};
  assign io_z = {_T_4,ps_0};
endmodule
```
I got a strange [gds](https://pastebin.com/ZUE765gv)(encoded with base64), which seems to be a issue from the `gds` shrink script, seems like `M1` of `gds` didn't fullly get shrinked.

sequencer:
used both 1.4 and git master version of gdspy, got the same result.

harrisonliew:
I just realized that when I had merged in #515, I forgot to clear my tech cache. I discovered a bug in the `gdspy` version of the multi-Vt GDS (the case for RVT was not handled, so its GDS is empty). `gdspy` errored out, producing a GDS that was only 68 bytes long.

Can you try my most recent commit after wiping your tech cache? I was able to get a proper GDS with everything properly shrunk now.

sequencer:
I'm sorry for the bug I introduced, but I got a same issue here.

sequencer:
PaR works, generated gds is shrunken now!

sequencer:
Emmm, I fixed the strange bug in my side, please checkout origin/gds_scale_bug_fix for review. We seems to encounter the problem that behavior of the our gds stream out progroms are different.
bug mentioned in heitzmann/gdspy#100 is fixed, let's remove our hotfix when gdspy 1.5 published

harrisonliew:
I tried it out, and found you cannot leave the ref magnification at 0.25 otherwise the standard cells are all scaled down by 4x. But, the label should be so that the label text isn't 4x too large. Let me know if this does not match with the behavior with your tool.

sequencer:
Sorry! l150 is a bug I introduced by copy&paste! Is this script work now in your side?

Actually I didn’t meet the gdspy bug you need to work around in the hot fix.
On my side if I don’t set the MAG tag, M1 of stdcell won’t shrink down and will stick on each other.
I’m surprised why can you successfully scale down without setting MAG tag. Maybe our gds cell hierarchy are slightly different.(yours M1 is `CustomPlus`, mine exist in ref that need to be translated with set MAG tag)
Can you generate the gds of `RippleCarrySum2` and send me a copy without execute the shrink script. And let me compare what on earth the difference is?

> the standard cells are all scaled down by 4x.

That might be my misunderstanding. I think I should scale down the stdcell 4x otherwise stdcell and PaR result won’t match up.

I’ll look into this around 3 in UTC(this reply is just wake up from dreams ;P )

harrisonliew:
[RippleCarrySum2.gds.txt](https://github.com/ucb-bar/hammer/files/3633117/RippleCarrySum2.gds.txt)

I par'd your `RippleCarrySum2` with what's in this branch up to 80521da and I see no issues (remove the .txt extension). Are you using Innovus and merging the standard cell GDSs inside, or some other tool? For me, the MAG tag seems to make no difference, except in the resulting size of the labels/cell references. The standard cell GDSs as-is are already the scaled down (i.e. true 7nm) size. I'm still confused how your M1s are different.

sequencer:
gds/lef of asap7 are not corresponded, innovus PaR is guided by lef,
if not merged, gds illustrated in the 28nm(not scaled) matches well, maybe use precision information from lef for illustrating(without MAG tag);
if merged, gds will be merged with the gds precision information, thus `post_par_script` only needed to scale down the PaR result. with the filter here:
https://github.com/ucb-bar/hammer/blob/3a822100275b73c12643d5602ebf42a0949d9529/src/hammer-vlsi/technology/asap7/__init__.py#L130
So default `par.inputs.gds_merge` should be set to `true`, and we may need to expose `gds_merge` config to `post_par_script`?

src/hammer-vlsi/technology/asap7/defaults.yml
```diff
@@ -79,3 +79,11 @@ mentor.extra_env_vars:
   - VAR_M5_XOFFSET: "0"
   - VAR_M6_YOFFSET: "0"
   - VAR_M7_XOFFSET: "0"
+
+par.inputs:
+  # gds_merge: ture is mandatory,
+  # since asap7/__init__.py:130, the shrink script will filter stdcell module with keyword "ASAP7_75t"
+  # thus we should merge gds by PaR tools firstly,
+  # TODO: try to change export_config_outputs for PaR inputs to expose gds_merge to output json?
```
edwardcwang: I think we do want to try and pass it through since the reason why this is an input is because some projects may want to use a 3rd-party tool to merge the GDS

edwardcwang: This may suffice for now if we file an issue later

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+        vendor only provide SLVT gds, this patch will generate other 3(LVT, RVT, SRAM) VT gds file.
+        """
+        self.logger.info("generate gds for Multi-VT cells")
+
+        orig_gds = os.path.join(self.extracted_tarballs_dir, "ASAP7_PDKandLIB.tar/ASAP7_PDKandLIB_v1p5/asap7libs_24.tar.bz2/asap7libs_24/gds/asap7sc7p5t_24.gds")
+        # load original_gds
+        asap7_original_gds = gdspy.GdsLibrary(infile=orig_gds)
+        original_cells = asap7_original_gds.cell_dict
+        # WTF is this?
```
colinschmidt: I don't understand if this comment is about what that dictionary entry is or about why its being deleted.

sequencer:
> We definitely have not addressed the RC scaling. Let us know what Vinay says

Vinay says:
  Since all the LEFs are magnified by 4x for PnR, you will need to shrink the post-PnR GDS when you import it into Virtuoso.

I think here is his flow:
Virtuoso design stdcell ->
RC extract ->
4x stream out gds ->
innvous PaR

So as you are right, we should directly extract RC from shrunk gds generated by `post_par_script`.

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+                "mvt_layer": 97
+                },
+            "SRAM": {
+                "lib": gdspy.GdsLibrary(),
+                "mvt_layer": 110
+                },
+        }
+        # create new libs
+        for vt, multi_lib in multi_libs.items():
+            multi_lib["lib"].name = asap7_original_gds.name.replace('SL', vt)
```
colinschmidt: Nit: This seems to be the only play you use double quotes to index this dict.

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+                    mvt_polygon = gdspy.PolygonSet(boundary_polygon, multi_lib['mvt_layer'], 0)
+                    mvt_cell = cell.copy(name=cell.name.replace('SL', vt), exclude_from_current=True, deep_copy=True).add(mvt_polygon)
+                else:
+                    # RVT, just copy the cell
+                    mvt_cell = cell.copy(name=cell.name.replace('SL', vt), exclude_from_current=True, deep_copy=True)
+                # add mvt_cell to corresponding multi_lib
+                multi_lib['lib'].add(mvt_cell)
+
+        for vt, multi_lib in multi_libs.items():
+            # write multi_lib with 2.5e-10 precision
```
colinschmidt: I think you can combine the name setting into for loop into this one.

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+            multi_lib['lib'].precision = 2.5e-10
+            multi_lib['lib'].write_gds(os.path.splitext(orig_gds)[0] + '_' + vt + '.gds')
+
+    def fix_sram_cdl_bug(self) -> None:
+        """
+        vendor's SRAM cdl use slvt cell, this patch will sed cells name in which, fix this bug.
```
colinschmidt: I don't understand this comment.

colinschmidt: "this patch will replace 'slvt' in SRAM cell names with 'sram' to indicate the correct VT for the SRAM cells"

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+        Called by the `run_par` method
+        """
+        self.logger.info("Scaling down place & routed GDS")
+
+        # load original_gds
+        par_gds_file = d.get("par.outputs.output_gds")
+        gds_lib = gdspy.GdsLibrary(infile=par_gds_file)
+        # Iterate through cells that aren't part of standard cell library and scale
+        for k,v in gds_lib.cell_dict.items():
+            if not 'ASAP7_75t' in k:
```
colinschmidt: Is `startswith` still a correct filter? I'm worried people might develop other cells that are meant to work with the  `ASAP7_75t` cells and therefore have that string in their cell names. If we can narrow the filter that would help a bit.

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+                for poly in v.polygons:
+                    poly.scale(0.25)
+                for path in v.paths:
+                    path.scale(0.25)
+                    # gdspy bug: we also need to scale custom path extensions
+                    # wait gdspy/pull#101
```
colinschmidt: "Will be fixed by" instead of "wait"

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+                    for i, end in enumerate(path.ends):
+                        if isinstance(end, tuple):
+                            path.ends[i] = tuple([e*0.25 for e in end])
+                for label in v.labels:
+                    # bug fix for some EDA tools that didn't set MAG field in gds file
+                    # maybe this is a excepted behavior in asap7 PDK
```
colinschmidt: expected?

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -0,0 +1,156 @@
...
+                    label.magnification = 0.25
+                    label.translate(-label.position[0]*0.75, -label.position[1]*0.75)
+
+                for ref in v.references:
+                    ref.magnification = 0.25
+                    ref.translate(-ref.origin[0]*0.75, -ref.origin[1]*0.75)
+                    ref.magnification = 1
+
+        # Overwrite original GDS file & set precision to 2.5e-10
+        gds_lib.precision = 2.5e-10
```
colinschmidt: Don't we need to check that the precision was 1e-9 before doing this? The scaling was all done relatively so it should handle whatever precision the original gds had.

src/hammer-vlsi/technology/asap7/asap7.tech.json
```diff
@@ -9,6 +10,26 @@
...
...P7_75t_L\nLVS BOX DECAPx2_ASAP7_75t_SL\nLVS BOX DECAPx2_ASAP7_75t_SRAM\nLVS BOX DECAPx4_ASAP7_75t_R\nLVS BOX DECAPx4_ASAP7_75t_L\nLVS BOX DECAPx4_ASAP7_75t_SL\nLVS BOX DECAPx4_ASAP7_75t_SRAM\nLVS BOX DECAPx6_ASAP7_75t_R\nLVS BOX DECAPx6_ASAP7_75t_L\nLVS BOX DECAPx6_ASAP7_75t_SL\nLVS BOX DECAPx6_ASAP7_75t_SRAM\nLVS BOX DECAPx10_ASAP7_75t_R\nLVS BOX DECAPx10_ASAP7_75t_L\nLVS BOX DECAPx10_ASAP7_75t_SL\nLVS BOX DECAPx10_ASAP7_75t_SRAM\nLVS FILTER DECAPx1_ASAP7_75t_R OPEN\nLVS FILTER DECAPx1_ASAP7_75t_L OPEN\nLVS FILTER DECAPx1_ASAP7_75t_SL OPEN\nLVS FILTER DECAPx1_ASAP7_75t_SRAM OPEN\nLVS FILTER DECAPx2_ASAP7_75t_R OPEN\nLVS FILTER DECAPx2_ASAP7_75t_L OPEN\nLVS FILTER DECAPx2_ASAP7_75t_SL OPEN\nLVS FILTER DECAPx2_ASAP7_75t_SRAM OPEN\nLVS FILTER DECAPx4_ASAP7_75t_R OPEN\nLVS FILTER DECAPx4_ASAP7_75t_L OPEN\nLVS FILTER DECAPx4_ASAP7_75t_SL OPEN\nLVS FILTER DECAPx4_ASAP7_75t_SRAM OPEN\nLVS FILTER DECAPx6_ASAP7_75t_R OPEN\nLVS FILTER DECAPx6_ASAP7_75t_L OPEN\nLVS FILTER DECAPx6_ASAP7_75t_SL OPEN\nLVS FILTER DECAPx6_ASAP7_75t_SRAM OPEN\nLVS FILTER DECAPx10_ASAP7_75t_R OPEN\nLVS FILTER DECAPx10_ASAP7_75t_L OPEN\nLVS FILTER DECAPx10_ASAP7_75t_SL OPEN\nLVS FILTER DECAPx10_ASAP7_75t_SRAM OPEN",
```
colinschmidt: Can this span multiple lines?

src/hammer-vlsi/technology/asap7/defaults.yml
```diff
@@ -5,18 +5,65 @@ technology.asap7:
...
+  # The PaR tool actually runs in the 4x scaled up lef/gds.
+  # after PaR, post_par_script in asap7/__init__.py will scale down the generated gds to 7nm.
+  core.node: 28
   inputs:
     # Supply voltages.
     supplies:
       VDD: "0.70 V"
       GND: "0 V"
+    # mmmc corners config.
+    mmmc_corners: [
```
colinschmidt: This doesn't seem right to me. You should just have a single setup and hold corner for the slowest and fastest corners.

src/hammer-vlsi/technology/asap7/defaults.yml
```diff
@@ -27,8 +74,16 @@ technology.core:
...
+mentor.extra_env_vars:
+  - VAR_M4_YOFFSET: "0"
+  - VAR_M5_XOFFSET: "0"
+  - VAR_M6_YOFFSET: "0"
+  - VAR_M7_XOFFSET: "0"
 
-# Set the layer that blocks vias under bumps
-vlsi.technology.bump_block_cut_layer: "V9"
+par.inputs:
+  # gds_merge: ture is mandatory,
```
colinschmidt: typo: ture->true

jwright6323:
> > We definitely have not addressed the RC scaling. Let us know what Vinay says
> 
> Vinay says:
> Since all the LEFs are magnified by 4x for PnR, you will need to shrink the post-PnR GDS when you import it into Virtuoso.
> 
> I think here is his flow:
> Virtuoso design stdcell ->
> RC extract ->
> 4x stream out gds ->
> innvous PaR
> 
> So as you are right, we should directly extract RC from shrunk gds generated by `post_par_script`.

Bur you need to be able to use Innovus to do extraction so that the tool can time things properly- you don't want to do the extraction on a GDS. It doesn't really matter a whole lot, though, because this isn't a "real" technology, but it would be nice if an extracted SPICE sim would correlate with Innovus timing reports.

sequencer:
And I just verified the script provided by Vinay:
I don't know why `-noDetectVias` was added in that script
cadence help said 
`-noDetectVias \ # <Do Not Detect The Via Defs>` 
I believe we didn't provide any correspond code with `gdspy`.

sequencer:
> Bur you need to be able to use Innovus to do extraction so that the tool can time things properly- you don't want to do the extraction on a GDS. It doesn't really matter a whole lot, though, because this isn't a "real" technology, but it would be nice if an extracted SPICE sim would correlate with Innovus timing reports.

I think in his flow, he extracted RC by Quantus, and provides the result to Innovus then running the timming reports.

So here are three different timming reports:
1. timming report based on `.lib` generated by Liberate 
1.1 we can assume `.lib` is based on 7nm, thus timming report based on netlist generated by synthesis is ok to believe.
2. timming report based on `.lib` + RC data extracted by Quantus(These information are directly annotate to wire)
2.1. RC data extracted by 7nm gds also provides a right result. and timming report based on PaR should use shrunk `PaR` gds to extract correct length for wire and re-input the RC timming information to innovus.
3. timming report based on `pure .sp`, this is somehow a `.lib` characterization flow, in my knowledge, I don't think this use case should appear in the common pure digital flow.

So we may need to abstract RC extraction flow to integrate Cadence Quantus or Mentor Calibre into Hammer, we may firstly give Innvous input with custom SPEF to provide RC data generated manually.

jwright6323:
@sequencer I think we're in agreement and just discussing things in a different way. Let me put it this way- If you draw a 4Wx4H rectangle in the place-and-route tool, the extracted parasitics should match a WxH rectangle extracted from a post-shrunk GDS. It's unclear to me if the QRC techfile we pass to innovus implements this or not.

src/hammer-tech/hammer_tech.py
```diff
@@ -717,6 +731,18 @@ def extract_tarballs(self) -> None:
                             tarfile.open(file).extractall(path=os.path.join(root, f + "_dir"))
                             os.remove(file)
                             os.renames(os.path.join(root, f + "_dir"), file)
+                self.post_install_script()
+
+    def post_install_script(self) -> None:
+        """a script to apply any needed hotfixes to technology libraries, tech __init__.py will override this"""
+        pass
+
+    def post_par_script(self, d: dict) -> None:
```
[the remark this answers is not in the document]
harrisonliew: Hm, as is it will always run after the par tool is done executing which would be incorrect behavior. I'll fix it.

src/hammer-vlsi/technology/asap7/README.md
```diff
@@ -0,0 +1,20 @@
+ASAP7 Technology Library
+===========
+
+HAMMER's default demonstration PDK is [ASAP7](http://asap.asu.edu/asap/). There are some special setup and known issues with this open PDK.
+
+Setup and Environment
+=====================
+
+In addition to requirements for `hammer-vlsi`, using ASAP7 also requires:
+- The PDK tarball must not be pre-extracted, i.e. must specify `technology.asap7.tarball_dir` only.
```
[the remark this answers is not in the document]
harrisonliew: The patched version sits in the tech cache dir of the build directory. This is fine, right?

src/hammer-vlsi/technology/asap7/__init__.py
harrisonliew: Turns out that gdspy could not interpret the original precision in the P&R'd GDS. There might be a better way to do this, I'll investigate.

src/hammer-vlsi/technology/asap7/asap7.tech.json
harrisonliew: Seemingly not: https://stackoverflow.com/questions/53715056/multiline-string-as-a-value-in-json

sequencer:
@jwright6323 I talked with Edward about this question, and it turns out that I misunderstood that QRC was provided in `mmmc_corner`(I used to think only `.lib` is provided), which means that we are using `7nm` QRC to direct `28nm` routing.
I'll ask Vinay this question for more information.

harrisonliew:
@sequencer I clarified with Vinay today already. This is his email:

> The QRCtechfile provided with the PDK is for PnR with scaled geometries. Please see Section V. A and Fig. 9 of our ICCAD paper:
> https://ieeexplore.ieee.org/document/8203889
> 
> Later on, for the PDK 1p5 version, the Calibre parasitic extraction rule deck was updated but the QRC techfile was not updated and the libraries were not re-characterized with the latest Calibre rule deck. Hence, you are bound to see some mismatch between Calibre and QRCtechfile. We hope to fix this in the next release. 

Therefore, we can be quite sure that the method 1) of having synthesis/par analyze timing is fine.

sequencer:
Yes, and we seems not needed to use  `MMMC_corner`, which might be a misleading to PaR tools.

src/hammer-vlsi/technology/asap7/__init__.py
harrisonliew: No, it has to happen later because we have to change the names first, then do the layer placement, then write the individual GDSs after we're done.

src/hammer-vlsi/technology/asap7/__init__.py
harrisonliew: I ended up grabbing the cell list from the PDK's GDS folder to check against specific cells. This way people naming cells with `ASAP7_75t` in them won't be excluded.

src/hammer-vlsi/technology/asap7/__init__.py
harrisonliew: gdspy can import the source units using `read_gds(units='import')` after all.

edwardcwang:
#525 :(

harrisonliew:
I have succesfully pushed a small-ish block through and gotten LVS clean results. Would like to bump this PR for an additional round of review.

After this gets merged, the Chipyard pointer will be bumped and the next item to work on will be adding SRAMs.

edwardcwang:
@sequencer

sequencer:
I'm working on another python script to establish container environment for Hammer, after I finish this I'll rebase my work and directly give a Hammer container generator.(to resolve the EDA environment problem)

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -124,14 +127,19 @@ def scale_gds_script(self, gds_file: str) -> str:
         Note: Need to escape TCL constructs such as [] and {}!
         """
 
-        return """
-#!/usr/bin/python3
+        return """#!/usr/bin/python3
```
sequencer: Why leave this script here, I thought `post_par_script` would work for current case?

harrisonliew: No, the issue that @colinschmidt raised in the review was that `post_par_script` always ran after the `par` action. However, this is wrong because if you say only ran a subset of `par` steps that did not include writing the output GDS, you would run the script when it should not have been run. So, there is no more `post_par_script`.

The other thing this does is demonstrate how you can use a hook provided by a technology. See ucb-bar/chipyard#263.

harrisonliew: I realized that this script is still insufficient for hierarchical P&R. It would be wrong to scale down submodule GDSs and then scale them down again at the top-level. Just checking against the modules listed in the hierarchical placement constraints will also be insufficient because deduplication of referenced cells in the GDS will cause those cells to be shrunk again. This may need to be punted and we'll note for now that ASAP7 cannot be used with hierarchical flow.

sequencer: I understand it now.
should we change our method from editing par gds to patch `28nm -> 7nm` lef in `post_install_script`, so that we can directly use real 7nm process.
This may eventually solve our problem.

harrisonliew: I don't think so, because it will mess up the parasitic extraction if P&R is done at 7nm, unless we also patch the QRC techfile (but I don't know how).

Also, the reason for 28nm was academic Cadence licensing, so I think we should keep it at 28nm so that other schools can still use Hammer with ASAP7.

sequencer: I understand, you are right.
QRC cannot be patched since it is a close-source spec. 
Academic issue is surly the biggest problem.

> this script is still insufficient for hierarchical P&R

Yes that's the problem I used to have, so I set the `gds_merge` default to true.
and also different PaR tool may have different strategy to deliver the gds patterning.
I think we should figure out what should be scaled and what shouldn't to satisfy `gds_merge` both `true` or `false`.

harrisonliew: Okay, let's think this out. Let's say module `A` is a parent of module `B`. Module `A` contains a blackbox `A_bb` and module `B` contains a blackbox `B_bb`.

After P&R-ing module `B`, all shapes in `B` and `B_bb` should be scaled down.

After P&R-ing module `A`, all shapes in `A` and `A_bb` should be scaled but `B` and `B_bb` should not be scaled.

Perhaps we can write out the list of scaled cells when scaling `B` (in format `cellname_filename` as will be generated by the `-uniquify` option in Innovus), then passing this list to its parent module `A` and appending to the `cell_list` variable in the existing script. We will have to traverse the hierarchical tree to make sure we pull the right files, but at least `gds_merge` can always be `true` or `false`.

Any other ideas?

sequencer: > After P&R-ing module B, all shapes in B and B_bb should be scaled down.

How about making scale down as the final step?
We can just use 28nm during all PaR, and leave the scale problem until tape-in.

harrisonliew: We want the ability to do hierarchical DRC & LVS, which requires the leaf modules be already scaled down.

sequencer: Emmm, I see. If this is the only reason, can we do like this?
Module `B` netlist → PaR → gds → scale down → DRC & LVS
------------------------------------------↓
--------------------------------------merge → gds → scale down → DRC & LVS
------------------------------------------↑
Module `A` netlist → PaR → gds → scale down → DRC & LVS
But this method might introduce some boundary problem to DRC & LVS after merging.

harrisonliew: Maybe, but if module `A` and `B` don't have merging done, DRC & LVS won't be valid. We would need to have another merge & scale step (which would create 2 GDSs).

Also, if you scale down `A` and `B` without merging hard macros like analog IP blocks, the final scale down step will need to scale down those hard macros without also scaling down geometries generated in `A` and `B`. I think this is just a variation of the same problem.

sequencer: Yes. This method is only for module level check DRC & LVS,
my approach will add additional trouble to this issue.
I'm read the doc from `INNOVUS/doc/innovusUG/Importing_and_Exporting_Designs.html`
will this give a help for this problem?

harrisonliew: I think it just supports my thought that we can take advantage of the `-uniquify` appending the module name in front of the cell name to filter out cells that have already been scaled. Other than that, I don't see much else that's helpful in that guide.

sequencer: Yes, I do think so.

harrisonliew: Let's merge this PR and address #533 in a different PR.

PR #512, opened 2019-09-07
Tech hook

sequencer:
Have a try for this patch:)
This patch finishes this TODO:
```# TODO: execute this script differently depending on script type. Assumed to be shell for now.```
if a technology folder contains `__init__.py`
`hammer_tech` will try to load `tech` from technology folder, which allows overriding to `post_install_script()`, you can also source any bash script or use python directly to access any variable from `HammerTechnology`.
This patch will affect hammer core codebase, may need reviewed by @edwardcwang and @jwright6323.

### CODE REVIEW

File: src/hammer-vlsi/hammer_vlsi/driver.py [hunk 1]
```diff
@@ -140,6 +140,7 @@ def load_technology(self, cache_dir: str = "") -> None:
         self.log.info("Loading technology '{0}'".format(tech_str))
         tech_opt = None  # type: Optional[hammer_tech.HammerTechnology]
         for base_path in tech_paths:
+            sys.path.append(base_path)
```
edwardcwang: Related change?

src/hammer-tech/hammer_tech.py
```diff
@@ -386,7 +387,12 @@ def load_from_json(cls, technology_name: str, json_str: str, path: str) -> "Hamm
         :param path: Path to set as the technology folder (e.g. foo/bar/technology/asap7)
         """
 
-        tech = HammerTechnology()
+        # try to override tech, if __init__.py exist.
```
edwardcwang: If we are to make this change we should document somewhere this format for `__init__.py`:

```python
from hammer_tech import HammerTechnology

class MyTech(HammerTechnology):
  ...

tech = MyTech()
```

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+
+from hammer_tech import HammerTechnology
+
+class ASAP7Tech(HammerTechnology):
+    def post_install_script(self) -> None:
+        self.remove_duplication_in_drc_lvs()
+        self.generate_multi_vt_gds()
+        self.fix_sram_cdl_bug()
+
+    def remove_duplication_in_drc_lvs(self) -> None:
```
edwardcwang: Can you add pydoc explaining the context behind this hotfix?

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+            shutil.copystat(drc_deck, tf.name)
+            shutil.copy(tf.name, drc_deck)
+
+        with tempfile.NamedTemporaryFile(delete=False) as tf:
+            with open(lvs_deck, 'r') as f:
+                tf.write(pattern.sub("", f.read()).encode('utf-8'))
+            shutil.copystat(lvs_deck, tf.name)
+            shutil.copy(tf.name, lvs_deck)
+
+    def generate_multi_vt_gds(self) -> None:
```
edwardcwang: ditto

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+                            new_el = deepcopy(el)
+                            new_el.layer = num
+                            cell.append(new_el)
+
+            # Write to new GDS
+            new_gds = os.path.splitext(orig_gds)[0] + '_' + vt + '.gds'
+            with open(new_gds, 'wb') as stream:
+                new_lib.save(stream)
+
+    def fix_sram_cdl_bug(self) -> None:
```
edwardcwang: ditto

src/hammer-vlsi/hammer_vlsi/driver.py
sequencer: This will append technology search dir as a python path for importing modules like `import asap7`

src/hammer-tech/hammer_tech.py
sequencer: What's your suggestion? I think currently we can leave `technology/asap/__init__.py` as an example is ok?

edwardcwang: Yeah leaving the asap7 example is fine; we may also want to add it to a readme - maybe `src/hammer-tech/README.md`?

sequencer: I just tested the `sphinx` doc generation, it works well, I think we may not need to add a `src/hammer-tech/README.md`

sequencer:
BTW, if shell has already provided a convenient tools like `sed` `patch`, we can directly use [sh](https://amoffat.github.io/sh/) for easy integration with shell. I did these patch with pure python standard library just because of the OCD:)

src/hammer-tech/hammer_tech.py
edwardcwang: Hmm, how about this?

1) Let's put this sketch into the docstring for the `HammerTechnology` class

2) Can we add a unit test that tests this behaviour?

sequencer: great! working on this now.

PR #514, opened 2019-09-11
Specify technology time units

### LINKED ISSUE #513: Technology schema should include library time unit
ASAP7's libs are all specified in `ps`, and if we give the timing constraints as `ns`, certain synthesis tools will only do timing analysis in the library's time units.

Would suggest adding a property similar to `grid_units` and parsing similarly and the default should still be `ns`.

harrisonliew:
Fixes #513

### DIFF
--- a/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
+++ b/src/hammer-vlsi/hammer_vlsi/hammer_tool.py
@@ -870,6 +870,13 @@ class HammerTool(metaclass=ABCMeta):
             output.append(clock)
         return output
 
+    def get_time_unit(self) -> str:
+        """
+        Return the library time unit.
+        """
+        return self.technology.config.time_unit if self.technology.config.time_unit is not None else "ns"
+
+
     def get_all_supplies(self, key: str) -> List[Supply]:
         supplies = self.get_setting(key)
         output = []  # type: List[Supply]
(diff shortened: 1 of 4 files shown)


src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -870,6 +870,13 @@ def get_clock_ports(self) -> List[ClockPort]:
             output.append(clock)
         return output
 
+    def get_time_unit(self) -> str:
+        """
+        Return the library time unit.
+        """
+        return self.technology.config.time_unit if self.technology.config.time_unit is not None else "ns"
```
jwright6323: I think this should return a `TimeValue`

harrisonliew:
It appears the `ValueWithUnits` doesn't currently expose the specified units (e.g. the "p" in "ps), so it's clunky to translate to a `TimeValue` first and then extract that `TimeValue`'s units to pass to `value_in_units` when writing the clock constraints. To make this work I'd have to expose the private value `_value_prefix`.

That said, I think ideally we want to just parse this from the technology's libs, which has the format `time_unit : "1ps";`

edwardcwang:
I think it's fine to just translate to a TimeValue first and then extract that TimeValue's units to pass to value_in_units when writing the clock constraints as you propose. The TimeValue is meant to be precision-neutral.

PR #520, opened 2019-09-18
power_spec_type is not optional, null is invalid

sequencer:
https://github.com/ucb-bar/hammer/blob/master/src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py#L1655 
type is not optional.

### DIFF
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -345,7 +345,7 @@ vlsi.inputs:
   # Valid options:
   # cpf - Use the common power format commonly used in Cadence tools
   # upf - Use the universal power format, IEEE 1801-2015
-  power_spec_type: null
+  power_spec_type: upf
   # Optional: Contents of a power specification in the above type (str)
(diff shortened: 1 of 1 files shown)


src/hammer-vlsi/defaults.yml
```diff
@@ -345,7 +345,7 @@ vlsi.inputs:
   # Valid options:
   # cpf - Use the common power format commonly used in Cadence tools
   # upf - Use the universal power format, IEEE 1801-2015
-  power_spec_type: null
+  power_spec_type: upf
```
colinschmidt: I think we should use cpf as a default given that has the most supported

edwardcwang: On second thought since the default is `power_spec_mode: empty`, if a user sets `power_spec_mode` to manual without specifying a value, shouldn't it trigger an error?

harrisonliew:
Closing since use of `power_spec_type` is still tool vendor-specific.

Issue #502, opened 2019-09-01
Stackup & power straps APIs should include checks against the WIDTHTABLE


harrisonliew:
Ran into issues with ASAP7:

The WIDTHTABLE parameter in the tech LEF may specify a discrete set of allowable wire widths, and then any width is allowed after the last number in the table.

The power straps API (when calculating wire widths from the `track_width` parameter) doesn't know about these discrete widths and thus the par tool will not draw the illegal width power straps. This should be checked in Hammer.

harrisonliew:
Closed by #521

Issue #497, opened 2019-08-22
Power strap density check


dpgrubb13:
Have hammer check power strap density based on tech specific rules.

harrisonliew:
Closed by #521

PR #521, opened 2019-09-19
Power strap API checks density & against width table

harrisonliew:
First attempt at fixing #497 & #502

Optional tech schema for quantized wire widths (LEF58_WIDTHTABLE)
- Checks desired strap width against provided table
- Quantizes down, except for highest value in table, for which all widths at least that one are allowed
- Issues logger warning if quantization happened (I don't like how I implemented this)

Simple check for strap density (sum of widths/pitch), warns if over 85%.

### DIFF
--- a/src/hammer-tech/schema.json
+++ b/src/hammer-tech/schema.json
@@ -307,13 +307,7 @@
                   "type": "array",
                   "items": {
                     "title": "WidthTableEntry",
-                    "type": "object",
-                    "additionalproperties": false,
-                    "properties": {
-                      "width": {
-                        "type": "number"
-                      }
-                    }
+                    "type": "number"
                   }
                 }
               }
--- a/src/hammer-tech/schema.json
+++ b/src/hammer-tech/schema.json
@@ -302,6 +302,19 @@
                       }
                     }
                   }
+                },
+                "power_strap_width_table": {
+                  "type": "array",
+                  "items": {
+                    "title": "WidthTableEntry",
+                    "type": "object",
+                    "additionalproperties": false,
+                    "properties": {
+                      "width": {
+                        "type": "number"
+                      }
+                    }
+                  }
                 }
               }
             }
--- a/src/hammer-tech/stackup.py
+++ b/src/hammer-tech/stackup.py
@@ -7,7 +7,7 @@
 #  See LICENSE for licence details.
 
 from enum import Enum
-from typing import List, NamedTuple, Tuple, Dict
+from typing import List, NamedTuple, Tuple, Dict, Optional
 from hammer_utils import reverse_dict, coerce_to_grid
 from hammer_logging import HammerVLSILoggingContext
 from decimal import Decimal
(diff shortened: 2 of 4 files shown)


src/hammer-tech/stackup.py
```diff
@@ -96,6 +97,15 @@ def from_list(grid_unit: Decimal, l: List[dict]) -> List["WidthSpacingTuple"]:
             current_spacing = wst.min_spacing
         return out
 
+class WidthTableEntry(dict):
+    """
+    An allowable wire width from technology LEF width table
```
edwardcwang:
```suggestion
    An allowable wire width from technology LEF width table.
```

src/hammer-tech/stackup.py
```diff
@@ -225,7 +239,36 @@ def max_width_from_pitch(self, pitch: Decimal) -> Decimal:
...
-    def get_width_spacing_start_twt(self, tracks: int) -> Tuple[Decimal, Decimal, Decimal]:
+    def quantize_to_width_table(self, width: Decimal, layer: str, logger: Optional[HammerVLSILoggingContext]) -> Decimal:
+        """
+        Compare a desired width to the width table, if specified for a technology.
+        Will return the allowable width smaller than or equal to the desired width,
+        except if the desired width is greater than or equal to the last width in the width table.
+        Issues a logger warning for the user if the returned width was quantized.
+        """
+        width_table = self.power_strap_width_table
+        if width_table == []:
```
edwardcwang:
```suggestion
        if len(width_table) == 0:
```

src/hammer-tech/schema.json
```diff
@@ -302,6 +302,19 @@
...
+                },
+                "power_strap_width_table": {
+                  "type": "array",
+                  "items": {
+                    "title": "WidthTableEntry",
+                    "type": "object",
+                    "additionalproperties": false,
+                    "properties": {
+                      "width": {
+                        "type": "number"
```
edwardcwang: @edwardcwang Take a look

src/hammer-tech/stackup.py
```diff
@@ -225,7 +239,36 @@ def max_width_from_pitch(self, pitch: Decimal) -> Decimal:
         """
         return self.min_spacing_and_max_width_from_pitch(pitch)[1]
 
-    def get_width_spacing_start_twt(self, tracks: int) -> Tuple[Decimal, Decimal, Decimal]:
+    def quantize_to_width_table(self, width: Decimal, layer: str, logger: Optional[HammerVLSILoggingContext]) -> Decimal:
```
edwardcwang: Could we unit test this method?

harrisonliew: Did a test using two identical metal objects, one with a width table and one without (just 1 entry of min width). A little klugey but I didn't want to mess up the other existing power strap tests and we probably want to improve the behavior in a future PR anyway to adjust the spacing as well.

edwardcwang: > we probably want to improve the behavior in a future PR anyway to adjust the spacing as well

👍

src/hammer-tech/schema.json
edwardcwang: Solved!

PR #529, opened 2019-10-03
Update many docs

harrisonliew:
Working in parallel with @dpgrubb13.

### DIFF
--- a/doc/Hammer-Flow/LVS.rst
+++ b/doc/Hammer-Flow/LVS.rst
@@ -1,3 +1,59 @@
 LVS
 ===============================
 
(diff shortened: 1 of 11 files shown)


doc/Hammer-Flow/LVS.rst
```diff
@@ -1,3 +1,59 @@
 LVS
 ===============================
 
+Hammer has an action for running LVS decks on a post-P&R'd GDS and gate-levl netlist. This action invokes a ``HammerLVSTool`` that must be provided as part of a tool plugin.
+
+LVS Setup Keys
+--------------
+
+* Namespace: ``vlsi.core``
```
edwardcwang: I think it would be good not to duplicate these with the definitions. DRY

PR #531, opened 2019-10-04
Post par sim

dpgrubb13:
Fixes for post PAR simulation.


harrisonliew:
LGTM. Quick merge so that 151 students can do post-par sim.

PR #535, opened 2019-10-10
fixes #476, json now equal with yml

### LINKED ISSUE #476: json takes priority over yaml?
Apparently `-p foo.json -p bar.yaml` will not use any keys in `bar.yaml` to override keys in `foo.json`. From the documentation this seems to be intentional, but I don't really understand the reasoning behind it. Why not go left to right? This has caused a number of problems now, and is actually pretty annoying to work around.

harrisonliew:
Please comment if you think there's a philosophical reason to keep this, otherwise it blocks some functionality because one cannot just `-p` another yml with small settings changes, e.g. when doing a `make redo-par` that takes the `par-output-full.json`.

### DIFF
--- a/src/hammer_config/config_src.py
+++ b/src/hammer_config/config_src.py
@@ -955,13 +955,14 @@ def load_config_from_paths(config_paths: Iterable[str], strict: bool = False) ->
 def load_config_from_defaults(path: str, strict: bool = False) -> List[dict]:
     """
     Load the default configuration for a hammer-vlsi tool/library/technology in
-    the given path, which consists of defaults.yml and defaults.json.
+    the given path, which consists of defaults.yml and defaults.json (with
+    defaults.json taking priority).
 
     :param config_paths: Path to defaults.yml and defaults.json.
     :param strict: Set to true to error if the file is not found.
     :return: A list of configs in order of specification.
     """
     return load_config_from_paths([
-        os.path.join(path, "defaults.yml"),
-        os.path.join(path, "defaults.json")
+        os.path.join(path, "defaults.json"),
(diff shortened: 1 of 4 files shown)


src/hammer_config/config_src.py
```diff
@@ -944,28 +944,22 @@ def combine_meta(config_dict: dict, meta_setting: str) -> dict:
...
-    return list(map(lambda path: load_config_from_file(path, strict), sorted_paths))
+    return list(map(lambda path: load_config_from_file(path, strict), config_paths))
 
 
 def load_config_from_defaults(path: str, strict: bool = False) -> List[dict]:
     """
     Load the default configuration for a hammer-vlsi tool/library/technology in
-    the given path, which consists of defaults.yml and defaults.json (with
-    defaults.json taking priority).
+    the given path, which consists of defaults.yml and defaults.json.
```
edwardcwang: Should note here that the .yml takes precedence now. Or switch the order below.

harrisonliew: I'll switch the order and make a note.

harrisonliew: Wait, this is already correct. I didn't need to edit this method.

jwright6323: I misread this when I thumbs-up'ed- I thought this was a suggestion. They are equal in priority now, correct? What is the order of precedence, left-to-right?

harrisonliew: The one later in the list has precedence, so right-to-left, I guess?

jwright6323: Got it, that makes sense to me. I'd document that.

PR #540, opened 2019-10-23
make special cells return None if not defined in tech plugin

dpgrubb13:
This is meant to make it so the tech plugin does not have to use the special cells API.  The special cells API is currently pretty simple and would typically be overridden by a hook, so I think it is better to allow the user to not define any special cells in the tech plugin if they plan on using a hook anyways.


edwardcwang:
Needs to return empty list instead of None if you are to do this.

dpgrubb13:
Yeah I was going to fix this in conjunction with opening a cad plugins PR.

edwardcwang:
That's fine, and I think those are orthogonal issues...?

harrisonliew:
LGTM, tested working in a simple ASAP7 design.

PR #541, opened 2019-10-23
Check invalid tarballs

harrisonliew:
Addresses a small issue that came up where tech cache was created despite an invalid tarball was specified, so it skipped the post-extraction hacking for ASAP7.

### DIFF
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -712,6 +712,9 @@ class HammerTechnology:
         for tarball in self.config.tarballs:
             target_path = os.path.join(self.extracted_tarballs_dir, tarball.path)
             tarball_path = os.path.join(self.get_setting(tarball.base_var), tarball.path)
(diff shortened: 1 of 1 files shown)


src/hammer-tech/hammer_tech.py
```diff
@@ -712,6 +712,9 @@ def extract_tarballs(self) -> None:
         for tarball in self.config.tarballs:
             target_path = os.path.join(self.extracted_tarballs_dir, tarball.path)
             tarball_path = os.path.join(self.get_setting(tarball.base_var), tarball.path)
+            if not os.path.isfile(tarball_path):
+                raise ValueError("Path {0} does not point to a valid tarball!".format(tarball_path))
+                continue
```
edwardcwang: Isn't this `continue` redundant?

harrisonliew: True, `ValueError` handles it. It's left from when I was doing a logger error instead.

Issue #474, opened 2019-06-28
Tech plugin tool version support


jwright6323:
This is related to #447. It should be easy to describe what tools a particular tech plugin supports, and also their minimum versions. I think this is a difficult thing to get right, so we need to put some thought into it.

Issue #198, opened 2018-09-20
Plugin-common code/defaults don't have a good place to live


edwardcwang:
$subject

jwright6323:
See #281

jwright6323:
To be more descriptive, there's no good place to put Cadence-wide common code, for example.

PR #543, opened 2019-11-06
Updates to step infrastructure

### LINKED ISSUE #87: Can't insert hooks before default first_step
Not a huge deal because the first steps are short and don't affect QoR but still would be nice.

### LINKED ISSUE #124: Make it clear when you try to hook onto a step that does not exist
Right now it just says "Target step xyz does not exist" which can be confusing to users who may think they specified a wrong step for only_step/etc.

### LINKED ISSUE #310: --to_step vs --through_step
It might be useful to have separate `--to_step` and `--through_step` CLI options. Currently the `--to_step` option behaves like what I would imagine `--through_step` would, by executing the step inclusively. IMO this is the more useful variant, so it's fine as-is, but if we also want a non-inclusive variant the names should change.

### LINKED ISSUE #311: Tech plugins should be able to override hooks
e.g. <mytech> plugin should be able to provide hooks in a programmatic way

### LINKED ISSUE #335: Document how to use Hammer hooks
$subject

### LINKED ISSUE #430: Steps need concept of state
In certain tools, there are certain settings that do not get written to their databases between steps. E.g. standard-cell cell obstructions that must be deleted for DRC early on are not carried over when you do a --from_step or --only_step <later step>.

We will need a way to figure out which sub-commands apply to what part in the design and insert them appropriately.

harrisonliew:
This PR fixes multiple issues:
* Closes #310 by implementing `--after_step` and `--until_step` command line arguments. These exclude the specified steps, with appropriate legalization checks.
* Closes #311 by implementing the ability for the tech plugin subclasses to implement a `get_tech_<action>_hooks` so that it can add hooks and the tool plugin to also implement hooks. The priority is implemented also.
* Closes #430 by implementing a concept of "persistent" hooks. Persistent = always run at beginning, pre_persistent = always run at beginning unless first step to run is after the target step, post_persistent = always run at beginning if target step is before the first step to run.
* Closes #87 (I believe) by being able to insert hooks w.r.t. any step.
* Closes #335 (documentation)
* Closes #124 by differentiating "target step does not exist error" message between those specified by command line vs. hooks
* Will also close ucb-bar/hammer-cadence-plugins#5 along with a companion PR.
* Maybe addresses #474 because tech hooks are matched to specific tool names, and version checks can be done in the hooks themselves. Open to discussion about how to make this more robust.
* May also address #198 if plugin-common code can live as hook methods in any file.

In the process, I also unified the resume/pause implementation in `run_steps` and simplified the flow logic as well. It only prints "Running sub-step..." at debug level if it's actually running it, otherwise will indicate "Sub-step skipped..." at info level. Logger printing of step execution is now in the correct order of execution.

Also fixed up docs formatting/linking while I was at it.

Asking for comment on whether this is worth fixing:
Currently, tech/tool hooks have to have the hook function outside of the subclass (of HammerTechnology/HammerTool). This is non-ideal from a code readability standpoint because it has to live way at the top/bottom of the file or in a different file altogether, instead of intermingled with steps. However, we can't place them into the class because instance methods aren't callable. Thoughts?

### DIFF
--- a/doc/Hammer-Basics/Hammer-Overview.rst
+++ b/doc/Hammer-Basics/Hammer-Overview.rst
@@ -31,7 +31,7 @@ Tech Plugins
 
 A techonology plugin consists of two or more files: a ``*.tech.json`` and a ``defaults.yml``.
 
-The ``*.tech.json`` contains pointers to relevant PDK files and fundamental technology constants.  These values are not meant to be overriden, nor can they be. 
+The ``*.tech.json`` contains pointers to relevant PDK files and fundamental technology constants.  These values are not meant to be overriden, nor can they be for the time being. 
 
 ``defaults.yml`` sets default technology variables for Hammer to consume, which may be specific to this technology or generic to all. These values may be overriden by design-specific configurations. An example of this is shown in ``hammer/src/hammer-vlsi/technology/asap7/``, and how to setup a technology plugin is documented in more detail in the :ref:`technology` section.
 
--- a/doc/Hammer-Basics/Hammer-Overview.rst
+++ b/doc/Hammer-Basics/Hammer-Overview.rst
@@ -29,7 +29,11 @@ This demonstrates two different namespaces, ``vlsi.core`` and ``vlsi.inputs``, a
 Tech Plugins
 -------------------------------
 
-A techonology plugin consists of two or more files: a ``*.tech.json`` and a ``defaults.yml``.  The ``*.tech.json`` contains pointers to relevant PDK files and fundamental technology constants.  These values are not meant to be overriden, nor can they be. ``defaults.yml`` sets default technology variables for Hammer to consume, which may be specific to this technology or generic to all. These values may be overriden by design-specific configurations. An example of this is shown in ``hammer/src/hammer-vlsi/technology/asap7/``, and how to setup a technology plugin is documented in more detail in the Technology portion of the docs.
+A techonology plugin consists of two or more files: a ``*.tech.json`` and a ``defaults.yml``.
+
+The ``*.tech.json`` contains pointers to relevant PDK files and fundamental technology constants.  These values are not meant to be overriden, nor can they be. 
+
+``defaults.yml`` sets default technology variables for Hammer to consume, which may be specific to this technology or generic to all. These values may be overriden by design-specific configurations. An example of this is shown in ``hammer/src/hammer-vlsi/technology/asap7/``, and how to setup a technology plugin is documented in more detail in the :ref:`technology` section.
 
 Tool Plugins
 -------------------------------
@@ -66,10 +70,15 @@ Using hooks requires the designer to extend the ``CLIDriver`` class. A good exam
 In both of these commands, an environment configuration is passed to Hammer using a ``-e`` flag, which in this case is ``env.yml``.
 ``env.yml`` contains pointers to the required tool licenses and environment variables.
 These environment settings will not be propagated to the output configuration files after each action.
+
 Any number of other YML or JSON files can then be passed in using the ``-p`` flag.
 In this case, there is only one, ``config.yml``, and it needs to set all the required keys for the step of the flow being run.
+
 ``--obj_dir build`` designates what directory Hammer should use as a working directory.
 All default action run directories and output files will be placed here.
+
 Finally, ``par`` designates that this is a place-and-route action.
+
 In this case, Hammer will write outputs to the path ``$PWD/build/par-rundir``.
+
 For the full list of Hammer command-line arguments, look in the ``src/hammer-vlsi/hammer_vlsi/cli_driver.py`` file.
--- a/doc/Hammer-Basics/Hammer-Overview.rst
+++ b/doc/Hammer-Basics/Hammer-Overview.rst
@@ -72,4 +72,4 @@ In this case, there is only one, ``config.yml``, and it needs to set all the req
 All default action run directories and output files will be placed here.
 Finally, ``par`` designates that this is a place-and-route action.
 In this case, Hammer will write outputs to the path ``$PWD/build/par-rundir``.
-For the full list of Hammer command-line arguments, look in the ``src/hammer-vlsi/hammer_vlsi/driver.py`` file.
+For the full list of Hammer command-line arguments, look in the ``src/hammer-vlsi/hammer_vlsi/cli_driver.py`` file.
--- a/doc/Hammer-Basics/Hammer-Overview.rst
+++ b/doc/Hammer-Basics/Hammer-Overview.rst
@@ -81,4 +81,4 @@ Finally, ``par`` designates that this is a place-and-route action.
 
 In this case, Hammer will write outputs to the path ``$PWD/build/par-rundir``.
 
-For the full list of Hammer command-line arguments, look in the ``src/hammer-vlsi/hammer_vlsi/cli_driver.py`` file.
+For the full list of Hammer command-line arguments, run `hammer-vlsi --help` or take a peek in the ``src/hammer-vlsi/hammer_vlsi/cli_driver.py`` file.
--- a/doc/Hammer-Use/Flow-Control.rst
+++ b/doc/Hammer-Use/Flow-Control.rst
@@ -14,10 +14,10 @@ Command-line Interface
 
 Flow control is specified with the following optional Hammer command-line flags, which must always target a valid step/hook:
 
-* ``--from_step <target>``: this starts the tool from (inclusive) the target step
-* ``--after_step <target>``: this starts the tool from (exclusive) the target step
-* ``--to_step``: this stops the tool at (inclusive) the target step
-* ``--until_step``: this stops the tool at (exclusive) the target step
+* ``--start_before_step <target>``: this starts the tool from (inclusive) the target step. Alternate flag: ``--from_step``
+* ``--start_after_step <target>``: this starts the tool from (exclusive) the target step. Alternate flag: ``--after_step``
+* ``--stop_after_step``: this stops the tool at (inclusive) the target step. Alternate flag: ``--to_step``
+* ``--stop_before_step``: this stops the tool at (exclusive) the target step. Alternate flag: ``--until_step``
 * ``--only_step``: this only runs the target step
 
 As ``hammer-vlsi`` is parsing through the steps for a given tool, it will print debugging information indicating which steps are skipped and which are run.
@@ -25,10 +25,10 @@ As ``hammer-vlsi`` is parsing through the steps for a given tool, it will print 
 Certain combinations are not allowed:
 
 * ``--only_step`` is not compatible with any of the other flags
-* ``--from_step`` and ``after_step`` may not be specified together
-* ``--to_step`` and ``--until_step`` may not be specified together
-* ``--from_step`` and ``--until_step`` may not be the same step
-* ``--after_step`` and ``--to_step`` may not be the same step
-* ``--after_step`` and ``--until_step`` may not be the same or adjacent steps
+* ``--start_before_step`` and ``start_after_step`` may not be specified together
+* ``--stop_after_step`` and ``--stop_before_step`` may not be specified together
+* ``--start_before_step`` and ``--stop_before_step`` may not be the same step
+* ``--start_after_step`` and ``--stop_after_step`` may not be the same step
+* ``--start_after_step`` and ``--stop_before_step`` may not be the same or adjacent steps
 
-Logically, a target of ``to/until_step`` before ``from/after_step`` will also not run anything (though it is not explicitly checked).
+Logically, a target of ``stop_{before|after}_step`` before ``start_{before|after}_step`` will also not run anything (though it is not explicitly checked).
--- a/doc/Hammer-Use/Flow-Control.rst
+++ b/doc/Hammer-Use/Flow-Control.rst
@@ -1,0 +1,34 @@
+.. _flow-control:
+
+Flow Control
+=======================================
+
+Physical design is necessarily an iterative process, and designers will often require fine control of the flow within any given action.
+This allows for rapid testing of new changes in the Hammer IR  to improve the quality of results after certain steps.
+
+Given the flow defined by tool steps and hooks (described in the next section), users can select ranges of these to run.
+The tool script will then be generated with commands corresponding to only the steps that are to be run.
+
+Command-line Interface
+----------------------
+
+Flow control is specified with the following optional Hammer command-line flags, which must always target a valid step/hook:
+
+* ``--from_step <target>``: this starts the tool from (inclusive) the target step
+* ``--after_step <target>``: this starts the tool from (exclusive) the target step
+* ``--to_step``: this stops the tool at (inclusive) the target step
+* ``--until_step``: this stops the tool at (exclusive) the target step
+* ``--only_step``: this only runs the target step
+
+As ``hammer-vlsi`` is parsing through the steps for a given tool, it will print debugging information indicating which steps are skipped and which are run.
+
+Certain combinations are not allowed:
+
+* ``--only_step`` is not compatible with any of the other flags
+* ``--from_step`` and ``after_step`` may not be specified together
+* ``--to_step`` and ``--until_step`` may not be specified together
+* ``--from_step`` and ``--until_step`` may not be the same step
+* ``--after_step`` and ``--to_step`` may not be the same step
+* ``--after_step`` and ``--until_step`` may not be the same or adjacent steps
+
+Logically, a target of ``to/until_step`` before ``from/after_step`` will also not run anything (though it is not explicitly checked).
--- a/doc/Hammer-Use/Flow-Control.rst
+++ b/doc/Hammer-Use/Flow-Control.rst
@@ -4,7 +4,7 @@ Flow Control
 =======================================
 
 Physical design is necessarily an iterative process, and designers will often require fine control of the flow within any given action.
-This allows for rapid testing of new changes in the Hammer IR  to improve the quality of results after certain steps.
+This allows for rapid testing of new changes in the Hammer IR  to improve the quality of results (QoR) after certain steps.
 
 Given the flow defined by tool steps and hooks (described in the next section), users can select ranges of these to run.
 The tool script will then be generated with commands corresponding to only the steps that are to be run.
--- a/doc/Hammer-Use/Hooks.rst
+++ b/doc/Hammer-Use/Hooks.rst
@@ -60,7 +60,8 @@ Hooks may be provided by the technology plugin, the tool plugin, and/or the user
 
 A technology plugin specifies hooks in its ``__init__.py`` (as a method inside its subclass of ``HammerTechnology``). It should implement a ``get_tech_<action>_hooks(self, tool_name: str)`` method. The tool name parameter may be checked by the hook implementation because multiple tools may implement the same action. Technology plugin hooks may only target technology default steps to insert/replace.
 
-The linked Chipyard example includes an example of how the ASAP7 technology plugin injects a hook to scale down a GDS post-place-and-route in Innovus.
+The included ASAP7 technology plugin provides an example of how to inject two different types of hooks: 1) a persistent hook invoked anytime after the ``init_design`` step to set top & bottom routing layers, and 2) a hook to scale down a GDS post-place-and-route using the ``gdspy`` package.
+Note that the persistent hook that is included does not necessarily need to be persistent (Innovus does retain this setting in snapshot databases), but it serves as an example for building your own tech plugin.
 
 A tool plugin specifies hooks in its ``__init__.py`` (as a method inside its subclass of ``HammerTool``). It should implement a ``get_tool_hooks(self)`` method. In contrast to the tech-supplied hooks, the action name and tool name are not specified because a tool instance can only correspond to a single action. Tool plugin hooks may target technology default steps and technology plugin hooks to insert/replace.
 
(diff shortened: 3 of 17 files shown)


src/hammer-vlsi/hammer_vlsi/cli_driver.py
```diff
@@ -396,13 +396,21 @@ def action(driver: HammerDriver, append_error_func: Callable[[str], None]) -> Op
...
                     return None
                 else:
                     post_load_func_checked(driver)
-                assert driver.syn_tool is not None, "load_synthesis_tool was successful"
-                success, output = driver.run_synthesis(extra_hooks)
+                assert driver.syn_tool is not None, "load_synthesis_tool was unsuccessful"
+                all_hooks = driver.syn_tool.get_tool_hooks() + \
+                            driver.tech.get_tech_syn_hooks(driver.syn_tool.name) + \
+                            list(extra_hooks or [])
+                success, output = driver.run_synthesis(all_hooks if len(all_hooks) > 0 else None)
```
edwardcwang: I believe you can just get away with the following here
```suggestion
                success, output = driver.run_synthesis(all_hooks)
```

src/hammer-vlsi/hammer_vlsi/cli_driver.py
```diff
@@ -415,8 +423,11 @@ def action(driver: HammerDriver, append_error_func: Callable[[str], None]) -> Op
                     return None
                 else:
                     post_load_func_checked(driver)
-                assert driver.par_tool is not None, "load_par_tool was successful"
-                success, output = driver.run_par(extra_hooks)
+                assert driver.par_tool is not None, "load_par_tool was unsuccessful"
+                all_hooks = driver.par_tool.get_tool_hooks() + \
+                            driver.tech.get_tech_par_hooks(driver.par_tool.name) + \
+                            list(extra_hooks or [])
+                success, output = driver.run_par(all_hooks if len(all_hooks) > 0 else None)
```
edwardcwang: ditto

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -400,47 +407,92 @@ def has_step(name: str) -> bool:
...
+
+                for i in range(len(new_steps)):
+                    if new_steps[i].name == action.target_name:
+                        step_id = i
+                        break
+                for i in range(len(persistent_steps)):
+                    if persistent_steps[i].step.name == action.target_name:
+                        pstep_id = i
+                        break
+                assert bool(step_id != -1) != bool(pstep_id != -1)
```
edwardcwang: Is there a reason the `bool` is needed?

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -400,47 +407,92 @@ def has_step(name: str) -> bool:
         # Copy the list of steps
         new_steps = list(steps)
 
-        # Where to resume, if such a hook exists
+        # Persistent steps are processed differently from normal steps
+        persistent_steps = []  # type: List[HammerToolHookAction]
```
edwardcwang: Might want to clarify why this shouldn't be HammerToolStep

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -452,57 +504,83 @@ def has_step(name: str) -> bool:
                 # Cajole the type checker into accepting that step is a HammerToolStep
                 step = cast(HammerToolStep, step)
                 check_hammer_step_function(step.func)
+        for pstep in persistent_steps:
+            if not isinstance(pstep, HammerToolHookAction):
+                raise ValueError("Element in List[HammerToolHookAction] is not a HammerToolHookAction")
```
edwardcwang: Should this be HammerToolStep?

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -452,57 +504,83 @@ def has_step(name: str) -> bool:
...
+            else:
+                # Cajole the type checker into accepting that pstep.step is a HammerToolStep
+                step = cast(HammerToolStep, pstep.step)
```
edwardcwang: Ditto?

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -596,12 +674,12 @@ def make_insertion_hook(step: str, location: HookLocation, func: HammerStepFunct
         )
 
     @staticmethod
-    def make_resume_hook(step: str, location: HookLocation) -> HammerToolHookAction:
+    def make_resume_pause_hook(step: str, location: HookLocation) -> HammerToolHookAction:
         """
-        Create a hook action is inserted relative to the given step.
+        Create a hook action which will start/stop the execution of the tool at/after the given step.
```
edwardcwang: Clarify in pydoc that `location` controls what happens exactly

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -191,4 +193,29 @@ def scale_gds_script(self, gds_file: str) -> str:
...
         """.format(cell_list_file=os.path.join(self.extracted_tarballs_dir, 'ASAP7_PDKandLIB.tar/ASAP7_PDKandLIB_v1p5/asap7libs_24.tar.bz2/asap7libs_24/gds/cell_list.txt'), gds_file=gds_file)
 
+    def get_tech_par_hooks(self, tool_name: str) -> List[HammerToolHookAction]:
+        hooks = {"innovus": [
+            HammerTool.make_post_insertion_hook("write_design", scale_final_gds)
+            ]}
+        return hooks.get(tool_name, [])
+
+def scale_final_gds(ht: HammerTool) -> bool:
+    assert isinstance(ht, HammerPlaceAndRouteTool)
```
edwardcwang:
```suggestion
    assert isinstance(ht, HammerPlaceAndRouteTool), "scale_final_gds can only be run on par"
```

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -191,4 +193,29 @@ def scale_gds_script(self, gds_file: str) -> str:
...
+    assert isinstance(ht, CadenceTool)
```
edwardcwang:
```suggestion
    assert isinstance(ht, CadenceTool), "scale_final_gds requires innovus"
```

^ actually should it be something more specific?

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
harrisonliew: In the clarification, it needs to be `HammerToolHookAction` still because its `location` and `target_name` are accessed later

doc/Hammer-Basics/Hammer-Overview.rst
```diff
@@ -66,10 +70,15 @@ Using hooks requires the designer to extend the ``CLIDriver`` class. A good exam
...
+
 ``--obj_dir build`` designates what directory Hammer should use as a working directory.
 All default action run directories and output files will be placed here.
+
 Finally, ``par`` designates that this is a place-and-route action.
+
 In this case, Hammer will write outputs to the path ``$PWD/build/par-rundir``.
-For the full list of Hammer command-line arguments, look in the ``src/hammer-vlsi/hammer_vlsi/driver.py`` file.
+
+For the full list of Hammer command-line arguments, look in the ``src/hammer-vlsi/hammer_vlsi/cli_driver.py`` file.
```
edwardcwang:
```suggestion
For the full list of Hammer command-line arguments, run `hammer-vlsi --help` or take a peek in the ``src/hammer-vlsi/hammer_vlsi/cli_driver.py`` file.
```

doc/Hammer-Use/Flow-Control.rst
```diff
@@ -0,0 +1,34 @@
+.. _flow-control:
+
+Flow Control
+=======================================
+
+Physical design is necessarily an iterative process, and designers will often require fine control of the flow within any given action.
+This allows for rapid testing of new changes in the Hammer IR  to improve the quality of results after certain steps.
```
edwardcwang:
```suggestion
This allows for rapid testing of new changes in the Hammer IR to improve the quality of results (QoR) after certain steps.
```

doc/Hammer-Use/Flow-Control.rst
```diff
@@ -0,0 +1,34 @@
...
+Given the flow defined by tool steps and hooks (described in the next section), users can select ranges of these to run.
+The tool script will then be generated with commands corresponding to only the steps that are to be run.
+
+Command-line Interface
+----------------------
+
+Flow control is specified with the following optional Hammer command-line flags, which must always target a valid step/hook:
+
+* ``--from_step <target>``: this starts the tool from (inclusive) the target step
+* ``--after_step <target>``: this starts the tool from (exclusive) the target step
```
edwardcwang: Should this be maybe `start_after_step`, `post_step`, or `resume_after`? `after step` is ambiguous in that it's not clear whether it means to stop after or start after.

doc/Hammer-Use/Hooks.rst
```diff
@@ -1,31 +1,68 @@
...
+Hooks are fundamentally Python methods that extend a given tool's set of available steps and can inject additional TCL commands into the flow. 
+Hook methods need to take in an instance of a particular ``HammerTool``, which provides them with the full set of Hammer IR available to the tool.
+Hooks (depending on how they are included, see below) get turned into step objects that can be targeted with ``--from/after/to/until_step`` and other hooks.
 
-Hooks can live in a Python file inside the design root so that it is available to the class that needs to extend the default ``CLIDriver``. An example of some skeletons of hooks are found in `Chipyard <https://github.com/ucb-bar/chipyard/blob/master/vlsi/example-vlsi>`__.
+Hooks can live in a Python file inside the design root so that it is available to the class that needs to extend the default ``CLIDriver``.
+An example of some skeletons of hooks are found in `Chipyard <https://github.com/ucb-bar/chipyard/blob/master/vlsi/example-vlsi>`__.
```
edwardcwang: We can also link to some examples of unit tests which use hooks (which we hopefully have...)

src/hammer-vlsi/hammer_vlsi/cli_driver.py
```diff
@@ -396,13 +396,22 @@ def action(driver: HammerDriver, append_error_func: Callable[[str], None]) -> Op
                 pre_action_func(driver)
 
             # If the driver didn't successfully load, return None.
+            # Note the step/hook priority is (in order of lowest to highest):
+            # 1. Tool-supplied core steps
+            # 2. Tool-supplied hooks (usually for persistence)
+            # 3. Tech-supplied hooks
+            # 4. User-supplied hooks
+            assert driver.tech is not None, "must have a technology"
```
edwardcwang: Does this break `dump`?

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -400,47 +407,102 @@ def has_step(name: str) -> bool:
...
+                for i, nstep in enumerate(new_steps):
+                    if nstep.name == action.target_name:
+                        step_id = i
+                        break
+                for i, pstep in enumerate(persistent_steps):
+                    assert pstep.step is not None, "Persistent(Pre/Post)Step requires a step"
+                    if pstep.step.name == action.target_name:
+                        pstep_id = i
+                        break
+                assert (step_id > -1) != (pstep_id > -1)
```
edwardcwang: Suggest documenting this assertion like below
```python
                assert (step_id > -1) != (pstep_id > -1), "document this assertion here"
```

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -191,4 +193,40 @@ def scale_gds_script(self, gds_file: str) -> str:
...
+            HammerTool.make_post_insertion_hook("write_design", scale_final_gds)
+            ]}
+        return hooks.get(tool_name, [])
+
+def asap7_innovus_settings(ht: HammerTool) -> bool:
+    assert isinstance(ht, HammerPlaceAndRouteTool), "Innovus settings only for par"
+    assert isinstance(ht, CadenceTool), "Innovus is Cadence"
+    """Settings that need to be reapplied at every tool invocation"""
+    ht.append('''
+set_db route_design_bottom_routing_layer 2
```
edwardcwang: Are these settings not persistent?

Also, are these settings only "dropped" for this technology or is it the same across all technologies?

src/hammer-vlsi/test.py
```diff
@@ -1103,25 +1103,147 @@ def change3(x: hammer_vlsi.HammerTool) -> bool:
 
     def test_resume_pause_hooks_with_custom_steps(self) -> None:
         """Test that resume/pause hooks work with custom steps."""
+        def step5(x: hammer_vlsi.HammerTool) -> bool:
+            with open(os.path.join(c.temp_dir, "step5.txt"), "w") as f:
```
edwardcwang: To make these a bit easier to reason about could you move them under `with self.create_context() as c:`?

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
```diff
@@ -613,47 +697,58 @@ def make_resume_hook(step: str, location: HookLocation) -> HammerToolHookAction:
...
 
     @staticmethod
-    def make_from_to_hooks(from_step: Optional[str] = None,
-                           to_step: Optional[str] = None) -> List[HammerToolHookAction]:
-        """
-        Helper function to create a HammerToolHookAction list which will run from and to the given steps, inclusive.
-
-        :param from_step: Run from the given step, inclusive. Leave as None to resume from the beginning.
-        :param to_step: Run to the given step, inclusive. Leave as None to run to the end.
+    def make_start_stop_hooks(from_step: Optional[str] = None,
```
edwardcwang: While I agree that a generic form of the API is needed like this one, it would be good to have explicitly inclusive and exclusive variants e.g. `make_start_stop_hooks_inclusive` and `make_start_stop_hooks_exclusive`. This will help make it clearer as the generic API requires some double-taking to understand when reading.

src/hammer-vlsi/test.py
```diff
@@ -1103,25 +1103,147 @@ def change3(x: hammer_vlsi.HammerTool) -> bool:
...
+            for i in range(1, 7):
                 file = os.path.join(c.temp_dir, "step{}.txt".format(i))
                 if i == 5:
                     self.assertEqual(self.read(file), "HelloWorld")
                 else:
                     self.assertFalse(os.path.exists(file))
 
+    def test_persistent_hooks(self) -> None:
+        """Test that persistent hooks work."""
+        def persist(x: hammer_vlsi.HammerTool) -> bool:
```
edwardcwang: ditto

doc/Hammer-Use/Flow-Control.rst
harrisonliew: I don't see how `after_step` could be interpreted as stopping after, but is this documentation not clear about which flag is for what?

If we want to completely disambiguate it, we should use `resume_pre_step`, `resume_post_step`, `pause_pre_step`, and `pause_post_step` (could replace resume with start and pause with stop).

src/hammer-vlsi/technology/asap7/__init__.py
harrisonliew: To be honest, I am not certain what does and does not persist, and there's no Innovus documentation indicating what gets dropped and why. I also have no idea if it varies by technology without much trial-and-error. That said, it doesn't hurt to re-apply these settings; there are reference flows that will always apply general Innovus database settings at every tool invocation, so I opted to just be cautious here.

src/hammer-vlsi/test.py
harrisonliew: If I moved them under `with self.create_context() as c:` then I would have to duplicate the code... I don't see what's hard to reason about here.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
harrisonliew: I disagree, because the directives are meant to be mixed, e.g. a `--from_step` (inclusive) can be paired with an `--until_step` (exclusive). Thus, it doesn't make sense to separate it out.

Also, this shouldn't be a user-facing API anyway, as it's abstracted under the CLIDriver.

I'll attempt to improve the documentation instead.

src/hammer-vlsi/hammer_vlsi/cli_driver.py
harrisonliew: I don't see how it could. It doesn't affect how any of the config is structured under the hood.

src/hammer-vlsi/test.py
harrisonliew: ditto comment above, it would duplicate code.

src/hammer-vlsi/test.py
edwardcwang: It's easier to reason about in the sense that `c` appears like it's being used before defined (which it's actually not but had to do a double take here)

doc/Hammer-Use/Flow-Control.rst
edwardcwang: Just looking at it directly it can imply `[to-]after-step`

src/hammer-vlsi/technology/asap7/__init__.py
edwardcwang: What does a `get_db` return if resumed? I agree it's good to do it well, just want to make sure if we put in code we know why it's the case.

src/hammer-vlsi/hammer_vlsi/cli_driver.py
edwardcwang: Even if we are dumping configs without a technology?

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
edwardcwang: So I understand the need for the usecase of mixing them; it's just that code like `make_start_stop_hooks("step5", None, "step5", None)` could be clearer if it were `make_start_stop_hooks_inclusive("step5",  "step5")` since then an API user won't have to guess where the "None"s should go. Alternatively if the signature was something like `make_start_stop_hooks(Optional[tuple of string, bool], Optional[tuple of string, bool])` where the bool indicates exclusiveness. This will make sure that we don't specify two froms or two tos.

src/hammer-vlsi/hammer_vlsi/cli_driver.py
harrisonliew: Yes, I tested it without a specified technology and it defaults to `nop`. This is fine because the `driver`'s `__init__` calls `load_technology` which requires that the `tech is not None`.

src/hammer-vlsi/technology/asap7/__init__.py
harrisonliew: For these two particular settings, `get_db` does have them already, so persistence is not strictly necessary in this case. However, I think we can leave this for now, as it serves as an example for people building tech plugins.

src/hammer-vlsi/hammer_vlsi/hammer_tool.py
harrisonliew: I took the `NamedTuple` approach to clear this up. Inclusivity takes precedence and `--only_step` takes precedence over all.

doc/Hammer-Use/Flow-Control.rst
jwright6323: Recording an offline discussion with @colinschmidt :

What about having `--{start|stop}_{before|after}_step` + `--only_step` (which is synonymous with `--start_before_step x --stop_after_step x` as the documented 5 steps, and then leave behind the old names (`--to_step`, `--from_step`, etc.) as undocumented aliases for these core 5?

src/hammer-vlsi/technology/asap7/__init__.py
edwardcwang: I see. My thought is that if we don't need it it's better to try and keep the plugin lean...

harrisonliew: From offline discussion with @edwardcwang, this will remain persistent but it is documented that persistence is not truly necessary for these particular commands; it instead serves as an example for other tech plugins for how to implement persistence.

doc/Hammer-Basics/Hammer-Overview.rst
```diff
@@ -29,7 +29,11 @@ This demonstrates two different namespaces, ``vlsi.core`` and ``vlsi.inputs``, a
 Tech Plugins
 -------------------------------
 
-A techonology plugin consists of two or more files: a ``*.tech.json`` and a ``defaults.yml``.  The ``*.tech.json`` contains pointers to relevant PDK files and fundamental technology constants.  These values are not meant to be overriden, nor can they be. ``defaults.yml`` sets default technology variables for Hammer to consume, which may be specific to this technology or generic to all. These values may be overriden by design-specific configurations. An example of this is shown in ``hammer/src/hammer-vlsi/technology/asap7/``, and how to setup a technology plugin is documented in more detail in the Technology portion of the docs.
+A techonology plugin consists of two or more files: a ``*.tech.json`` and a ``defaults.yml``.
+
+The ``*.tech.json`` contains pointers to relevant PDK files and fundamental technology constants.  These values are not meant to be overriden, nor can they be.
```
edwardcwang:
```suggestion
The ``*.tech.json`` contains pointers to relevant PDK files and fundamental technology constants.  These values are not meant to be overriden, nor can they be for the time being. 
```

PR #544, opened 2019-11-12
Post par sim fix

dpgrubb13:
- Fix for post-par sim argument passing
- Addition of flag for testbench name for sim arguments
- Addition of execution flag wrappers


harrisonliew:
Though this looks fine to me, I'm wondering: could these prepend & append flags be implemented using `meta` prepend & append directives?

PR #545, opened 2019-11-15
ASAP7 SRAMs

harrisonliew:
Basic library of dummy SRAMs for ASAP7. It is designed for 151 projects.

All pins are on M4, with the signal all on the left side and the power stripes running across. The M5 power stripes are able to connect up.

All SRAMs are scaled up by 4x, so they are subject to the scaling script.

Tested with default Rocketchip config.


src/hammer-vlsi/technology/asap7/sram_compiler/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+        return "sram_generator.asap7"
+
+    def version_number(self, version: str) -> int:
+        return 0
+
+    # Run generator for a single sram and corner
+    def generate_sram(self, params: SRAMParameters, corner: MMMCCorner) -> ExtraLibrary:
+        tech_cache_dir = os.path.abspath(self.technology.cache_dir)
+        base_dir=os.path.join(self.tool_dir, "memories")
+        if(params.family == "1RW" or params.family == "2RW"):
```
edwardcwang:
```suggestion
        if params.family == "1RW" or params.family == "2RW":
```

src/hammer-vlsi/technology/asap7/sram-cache-gen.py
```diff
@@ -0,0 +1,112 @@
...
+    for line in f:
+        list_of_srams.append(line)
+
+json = []
+
+for sram_name in list_of_srams:
+    if sram_name.startswith("SRAM2RW"):
+        m = re.match("SRAM2RW(\d+)x(\d+)", sram_name)
+        if m:
+            json.append("""
```
edwardcwang: To file as issue: this should use the Hammer IR library

src/hammer-vlsi/technology/asap7/sram_compiler/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+            fam_code = params.family
+        else:
+            self.logger.error("ASAP7 SRAM cache does not support family:{f}".format(f=params.family))
+        #TODO: this is really an abuse of the corner stuff
+        #TODO: when we have matching corners remove real_corner and symlink junk
+        if(corner.type == MMMCCornerType.Setup):
```
edwardcwang:
```suggestion
        if corner.type == MMMCCornerType.Setup:
```

src/hammer-vlsi/technology/asap7/sram_compiler/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+            speed_name = "slow"
+        elif(corner.type == MMMCCornerType.Hold):
```
edwardcwang:
```suggestion
        elif corner.type == MMMCCornerType.Hold:
```

src/hammer-vlsi/technology/asap7/sram_compiler/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+            speed_name = "fast"
+        elif(corner.type == MMMCCornerType.Extra):
```
edwardcwang:
```suggestion
        elif corner.type == MMMCCornerType.Extra:
```

src/hammer-vlsi/technology/asap7/sram-cache-gen.py
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env python3
+
```
edwardcwang: Can we type check this file with `from typing import ...`?

harrisonliew:
1. These memories are completely blank (full obstructions on layers M1-M3).

2. The `sram_compiler/__init__.py` is just used to extract which collateral is needed for the design. MacroCompiler uses `sram-cache.json` to select the SRAMs in the first place, and that json is generated by `sram-cache-gen.py` (which I don't believe needs to be type-checked... there are no functions and it doesn't interact with Hammer at all).

3. The `SRAMGeneratorTool` is already unit-tested (with `nop` tech). Do you want the test to be replaced with this technology?

edwardcwang:
1) LGTM

2) "sram_compiler/__init__.py is just used to extract which collateral is needed for the design" <-- I don't understand; isn't the cache of memories pre-generated/fixed? Also it'd be good to keep the entire codebase type-checked even if it's a helper script

3) Why not both?

sehuang:
This looks fine to me, just two clarifying questions for my sake:
* Is SRAMParameters adapted from the `sram-cache.json`? I'm assuming this is called by MacroCompiler once it's decided on which SRAM to use so it creates the SRAMParameters by slicing out the SRAM from the `sram-cache.json`?
* Is it regenerating the `sram-cache.json` each time or was this a one-time script to generate the cache?

harrisonliew:
2. The cache of memories is fixed (once, @sehuang) by `srams.txt` as an argument to `sram-cache-gen.py` (I've type-checked). `sram_compiler/__init__.py` is only used to generate the `ExtraLibrary`s needed by the particular design.

3. Sure, I've unit-tested.

@sehuang you are mostly correct about the SRAMParameters. It's technically generated from the json that MacroCompiler spits out. MacroCompiler consumes `sram-cache.json`.

PR #552, opened 2019-12-11
Add ability to hack installs into cache dir

harrisonliew:
Motivated by new use case after implementing ability to hack PDK tarballs (e.g. ASAP7).

This allows a tech plugin that accesses existing PDK installs to implement a `post_install_script` which could copy some file and hack it into the tech cache dir. Then, the tech JSON can point to this hacked file such as follows:

```
{
  "installs": [
    {
      "path": "tech-asap7-cache",
      "base var": ""
    }
  ],
...
```
The `path` must be the basename of the tech cache directory. In most cases, it is `tech-<tech_name>-cache`. `base var` must be left as an empty string.

### DIFF
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -672,7 +672,7 @@ class HammerTechnology:
             if len(matching_installs) == 1:
                 install = matching_installs[0]
                 if install.base_var == "":
-                    if "cache" in install.path.lower():
+                    if install.path == os.path.basename(self.cache_dir): # default is tech-<techname>-cache
                         base = self.cache_dir
                     else:
                         base = self.path
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -672,7 +672,10 @@ class HammerTechnology:
             if len(matching_installs) == 1:
                 install = matching_installs[0]
                 if install.base_var == "":
-                    base = self.path
+                    if "cache" in install.path.lower():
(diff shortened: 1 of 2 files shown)


colinschmidt:
The example you gave here doesn't use `cache` but `$CACHE`. Thats a bit confusing.
Also can we restrict this replacement. `cache` is a relatively common word in computer systems, so I would be worried about paths like:
`/path/to/my/project/design/generators/cache/cache.gds`
If we allow replacement anywhere in the variable.

src/hammer-tech/hammer_tech.py
```diff
@@ -672,7 +672,10 @@ def prepend_dir_path(self, path: str, lib: Optional[Library] = None) -> str:
             if len(matching_installs) == 1:
                 install = matching_installs[0]
                 if install.base_var == "":
-                    base = self.path
+                    if "cache" in install.path.lower():
```
edwardcwang: In addition to what Colin said, perhaps instead of matching any lowercase "cache", maybe restrict this to a specific install.path?
```suggestion
                    if install.path == "tech_cache":
```
or something

harrisonliew: Yeah, I think I agree with you two here. I still think that having the path be "tech_cache" is a bit unsatisfactory because it has to be such a specific string, but I can't think of a different way of doing it...

harrisonliew:
I updated this to check against the tech cache string directly and updated the original description accordingly. This is probably disambiguous enough?

Issue #557, opened 2020-01-06
Github is unable to detect your license file


mithro:
Your LICENSE file is not detected by GitHub's license detection system. This means you won't be automatically included in various lists and scanned by various tools.

You can find out more about how the LICENSE detection at https://help.github.com/en/github/creating-cloning-and-archiving-repositories/licensing-a-repository#detecting-a-license

edwardcwang:
Thanks for your interest in Hammer! The licence is a 3-clause BSD, so not sure why the server-side library (licensee) used by GitHub doesn't detect it. If you have any additional questions please feel free to be in touch.

mithro:
@edwardcwang You probably want to figure out why the 3-clause BSD isn't being detected as lots of people use that information as authoritative.

PR #558, opened 2020-01-15
Move nonstandard language out of license to readme.

colinschmidt:
Hopefully this will allow github's auto-detection feature correctly
identify the 3-clause BSD license we use. Fixing #557

I made this change after comparing our license file to the text here: https://choosealicense.com/licenses/bsd-3-clause/

I'm not sure if we can test this on a branch or not. I'm also unsure how often the auto-detection is run.


edwardcwang:
Just FYI AFAIK the Chisel/FIRRTL licenses also aren't autodetected so I think it probably doesn't matter too much...

PR #565, opened 2020-02-24
Add generic top_layer setting for blockage spacing.

colinschmidt:
Enables designs without floorplan information to still have blockage around macros.

### DIFF
--- a/src/hammer-vlsi/defaults.yml
+++ b/src/hammer-vlsi/defaults.yml
@@ -563,7 +563,7 @@ par:
   # Top metal layer that is pulled back around blocks and hardmacros
   # Used by floorplanning.
   # Overridden by individual placement constraints top_layer.
-  # type: Optional[String]
+  # type: Optional[str]
   blockage_spacing_top_layer: null
 
   # If power_straps_mode is 'generate', which method to use.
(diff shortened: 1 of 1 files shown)


colinschmidt:
It's just the `top_layer` field of `placement_constraint`

jwright6323:
Got it, and it is just specifying the top layer used by that block? Is there any reason not to reuse that for the blockage (or at least as a sane default if it is unspecified)?

colinschmidt:
I think it is reasonable and perhaps even implied by the comment that it is what should be used.

jwright6323:
Oh, I get it now. This is the global default and the others override.

src/hammer-vlsi/defaults.yml
```diff
@@ -560,6 +560,12 @@ par:
   # type: Decimal
   blockage_spacing: 0.0
 
+  # Top metal layer that is pulled back around blocks and hardmacros
+  # Used by floorplanning.
+  # Overridden by individual placement constraints top_layer.
+  # type: Optional[String]
```
```suggestion
  # type: Optional[str]
```

harrisonliew:
I'm a bit confused about this - you specify a layer to pull back globally across all blocks and hardmacros? Is the intention that doing when evoking the tool's auto-floorplanning utility, you don't have any `placement_constraints` and just want to see the impact of blocking off layers up to the specified one?

Issue #315, opened 2019-02-22
Some tech files should live in the "Stackup" object


jwright6323:
For example, QRC techfiles and technology LEFs are stackup-specific, so these should live in the "stackup" data structure. Right now, if you have multiple stackups, you'd need separate technology JSONs to support that. This is low priority since right now we don't support any technologies with multiple stackups, but we should fix it before we build too much infrastructure with the current APIs.

jwright6323:
Also, obligatory comment that this relies on #308

jwright6323:
Bumping the priority here, as it's an annoyance for building new plugins. The QRC techfile should definitely not be a library-specific thing.

jwright6323:
Also, funnily enough, the tech LEF is an independent library and one could argue it belongs with the std cells, although I think it's also reasonable to keep it separate.

PR #566, opened 2020-03-06
Cap table support

ErikFanderson:
This allows for specifying a "cap table file" in the tech.json for older technologies that do not come with a qrc techfile. There should not be any issue with supplying both a cap table file and qrc tech file. The tool should always prioritize the qrc tech file. I also edited the function "get_and_prepend_path" inside the "process_library_filter" function so that if an empty path is retrieved from the tech.json then an empty string will be returned. This behavior appears to have already been expected in the "get_mmmc_qrc" function.

### DIFF
--- a/src/hammer-tech/hammer_tech.py
+++ b/src/hammer-tech/hammer_tech.py
@@ -811,7 +811,7 @@ class HammerTechnology:
         # Next, extract paths and prepend them to get the real paths.
         def get_and_prepend_path(lib: Library) -> Tuple[Library, List[str]]:
             paths = filt.paths_func(lib)
-            full_paths = list(map(lambda path: self.prepend_dir_path(path, lib), paths))
+            full_paths = list(map(lambda path: self.prepend_dir_path(path, lib) if path else '', paths))
             return lib, full_paths
 
         libs_and_paths = list(map(get_and_prepend_path, filtered_libs))  # type: List[Tuple[Library, List[str]]]
--- a/src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
+++ b/src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
@@ -1473,7 +1473,16 @@ class CadenceTool(HasSDCSupport, HasCPFSupport, HasUPFSupport, TCLTool, HammerTo
         lib_args = self.technology.read_libs([hammer_tech.filters.qrc_tech_filter],
                                              hammer_tech.HammerTechnologyUtils.to_plain_item,
                                              extra_pre_filters=[
-                                                 self.filter_for_mmmc(voltage=corner.voltage, temp=corner.temp)])
+                                             self.filter_for_mmmc(voltage=corner.voltage, temp=corner.temp)],
+                                             must_exist=False)
+        return " ".join(lib_args)
+
+    def get_mmmc_cap_table(self, corner: MMMCCorner) -> str:
+        lib_args = self.technology.read_libs([hammer_tech.filters.cap_table_filter],
+                                             hammer_tech.HammerTechnologyUtils.to_plain_item,
+                                             extra_pre_filters=[
+                                             self.filter_for_mmmc(voltage=corner.voltage, temp=corner.temp)],
+                                             must_exist=False)
         return " ".join(lib_args)
 
     def get_qrc_tech(self) -> str:
@@ -1484,6 +1493,17 @@ class CadenceTool(HasSDCSupport, HasCPFSupport, HasUPFSupport, TCLTool, HammerTo
         """
         lib_args = self.technology.read_libs([
             hammer_tech.filters.qrc_tech_filter
+        ], hammer_tech.HammerTechnologyUtils.to_plain_item)
+        return " ".join(lib_args)
+
+    def get_cap_table(self) -> str:
+        """
+        Helper function to get the list of cap table files in space separated format.
+
+        :return: List of cap table files separated by spaces
+        """
+        lib_args = self.technology.read_libs([
+            hammer_tech.filters.cap_table_filter
         ], hammer_tech.HammerTechnologyUtils.to_plain_item)
         return " ".join(lib_args)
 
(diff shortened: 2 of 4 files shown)


edwardcwang:
Hi Erik, thank you for your contributions to Hammer! :)

A few things:

1) Just to clarify, are we referring to this format or another format for the cap table? https://github.com/The-OpenROAD-Project/PEX/blob/master/kits/nangate45/captables/NCSU_FreePDK_45nm.capTbl

2) Could you add a unit tests for this filter, following the format of the other unit tests for similar features?

3) It appears that the type checker is failing.

Thanks again for your contribution!

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -1565,16 +1585,18 @@ def append_mmmc(cmd: str) -> None:
...
                 list="{n}.hold_set".format(n=hold_corner.name)
             ))
-            # Next, create Innovus rc corners from qrc tech files
-            append_mmmc("create_rc_corner -name {name} -temperature {tempInCelsius} {qrc}".format(
+            # Next, create Innovus rc corners from qrc tech or cap table files 
+            append_mmmc("create_rc_corner -name {name} -temperature {tempInCelsius} {qrc} {cap_table}".format(
                 name="{n}.setup_rc".format(n=setup_corner.name),
                 tempInCelsius=str(setup_corner.temp.value),
-                qrc="-qrc_tech {}".format(self.get_mmmc_qrc(setup_corner)) if self.get_mmmc_qrc(setup_corner) != '' else ''
+                qrc="-qrc_tech {}".format(self.get_mmmc_qrc(setup_corner)) if self.get_mmmc_qrc(setup_corner) != '' else '',
```
harrisonliew: I'm sure you tested this, but does `-qrc_tech` and `-cap_table` options need to be exclusive? Also, the if/else on these lines seem redundant.

src/hammer-tech/hammer_tech.py
```diff
@@ -811,7 +811,7 @@ def process_library_filter(self,
         # Next, extract paths and prepend them to get the real paths.
         def get_and_prepend_path(lib: Library) -> Tuple[Library, List[str]]:
             paths = filt.paths_func(lib)
-            full_paths = list(map(lambda path: self.prepend_dir_path(path, lib), paths))
+            full_paths = list(map(lambda path: self.prepend_dir_path(path, lib) if path else '', paths))
```
colinschmidt: Why did this change? Is this just a bugfix?

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -1473,7 +1473,16 @@ def get_mmmc_qrc(self, corner: MMMCCorner) -> str:
         lib_args = self.technology.read_libs([hammer_tech.filters.qrc_tech_filter],
                                              hammer_tech.HammerTechnologyUtils.to_plain_item,
                                              extra_pre_filters=[
-                                                 self.filter_for_mmmc(voltage=corner.voltage, temp=corner.temp)])
+                                             self.filter_for_mmmc(voltage=corner.voltage, temp=corner.temp)],
```
colinschmidt: I think you deleted an indent here.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -1473,7 +1473,16 @@ def get_mmmc_qrc(self, corner: MMMCCorner) -> str:
...
+                                             must_exist=False)
+        return " ".join(lib_args)
+
+    def get_mmmc_cap_table(self, corner: MMMCCorner) -> str:
+        lib_args = self.technology.read_libs([hammer_tech.filters.cap_table_filter],
+                                             hammer_tech.HammerTechnologyUtils.to_plain_item,
+                                             extra_pre_filters=[
+                                             self.filter_for_mmmc(voltage=corner.voltage, temp=corner.temp)],
```
colinschmidt: Another indent.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -1565,16 +1585,18 @@ def append_mmmc(cmd: str) -> None:
                 name="{n}.hold_cond".format(n=hold_corner.name),
                 list="{n}.hold_set".format(n=hold_corner.name)
             ))
-            # Next, create Innovus rc corners from qrc tech files
-            append_mmmc("create_rc_corner -name {name} -temperature {tempInCelsius} {qrc}".format(
+            # Next, create Innovus rc corners from qrc tech or cap table files
```
colinschmidt: Is there a space at the end of the line here? If so please delete.

src/hammer-tech/hammer_tech.py
ErikFanderson: prepend_dir_path has an assertion  that checks to make sure the path is not empty. I am not entirely sure why the assertion exists, but this was my solution to get an empty list to be returned

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
ErikFanderson: They do not need to be exclusive. The tool will gladly accept both and will simply use the qrc tech file.

ErikFanderson:
> Hi Erik, thank you for your contributions to Hammer! :)
> 
> A few things:
> 
>     1. Just to clarify, are we referring to this format or another format for the cap table? https://github.com/The-OpenROAD-Project/PEX/blob/master/kits/nangate45/captables/NCSU_FreePDK_45nm.capTbl
> 
>     2. Could you add a unit tests for this filter, following the format of the other unit tests for similar features?
> 
>     3. It appears that the type checker is failing.
> 
> 
> Thanks again for your contribution!

1. Yes, that is the format I am referring to.
2. Yes, I will definitely do that. Please be patient with me as I sort through the testing infrastructure.
3. Yup, I will try to sort that out asap.

edwardcwang:
Great! Might be good to add a link to https://github.com/The-OpenROAD-Project/PEX/blob/master/kits/nangate45/captables/NCSU_FreePDK_45nm.capTbl in the yaml comments/documentation for reference.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
ErikFanderson: If/else either return "-qrc_tech {file}" or an empty string if the {file} is an empty string. So unless I'm missing something I don't believe it's redundant

ErikFanderson:
Update on this: Added a test for the filter in tech_test.py and also added a property to the Library stub class for type checking purposes. Not entirely sure why only some properties from the jsonschema generated Library class need to be defined in the stub class, but mypy no longer errors.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
harrisonliew: I see now!

src/hammer-tech/hammer_tech.py
edwardcwang: Hmm, how do empty paths end up showing up to begin with? And how is this cap table specific?

ErikFanderson: Well because you only need a cap_table_file OR a qrc_techfile then 1 of them can be empty. Previously hammer would error if you did not supply a qrc_techfile even though the function "generate_mmmc_script" already assumed that if there was no supplied qrc_techfile that an empty string would be retrieved when calling "get_mmmc_wrc". I guess this raises the question of whether we actually want an error to be thrown if neither of these files are supplied

harrisonliew: Hm, seeing as how other things use `prepend_dir_path` I also don't see what the value of the `len(path) > 0` assertion is, seeing as it is only used in certain cases (DRC/LVS decks, map files) but doesn't make whatever you specified in the tech json any more or less correct.

Seems like what we really should do is call an `os.path.exists` on the constructed path so that Python can catch bad paths in tech jsons before it's passed to the tool where sometimes at best only a warning is thrown.

edwardcwang:
Can we also add `https://github.com/The-OpenROAD-Project/PEX/blob/master/kits/nangate45/captables/NCSU_FreePDK_45nm.capTbl` to a pydoc or something to keep track of the format?

src/hammer-tech/hammer_tech.py
edwardcwang: If there are no qrc techfiles, then wouldn't `get_mmmc_qrc()` return an empty string since the `read_libs` call would return an empty list?

jwright6323: As a side note, we really need to detach the QRC techfile from the libraries (#315).

jwright6323: Maybe we can incorporate that larger change into this PR, so that we aren't creating more work we need to undo/fix later?

PR #572, opened 2020-05-20
Power and Sim tool support and improvements

colinschmidt:
This PR has become a little unwieldy as I continued to push small fixes to @dpgrubb13 's original work getting the power analysis tool infrastructure in.

I'm not being a very good citizen here but I think it would be nice if I didn't have to disentangle these commits from the original work, although that is obviously the "right" solution.

On topic:
This PR adds support for a power analysis tool to hammer. This tool can be run after par and sim-par. It will generate static and dynamic power reports. The dynamic power reports can be based on a VCD waveform or an SAIF activity file.
The sim commits in this PR add the ability to generate SAIF files for use in this power flow.
Both of these tools are also integrated into the `build_system: 'make'` system.

I've been using this branch along with ucb-bar/chipyard#512 to run post-par dynamic power analysis with the trigger saif mode.


src/hammer-vlsi/defaults.yml
```diff
@@ -560,6 +568,12 @@ par:
   # type: Decimal
   blockage_spacing: 0.0
 
+  # Top metal layer that is pulled back around blocks and hardmacros
+  # Used by floorplanning.
+  # Overridden by individual placement constraints top_layer.
```
edwardcwang, line 573:
```suggestion
  # Overridden by individual placement constraints on top_layer.
```

src/hammer-vlsi/defaults.yml
```diff
@@ -560,6 +568,12 @@ par:
...
+  # type: Optional[String]
```
edwardcwang, line 574:
```suggestion
  # type: Optional[str]
```

src/hammer-vlsi/defaults.yml
```diff
@@ -794,6 +808,28 @@ sim.inputs:
...
+    #   after enabling dumping before stopping dumping. See your simulator plugin for an example
+    # none - Start dumping at the beginning of simulation and stop at the end.
+    # type: str
+    mode: "none"
+    # The times the simulator should begin and end its dumping of the SAIF file.
+    # type: TimeValue
+    start_time: "0ns"
+    # type: TimeValue
+    end_time: "1ns"
+    # The triggers that will occur when the dump should start and when the dump should end.
```
edwardcwang, line 828: What are some examples of this setting?

src/hammer-vlsi/defaults.yml
```diff
@@ -817,6 +853,29 @@ sim.submit:
   settings: "vlsi.submit.settings"
   settings_meta: lazycrossref
 
+power.inputs:
```
edwardcwang, line 856: Could we add documentation for all these settings?

src/hammer-vlsi/defaults.yml
```diff
@@ -794,6 +808,28 @@ sim.inputs:
   # If set to True, the sim will be time annotated based on synthesis or PAR results.
   timing_annotated: false
 
+  # Settings related to activity file (SAIF) dumping
+  saif:
+    # Enable dumping an SAIF file, with activity factors
```
edwardcwang, line 813:
```suggestion
    # Enable dumping an SAIF file, with activity factors
    # type: bool
```

src/hammer-vlsi/hammer_vlsi/driver.py
```diff
@@ -528,6 +530,58 @@ def load_sim_tool(self, run_dir: str = "") -> bool:
...
+        power_tool.set_database(self.database)
+        power_tool.submit_command = HammerSubmitCommand.get("power", self.database)
+        power_tool.run_dir = run_dir
+
+        #power_tool.top_module = self.database.get_setting("power.inputs.top_module", nullvalue="")
+        #power_tool.netlist = self.database.get_setting("power.inputs.netlist", nullvalue="")
+        #power_tool.spef_file = self.database.get_setting("power.inputs.spef_file", nullvalue="")
+        power_tool.waveforms = self.database.get_setting("power.inputs.waveforms", nullvalue=[])
+        missing_inputs = False
+        #if power_tool.top_module == "":
```
edwardcwang, line 567: Is there a reason these are commented out?

src/hammer-vlsi/hammer_vlsi/driver.py
```diff
@@ -528,6 +530,58 @@ def load_sim_tool(self, run_dir: str = "") -> bool:
...
+        assert isinstance(power_tool_get, HammerPowerTool), "Power tool must be a HammerPowerTool"
+        power_tool = power_tool_get  # type: HammerLVSTool
+        power_tool.name = power_tool_name
+        power_tool.logger = self.log.context("power")
+        power_tool.technology = self.tech
+        power_tool.set_database(self.database)
+        power_tool.submit_command = HammerSubmitCommand.get("power", self.database)
+        power_tool.run_dir = run_dir
+
+        #power_tool.top_module = self.database.get_setting("power.inputs.top_module", nullvalue="")
```
edwardcwang, line 562: Is there a reason these are commented out?

src/hammer-vlsi/defaults.yml
colinschmidt: I think we should have a general discussion about this setting to see what level the API should be at.
It is currently just the raw commands and so I think the documentation would have to be in the tool plugins, but I did a little research into what we might be able to support on commercial simulators so we potentially could have a higher level API.

src/hammer-vlsi/defaults.yml
colinschmidt: Done

src/hammer-vlsi/defaults.yml
colinschmidt: Updated.

src/hammer-vlsi/defaults.yml
colinschmidt: Done.

src/hammer-vlsi/hammer_vlsi/driver.py
colinschmidt: Fixed

src/hammer-vlsi/defaults.yml

src/hammer-vlsi/defaults.yml

src/hammer-vlsi/hammer_vlsi/hammer_build_systems.py
```diff
@@ -36,12 +36,18 @@ def build_makefile(driver: HammerDriver, append_error_func: Callable[[str], None
         - par
         - drc
         - lvs
+        - sim
```
jwright6323, line 39: should these sim actions be split into `sim-par` and `sim-syn`?

src/hammer-vlsi/hammer_vlsi/hammer_build_systems.py
```diff
@@ -118,51 +127,92 @@ def build_makefile(driver: HammerDriver, append_error_func: Callable[[str], None
...
+        sim-par{suffix}: {sim_par_out}
+        power{suffix}: {power_out}
         drc{suffix}: {drc_out}
         lvs{suffix}: {lvs_out}
 
         {par_to_syn}
-        {syn_out}: {syn_deps}
-        \t$(HAMMER_EXEC) {env_confs} {p_syn_in} --obj_dir {obj_dir} syn{suffix}
+
+        {sim_out}: {syn_deps} $(HAMMER_SIM_DEPENDENCIES)
```
jwright6323, line 142: is this doing RTL simulation?

src/hammer-vlsi/defaults.yml
```diff
@@ -819,6 +827,30 @@ sim.inputs:
...
+    # trigger_raw - Inserts the given tcl directly into the simulators run script with no checking.
+    # none - Start dumping at the beginning of simulation and stop at the end.
+    # type: str
+    mode: "none"
+    # The times the simulator should begin and end its dumping of the SAIF file.
+    # type: TimeValue
+    start_time: "0ns"
+    # type: TimeValue
+    end_time: "1ns"
+    # The raw triggers that will occur when the dump should start and when the dump should end.
```
edwardcwang, line 849:
```suggestion
    # The raw triggers that will occur when the dump should start and when the dump should end.
    # The exact trigger format will depend on the underlying tool.
```

src/hammer-vlsi/defaults.yml
```diff
@@ -842,6 +874,44 @@ sim.submit:
...
+  tb_name: "TestDriver"
+
+  # The name of the device under test (DUT) from the simulation.
+  # type: str
+  tb_dut: "dut"
+
+  # The paths of all spef (extraction) files needed for the design.
+  # This may include a file per MMMC corner
+  # type: List[str]
+  spefs: null
```
edwardcwang, line 899: Should this be an empty list?
```suggestion
  spefs: []
```

src/hammer-vlsi/defaults.yml
```diff
@@ -819,6 +827,30 @@ sim.inputs:
   # If set to True, the sim will be time annotated based on synthesis or PAR results.
   timing_annotated: false
 
+  # Settings related to activity file (SAIF) dumping
+  saif:
+    # Enable dumping an SAIF file, with activity factors
+    # type: bool
+    enable: false
```
harrisonliew, line 834: Definitely more of a nit, but based on the logic in the companion PR, it appears to me that you could combine this `enable` flag with the `mode` key a certain way to maybe be less confusing to the user. Perhaps `none` below should be `full` and `none` or anything else that's not a valid option will not dump the SAIF file.

src/hammer-vlsi/defaults.yml
```diff
@@ -842,6 +874,44 @@ sim.submit:
...
+# Specific inputs for the power tool.
+# These may vary based on the specific tool plugin
+power.inputs:
+  # The path to a place and route database of the design to be analyzed
+  # type: str
+  database: null
+
+  # Enable static power rail analysis
+  # type: bool
+  static_rail: false
```
harrisonliew, line 886: Similar to above. Seems like the more consistent thing to do would be to remove the `static_rail` step if you don't want static rail analysis to be run rather than making this flag false.

src/hammer-vlsi/hammer_vlsi/cli_driver.py
```diff
@@ -349,6 +367,14 @@ def create_sim_action(self, custom_hooks: List[HammerToolHookAction],
         return self.create_action("sim", hooks if len(hooks) > 0 else None,
                                   pre_action_func, post_load_func, post_run_func)
 
+    def create_power_action(self, custom_hooks: List[HammerToolHookAction],
+                          pre_action_func: Optional[Callable[[HammerDriver], None]] = None,
+                          post_load_func: Optional[Callable[[HammerDriver], None]] = None,
+                          post_run_func: Optional[Callable[[HammerDriver], None]] = None) -> CLIActionConfigType:
+        hooks = self.get_extra_sim_hooks() + custom_hooks  # type: List[HammerToolHookAction]
```
harrisonliew, line 374: should this be `get_extra_power_hooks`?

src/hammer-vlsi/defaults.yml
colinschmidt: Yeah that seems reasonable and matches more of what the other APIs look like.
I think this weirdness stems from the evolution of the API as this PR grew in size.
Will fix soon.

src/hammer-vlsi/defaults.yml
colinschmidt: Yeah will do another pass on the flags.

src/hammer-vlsi/hammer_vlsi/cli_driver.py
colinschmidt: Yup. Fixed.

src/hammer-vlsi/hammer_vlsi/hammer_build_systems.py
colinschmidt: I think we should label them since its more important that there are targets for these althouh technically there is only one `sim` action, and its just called with different inputs.

src/hammer-vlsi/hammer_vlsi/hammer_build_systems.py
colinschmidt: yes this is RTL sim

src/hammer-vlsi/defaults.yml
colinschmidt: Fixed.

src/hammer-vlsi/defaults.yml
colinschmidt: Fixed.

src/hammer-vlsi/defaults.yml
colinschmidt: Fixed.

src/hammer-vlsi/defaults.yml

src/hammer-vlsi/hammer_vlsi/hammer_build_systems.py
colinschmidt: Fixed.

src/hammer-vlsi/hammer_vlsi/hammer_build_systems.py
colinschmidt: Fixed

PR #576, opened 2020-06-03
Refactor simulation reg and cell lists and cadence tool location

### LINKED ISSUE #574: Move all registers list for simulations into separate json
Rather than inlining this field, since it is often very large, we should have it be a separate file and the current key become a path to that file.
This should speed up hammer's launch when these json fragments are included.

colinschmidt:
This fixes #574 and shows an example of how we might solve #198 
I ended up wanting to put common TCL code into the CadenceTool class to fix the first issue so I decided to bite the bullet and try to make this change which lets this kind of code be added more safely.


src/hammer-vlsi/defaults.yml
```diff
@@ -802,13 +802,13 @@ sim.inputs:
   # type: SimulationLevel
   level: "rtl"
 
-  # List of all registers in the design passed from synthesis or place-and-route
+  # Path to a list of all registers in the design passed from synthesis or place-and-route
   # type: List[str]
```
edwardcwang, line 806:
```suggestion
  # type: str
```

edwardcwang:
Also asap7 may have a dependency on CadenceTool
```
Running mypy ../hammer-vlsi/technology/asap7/sram_compiler/__init__.py
../hammer-vlsi/technology/asap7/__init__.py:16: error: Module 'hammer_vlsi' has no attribute 'CadenceTool'
```

colinschmidt:
The ASAP7 dependency is worth discussing. We could move these snippets into the cadence plugin or elide the type check. The type check seems to only be there to avoid users including these hooks in other flows. The code in hammer is safe without the type check because the `get_tech_par_hooks` is already only adding them to cadence tools.
Usually bundling tool hooks with technologies works fine in practice because its much harder to get access to technologies than tools, however there is an anomaly here.
I'm interested to see what others have to say.

src/hammer-vlsi/defaults.yml

harrisonliew:
My $0.02 is that this refactoring of `CadenceTool` is fine as is, since `TCLTool` is all that seems necessary to get the hooks to write to the TCL scripts? Documentation already implies that ASAP7 must use Cadence tools, so I don't think we need to explicitly check for `CadenceTool` here.

Otherwise, rest of the PR LGTM.

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -202,7 +202,8 @@ def get_tech_par_hooks(self, tool_name: str) -> List[HammerToolHookAction]:
 
 def asap7_innovus_settings(ht: HammerTool) -> bool:
     assert isinstance(ht, HammerPlaceAndRouteTool), "Innovus settings only for par"
-    assert isinstance(ht, CadenceTool), "Innovus is Cadence"
+    assert isinstance(ht, TCLTool), "innovus settings can only run on TCL tools"
+    #assert isinstance(ht, CadenceTool), "Innovus is Cadence"
```
harrisonliew, line 206: Let's remove this entirely for clarity?

colinschmidt: Yeah makes sense.

harrisonliew:
Also, as a parting comment I find the importing/type checking of the common `CadenceTool` code to be somewhat clunky, but I can't think of a better way to do it.

colinschmidt:
Yeah I was surprised mypy wasn't able to just automatically pick it up, perhaps there is a newer version that does it, or we could maybe include some mypy code inside the users of CadenceTool

PR #577, opened 2020-06-06
Add gds_precision_mode and gds_precision default options

growly:
These values will be used by PAR tools to optionally specify the precision of the GDS files they output. Implementation for Innovus is in ucb-bar/hammer-cadence-plugins#49.


src/hammer-vlsi/defaults.yml
```diff
@@ -565,6 +565,18 @@ par.inputs:
...
   physical_only_cells_list: []
 
+  # GDS output precision. Controlled by a mode and a precision value. Valid modes are:
+  # auto - perform default PAR-tool behaviour, which is usually to use the
+  #        value specified in the LEF.
+  # manual - use the value from the 'par.inputs.gds_precision' key below.
+  gds_precision_mode: "auto"
+
+  # GDS output precision when written by PAR tool. Valid values depend on the
+  # specific PAR tool in use. Innovus, for example, restricts it to {100, 200,
```
jwright6323, line 575: See my comment in the companion PR, but I don't think we need to restrict it. It's fair to keep a comment saying these are likely the only sane values, though.

src/hammer-vlsi/defaults.yml
```diff
@@ -565,6 +565,18 @@ par.inputs:
...
+  # 1000, 2000, 10000, 20000}. This value is ignored if
+  # 'par.inputs.gds_precision_mode' is not manual.
+  gds_precision: 1000
```
edwardcwang, line 578:
```suggestion
  # type: int
  gds_precision: 1000
```

src/hammer-vlsi/defaults.yml
```diff
@@ -565,6 +565,18 @@ par.inputs:
   # type List[str]
   physical_only_cells_list: []
 
+  # GDS output precision. Controlled by a mode and a precision value. Valid modes are:
+  # auto - perform default PAR-tool behaviour, which is usually to use the
+  #        value specified in the LEF.
+  # manual - use the value from the 'par.inputs.gds_precision' key below.
+  gds_precision_mode: "auto"
```
edwardcwang, line 572:
```suggestion
  # type: str
  gds_precision_mode: "auto"
```

src/hammer-vlsi/defaults.yml
colinschmidt: For the second sentence here I think we should just say something like:
`Some tools restrict the value to a limited range and your tool plugin should check this for you`

jwright6323: As discussed I'm ok with leaving this in

PR #583, opened 2020-06-10
Make sourceme.sh set HAMMER_HOME to its root dir by default

growly:
If no ${HAMMER_HOME} is specified, the script finds the location of itself (even if called through symlinks) and uses that as the the default.

Users who wish to set this manually can do so prior to invoking the script, as was the previous behaviour.

### DIFF
--- a/sourceme.sh
+++ b/sourceme.sh
@@ -6,9 +6,9 @@ if [ -z "${HAMMER_HOME}" ]; then
   #  https://stackoverflow.com/questions/59895/how-to-get-the-source-directory-of-a-bash-script-from-within-the-script-itself
   #  https://stackoverflow.com/questions/35006457/choosing-between-0-and-bash-source
   export HAMMER_HOME="$( dirname "$( readlink -f "${BASH_SOURCE[0]}" )" )"
-  >&2 echo "Using (default) HAMMER_HOME=${HAMMER_HOME}"
+  >&2 echo "Using (guessed) HAMMER_HOME=${HAMMER_HOME}"
 else
-  >&2 echo "Using HAMMER_HOME=${HAMMER_HOME}"
+  >&2 echo "Using (existing) HAMMER_HOME=${HAMMER_HOME}"
 fi
 
 export HAMMER_VLSI="${HAMMER_HOME}/src/hammer-vlsi"
--- a/sourceme.sh
+++ b/sourceme.sh
@@ -6,9 +6,7 @@ if [ -z "${HAMMER_HOME}" ]; then
   #  https://stackoverflow.com/questions/59895/how-to-get-the-source-directory-of-a-bash-script-from-within-the-script-itself
   #  https://stackoverflow.com/questions/35006457/choosing-between-0-and-bash-source
   export HAMMER_HOME="$( dirname "$( readlink -f "${BASH_SOURCE[0]}" )" )"
-  >&2 echo "Using (guessed) HAMMER_HOME=${HAMMER_HOME}"
-else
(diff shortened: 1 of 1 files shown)


sourceme.sh
```diff
@@ -1,10 +1,24 @@
 #!/bin/sh
 # Convenience script for setting variables for in-tree use of hammer.
-if [ -z "$HAMMER_HOME" ]; then
-    >&2 echo "Must set HAMMER_HOME to root. Try export HAMMER_HOME="\$PWD" for the current dir."
+
+if [ -z "${HAMMER_HOME}" ]; then
+  # Try to find the location of the script even if it's called through a symlink.
+  #  https://stackoverflow.com/questions/59895/how-to-get-the-source-directory-of-a-bash-script-from-within-the-script-itself
+  SOURCE="${BASH_SOURCE[0]}"
```
jwright6323, line 7: This seems pretty complicated for what I would think would be a fairly simple thing. Does
```
dirname ${BASH_SOURCE[0]} | xargs readlink -f
```
 not work?

growly: some version of this does work for the platforms we care about and the invocation methods we are most likely to use (linux, bash, and one of sourcing or executing the script); much nicer, thanks

sourceme.sh
```diff
@@ -1,10 +1,17 @@
...
 # Convenience script for setting variables for in-tree use of hammer.
-if [ -z "$HAMMER_HOME" ]; then
-    >&2 echo "Must set HAMMER_HOME to root. Try export HAMMER_HOME="\$PWD" for the current dir."
+
+if [ -z "${HAMMER_HOME}" ]; then
+  # Try to find the location of the script even if it's called through a symlink.
+  #  https://stackoverflow.com/questions/59895/how-to-get-the-source-directory-of-a-bash-script-from-within-the-script-itself
+  #  https://stackoverflow.com/questions/35006457/choosing-between-0-and-bash-source
+  export HAMMER_HOME="$( dirname "$( readlink -f "${BASH_SOURCE[0]}" )" )"
+  >&2 echo "Using (guessed) HAMMER_HOME=${HAMMER_HOME}"
```
jwright6323, line 9: I'm not sure I'd add the (guessed)/(existing) parts here because it might actually confuse people. In this case if you source this more than once, it will tell you that it's using an (existing) HAMMER_HOME the second time because you set it the first time. IMO, a message like this would be a little more explanatory.
```suggestion
  >&2 echo "Setting HAMMER_HOME to ${HAMMER_HOME}"
```

sourceme.sh
```diff
@@ -1,10 +1,17 @@
...
 else
-export HAMMER_VLSI="$HAMMER_HOME/src/hammer-vlsi"
-export PYTHONPATH="$HAMMER_HOME/src:$HAMMER_HOME/src/jsonschema:$HAMMER_HOME/src/python-jsonschema-objects:$HAMMER_HOME/src/hammer-tech:$HAMMER_HOME/src/hammer-vlsi:$HAMMER_HOME/src/tools/pyyaml/lib3:$PYTHONPATH"
-export MYPYPATH="$PYTHONPATH"
-export PATH="$HAMMER_HOME/src/hammer-shell:$PATH"
+  >&2 echo "Using (existing) HAMMER_HOME=${HAMMER_HOME}"
```
jwright6323, line 11:
```suggestion
```
IMO don't echo anything here if HAMMER_HOME is set

sourceme.sh
growly: done

sourceme.sh
growly: done

PR #586, opened 2020-06-30
Error if we can't load the specified technology

colinschmidt:
I believe this code was only around to try to have a fallback in case the technology passed didn't have an `__init__.py` which I think isn't something we truly support.
This fix also allows for changes like #576 or any otherwise broken (via ImportError) tech plugins to be noticed rather than silently ignored.


edwardcwang:
While this LGTM to me in principle, it appears to break `nop`?

colinschmidt:
Thanks. Working on fixing the errors (it also affected dummy28 tech).

colinschmidt:
@edwardcwang Fixed.

PR #591, opened 2020-08-18
Add support for passing in LDFLAGS during simulation

abejgonzalez:
This PR is linked with https://github.com/ucb-bar/chipyard/pull/650.

It adds a new YAML input to pass in `LDFLAGS` into the VCS/C linker. It also renames the previous C compiler pass-in flag.


abejgonzalez:
Ok this is tested and seems to work. See https://github.com/ucb-bar/chipyard/pull/650 for more info.

PR #592, opened 2020-08-29
Power Docs

harrisonliew:
- Add documentation for power tool
- Fix up small errors in other flow documentation (e.g. inconsistency b/w `build` and `OBJ_DIR`)
- Add some missing settings in `defaults.yml`

Will need to be updated again when we add Joules plugin.


doc/Hammer-Flow/Simulation.rst
```diff
@@ -75,9 +87,17 @@ Simulation Input Keys
 
         * Setting to false means that the simulation will be entirely functional. Setting to true means that the simulation will be time annotated based on synthesis or P&R results.
 
+    * ``saif.mode`` (``"time"`, ``"trigger"``, ``"trigger_raw"``, or ``"full"``)
```
colinschmidt, line 90: I think you are missing a ` after time.

harrisonliew: Thanks, fixed

PR #598, opened 2020-12-30
Fix ASAP7 SRAM behav models

jerryz123:
Original behavioral verilog only worked if OEB was tied low.

Attached image demonstrates behavior when OEB is set low only during read-address cycle. Note the incorrect O_i in the subsequent cycle.
image

### DIFF
--- a/src/hammer-vlsi/technology/asap7/sram_compiler/memories/behavioral/sram_behav_models.v
+++ b/src/hammer-vlsi/technology/asap7/sram_compiler/memories/behavioral/sram_behav_models.v
@@ -76,7 +76,7 @@ always @ (posedge CE_i)
 		memory[A_i] = I_i;
 		
 
-always @ (data_out or OEB_i)
+always @ (posedge CE_i)
 	if (!OEB_i) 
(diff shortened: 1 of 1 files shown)


harrisonliew (approved):
This seems okay, but does it need to be synchronous with CE? Seems like it's intended to be a mux between `data_out` and `1'bz` conditional on `OEB_i`.

jerryz123:
> Seems like it's intended to be a mux between data_out and 1'bz conditional on OEB_i.

Doesn't this mean OEB_i should be set low the cycle after CSB_i? This is not how macrocompiler generates the active-low read-enable signal, which is what OEB is set as in the asap7 sram json.

An alternative fix could be to make macrocompiler tie OEB low. I don't really care which solution is used,

harrisonliew:
> Doesn't this mean OEB_i should be set low the cycle after CSB_i? This is not how macrocompiler generates the active-low read-enable signal, which is what OEB is set as in the asap7 sram json.

No, these are single-cycle read SRAMs, so OEB_i and CSB_i should be set in the same cycle.

PR #599, opened 2021-03-11
Major ASAP7 Updates

### LINKED ISSUE #571: ASAP7 version conflict in ASAP7 tutorial
Hi I'm trying to run through this example: https://chipyard.readthedocs.io/en/latest/VLSI/Tutorial.html. However when I download ASAP7 from the link given, it gives me release 1p6, but all the documentation and scripts seem to require 1p5. I'm not sure if this is the right place to this but these docs probably need to be updated with either a link to download 1p5 or scripts need to be updated to handle the changes in 1p6.

Additionally do you know if it is possible to download 1p5. I am a UC Berkeley masters student and I just need to run synthesis for a class project.

harrisonliew:
Small updates to plugin:
- Improve DRC/LVS results with don't use cells and tech-supplied floorplan update hook
- Ignore timing calculation for falling edge of reset to async DFF (missing removal_falling check)
- Small documentation update to indicate compatible PDK & gdspy versions

Big SRAM updates:
- Closes #598 (fix included)
- Closes #571 (documentation)
- New SRAMs that drastically reduce number of mapped SRAMs for most Rocket- and BOOM-based designs
  - 1RW(1024x16,17,32,37,38,44,64;128x12,14,22,40,64;2048x8;256x64;4096x8,16;512x64;64x20,21,22)
  - 2RW(32x33;64x24)
  - For perspective: `TinyRocketConfig` previously required 25 macros, now only requires 7
  - Still no masking (makes Verilog models too complex)
- Timing models thanks to https://github.com/sehuang/phyrilog!
  - Clk-q timing for `O` bus (generally around 60-400ps delay, 5-150ps transition)
  - Setup/hold timing for `A` and `I` buses (generally around +20-100ps, more for setup than hold)
  - Timing is equal for rise/fall for simplicity, but varies between SS/0.63V/100C and FF/0.77V/0C corners, and has slight dependence on SRAM depth
  - Excel calculation formulas: [asap7_sram_timing_model.xlsx](https://github.com/ucb-bar/hammer/files/6120041/asap7_sram_timing_model.xlsx)
  - Timing is roughly based on scaling from 12/14/16/22nm processes
- Updated sizes to be more realistic
  - Based on total bitcell area rather than just depth
  - Aspect ratios somewhat between 1 and 4 (generally tall to accommodate M4 min. pitch pins on one side)
  - Area roughly scaled from 22nm process (based on rough std. cell ratio)
- P&R tested for `TinyRocketConfig`, timing closure at 2GHz
  - Critical path endpoint is DCache (4096x8 SRAM)


edwardcwang:
Is https://github.com/sehuang/phyrilog meant to be accessible?

sehuang:
The repo is private as I wasn't paying much attention on committing things that could potentially be under NDA. I can add you to it @edwardcwang if you would like to take a look.

If the repo is clean, I can make it public and then it should be accessible.

harrisonliew:
I don't see anything under an NDA in your repo. The only thing that's sourced from the PDK is a layermap file. Everything else is open-sourceable.

I also want to end up submoduling your repo under Hammer for a situation where users want to compile an SRAM of (almost) arbitrary size.

edwardcwang:
Well, perhaps we can use it under a sram_generator abstraction? That way we can support real ones like OpenRAM.

sehuang:
> I don't see anything under an NDA in your repo. The only thing that's sourced from the PDK is a layermap file. Everything else is open-sourceable.
> 
> I also want to end up submoduling your repo under Hammer for a situation where users want to compile an SRAM of (almost) arbitrary size.

I'm not quite clear on the status of ASAP7 in terms of how open it is. Would including the layermap in the repo be going too far? If it is I can remove it from the repo before making it public and letting you submodule it into Hammer.

harrisonliew:
Yes, the idea is to use it as another sram_generator abstraction, albeit very rudimentary and subject to its own plugin-like structure. I'll be giving that a try in a bit after this PR goes through.

@sehuang no, I do not think so since layermaps only display layer names but do not reveal anything about the technology itself, which would be things like the transistor models and standard cell libraries. In fact, the layerprops file which is actually part of this PR also shows the layer names too.

sehuang:
Okay, sounds good. I've un-privated https://github.com/sehuang/phyrilog so it should now be accessible if you would like to submodule it.

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -212,9 +213,30 @@ def asap7_innovus_settings(ht: HammerTool) -> bool:
...
+    """
+    This is needed to move the core up by 1 site and re-do wiring tracks.
+    This resolves many DRCs and removes the need for the user to do it in placement constraints.
+    """
+    ht.append('''
+# Need to delete and recreate tracks based on tech LEF
+add_tracks -honor_pitch
+
+# Create place blockage on bottom row, fixes wiring issue + power vias for LVS
+set core_lly [get_db current_design .core_bbox.ll.y]
```
colinschmidt, line 228: This is interesting, and make me wonder what would happen if the user did this? Does it matter if the number of core sites is even or odd? Why does the issue not occur at the top?

src/hammer-vlsi/technology/asap7/sram-cache.json
```diff
@@ -1,10 +1,415 @@
 [
+{
+  "type" : "sram",
```
colinschmidt, line 3: If these all have the same ports do we have the ability to use the more compact sram-compiler syntax?

colinschmidt (approved):
Looks good to me, a few questions/comments.

src/hammer-vlsi/technology/asap7/__init__.py
harrisonliew: What do you mean if the user did this? Like, they created another blockage in the same place?

No, it doesn't matter if the number of rows is even or odd since it only places a blockage over the bottom-most row. For some reason that is unexplained, a variation of this is present in the reference script for ASAP7 and without it, everything shorts in the bottom row.

src/hammer-vlsi/technology/asap7/sram-cache.json
harrisonliew: Hm, I will need to examine, but I was under the impression that the ranges syntax was only available for MacroCompiler's --user-compiler mode.

harrisonliew: Verdict is that yes, it would work. However, given the selection of SRAMs I have (in `srams.txt`), it would save <10% of the json entries. Also would complicate the `sram-cache-gen.py` script used to generate new versions of the `sram-cache.json` file.

PR #600, opened 2021-04-02
Fix set_clock_groups syntax in HasSDCSupport

jerryz123:


src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -1476,10 +1476,10 @@ def sdc_clock_constraints(self) -> str:
                     groups[clock.group] = [clock.name]
             else:
                 ungrouped_clocks.append(clock.name)
-        if len(groups):
+        if len(groups) + len(ungrouped_clocks):
```
colinschmidt, line 1479: Isn't this just equivalent to `if len(clocks):` now?
Basically we want to do this as long as there are any clocks in the system.
The single clock or single group design will be treated as a no-op but I think for simplicity of this code that is fine.

colinschmidt (approved):
LGTM. Not sure why we avoided this aspect of the syntax before.
Also one small fix requested.

src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
jerryz123: Yeah `len(clocks)` is neater.

PR #602, opened 2021-04-14
Pass p_syn_in to hierarchical make rules.

jerryz123:
Previously, hierarchical steps did not see a dependence on inputs.yml, sram-generator-output.json, and your INPUT_CONFS in the hierarchical flow.

A bug could occur if the hierarchical module contained its own SRAM macro that no leaf module included. Without appending to p_syn_in, sram-generator-output.json would not be included, and the tools would not be able to resolve the path to this macro.


colinschmidt:
It is usually very annoying to have the later steps depend on INPUT_CONFS.
From my reading of the change you are just adding them as inputs without adding them as dependencies, which avoids this issue.
However I feel like you might want to only include the sram-generator-output.json instead of all inputs.

jerryz123:
I thought the workaround for that is to just use `redo` steps.

Not sure how I can parse out just the sram-generator-output.json without adding a weird special case here.

PR #603, opened 2021-05-10
add option to only report power for extra MMMC corners

harrisonliew:
Companion to ucb-bar/hammer-cadence-plugins#59


src/hammer-vlsi/defaults.yml
```diff
@@ -940,6 +940,10 @@ power.inputs:
   # type: List[str]
   saifs: []
 
+  # If true, power tool will report for only the extra MMMC corners, saving runtime.
```
dpgrubb13, line 943: Nit, but might be good to add in "(eg. nominal/typical)" to the comment again here, in case users miss the addition of extra corners in the mmmc_corners input and don't get what this is for.

dpgrubb13 (approved):
LGTM

PR #604, opened 2021-05-14
Basic hierarchical support for sim & power

harrisonliew:
Pass hier stuff to sim & power actions. 

TODO: tool plugin implementation will require some additional documentation/keys (e.g. benchmarks are only valid for top-level hierarchy).


jerryz123 (approved):
Seems reasonable to me. Most of the implementation looks like it will be in the tool plugins

PR #606, opened 2021-05-27
Decap API

harrisonliew:
This introduces a new decap API, which lets designers have much more control over how decaps are filled. Reasons for this:

- This control is often desired for extremely low-power (e.g. biosignal classifiers) where even decap leakage is of concern.
- A rail analysis flow can feedback hotspot areas where decap filling can then be targeted.

A summary of changes:

- Special cells has a new cell decap cell type
- The special cells size is altered to be a `List[str]` so that the decap value of each cell in the decap list can have a capacitance value attached. This is used for a P&R tool to fill up to a target amount of decapacitance.
- A new `DecapConstraint` is introduced along with a corresponding `vlsi.inputs.delays` key. The designer can target a decap density or a specific capacitance value in this list of constraints. For both targets, an area constraint can be specified with x, y, width, and height.
- Requisite documentation (including new docs for SpecialCells)

The addition of decap special cells is needed by ucb-bar/hammer-cadence-plugins#58.


src/hammer-tech/specialcells.py
```diff
@@ -60,7 +62,7 @@ class SpecialCell(NamedTuple('SpecialCell', [
     def from_setting(d: dict) -> "SpecialCell":
         size = d['size']
         if size is not None:
-            size = Decimal(str(d['size']))
+            size = list(map(lambda x: str(x), size))
```
dpgrubb13, line 65: Do we want to add a check here to ensure that the length of the size field matches the length of the name field?

dpgrubb13: The check is done in the plugin, but does it make sense to do it here instead, or do you think it's better to just retain flexibility for using the size field with any of the special cells?  I suppose it doesn't matter much either way.

harrisonliew: I think it's better to retain flexibility for now, since "size" is only being used by one special cell at the moment, and future use cases of this field are TBD.

dpgrubb13 (approved):

Issue #573, opened 2020-05-21
json schema keys inconsistently need underscores


growly:
Chatting to @harrisonliew, this seems to be a bug: some keys in the tech json use spaces ("gds map file") and some need underscores ("additional_drc_text")

harrisonliew:
Furthermore, they seem to break at random times/in random situations (example above - some plugins continue to use `additional drc text` without underscores just fine). However, keys with underscores seem to always work.

colinschmidt:
I wonder if this has to do with python versions.
It seems unlikely to be truly "random"

colinschmidt:
Resolution: See if we can nail down the situations in which it fails and which it doesn't. Potentially writing a unit test that will test against versions of python to see if that is the issue.
Also consider making the incompatible change of requiring underscores in keys going forward.

harrisonliew:
This is actually somewhat bypassed by this, but not a good workaround at all:
https://github.com/ucb-bar/hammer/blob/master/src/hammer-tech/library_filter.py#L14-L30

harrisonliew:
This is due to legacy uses of spaces requiring the forked [python-jsonschema-objects](https://github.com/edwardcwang/python-jsonschema-objects/tree/f7b828956867642a3569b74fa515eb22df0ebd5e). You must not have a published version installed in your Python environment. Will investigate if there's a way to shim this to remove this dependency.

harrisonliew:
No longer an issue in new Hammer.

PR #607, opened 2021-05-28
Rail Analysis

harrisonliew:
Updates to enable rail analysis (so far in Voltus ucb-bar/hammer-cadence-plugins#58):

- Add specification for Spice model files in tech library. A path is mandatory, but a lib corner is optional in the case that a model file uses the `.LIB` flag for corners.
- Add specification for foundry-supplied power grid views in tech library.
- Add lefdef layer map and Spice models for ASAP7.

~~HELP NEEDED for type checking: why does the python_jsonschema_objects.ObjectBuilder generate underscored properties for some but not all things in the schema.json (#573)~~


src/hammer-tech/hammer_tech.py
```diff
@@ -646,7 +670,11 @@ def prepend_dir_path(self, path: str, lib: Optional[Library] = None) -> str:
         assert len(path) > 0, "path must not be empty"
 
         # If the path is an absolute path, return it as-is.
-        if path[0] == "/":
+        if path[0] == os.sep:
```
edwardcwang, line 673: Does `os.path.isabs()` work for this purpose?

edwardcwang (approved):
Generally looks good to me.

src/hammer-tech/hammer_tech.py
harrisonliew: Ah, it does, good point.

dpgrubb13 (approved):

PR #610, opened 2021-07-07
Save and restore terminal settings around commands Fixes #69

### LINKED ISSUE #69: After abruptly terminating hammer-vlsi, the terminal needs reset
The terminal is left in a weird state and I have to clear or reset it to normalize.

colinschmidt:
Tested that this works for me when I ctrl+c during a tool run

### DIFF
--- a/src/hammer-vlsi/hammer_vlsi/submit_command.py
+++ b/src/hammer-vlsi/hammer_vlsi/submit_command.py
@@ -10,6 +10,8 @@
 
 import atexit
 import subprocess
+import termios
+import sys
 import datetime
 from abc import abstractmethod
 from functools import reduce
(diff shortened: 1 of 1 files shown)


colinschmidt:
FYI: I followed something similar to this example: https://docs.python.org/3/library/termios.html#example

edwardcwang (approved):

harrisonliew (approved):
Seems like the CI is unhappy about termios, but otherwise, LGTM.

src/hammer-vlsi/hammer_vlsi/submit_command.py
```diff
@@ -123,12 +125,18 @@ def submit(self, args: List[str], env: Dict[str, str],
         # Just run the command on this host.
 
         prog_tag = self.get_program_tag(args)
+        term_settings = termios.tcgetattr(sys.stdin.fileno())
```
edwardcwang, line 128:
```suggestion
        term_settings: Optional[list] = None
        try:
            termios.tcgetattr(sys.stdin.fileno())
        except termios.error:
            # Command should still run even if some tty features are not present
            pass
```

src/hammer-vlsi/hammer_vlsi/submit_command.py
```diff
@@ -123,12 +125,18 @@ def submit(self, args: List[str], env: Dict[str, str],
...
 
         logger.debug("Executing subprocess: " + ' '.join(args))
         subprocess_logger = logger.context("Exec " + prog_tag)
         proc = subprocess.Popen(args, shell=False, stderr=subprocess.STDOUT,
                                 stdout=subprocess.PIPE, env=env, cwd=cwd)
+        # These are run in reverse order of registration
+        # Terminate is first to allow for the possibility of graceful shutdown
+        # Then the program will be forceably terminated and we restore the terminal settings
+        atexit.register(termios.tcsetattr, sys.stdin.fileno(), termios.TCSANOW, term_settings)
```
edwardcwang, line 137:
```suggestion
        if term_settings is not None:
            atexit.register(termios.tcsetattr, sys.stdin.fileno(), termios.TCSANOW, term_settings)
```

edwardcwang (changes requested):
Hmm how about something like this?

edwardcwang:
(didn't actually test but something like this idea)

colinschmidt:
I did a slightly more "intent-full" design. We only need to worry about terminal settings if we are a tty.

edwardcwang (commented):
Will squash merge once the PR updates

edwardcwang (approved):

PR #611, opened 2021-07-24
Early rail analysis

harrisonliew:
Add `par.rail_analysis_power` key for ucb-bar/hammer-cadence-plugins#67. Defaults to 1 (unit of Watts).


harrisonliew:
After discussion with @dpgrubb13, it makes more sense for this key to be in the Innovus plugin.

PR #612, opened 2021-07-26
Fix 512x64 SRAM model

harrisonliew:
Fixes ucb-bar/chipyard#929


harrisonliew:
Reviewed by @baichen318

PR #618, opened 2021-09-28
Assume a string in custom constraints is length 1 list

colinschmidt:
This makes it easier to use transclude to include an entire file.
Otherwise your file will be split out one character per line.


src/hammer-vlsi/hammer_vlsi/hammer_vlsi_impl.py
```diff
@@ -1515,6 +1515,8 @@ def sdc_pin_constraints(self) -> str:
 
         # Custom sdc constraints that are verbatim appended
         custom_sdc_constraints = self.get_setting("vlsi.inputs.custom_sdc_constraints")  # type: List[str]
+        if isinstance(custom_sdc_constraints, str):
+            custom_sdc_constraints = (custom_sdc_constraints,)
```
edwardcwang, line 1519:
```suggestion
        sdcs = self.get_setting("vlsi.inputs.custom_sdc_constraints")  # type: Union[List[str], str]
        if isinstance(custom_sdc_constraints, str):
            custom_sdc_constraints = [sdcs]
        else:
            custom_sdc_constraints = sdcs  # type: List[str]
```

edwardcwang (commented):
Monir nitpick

harrisonliew (approved):
Slightly modified Edward's suggestion for type checking, LGTM.

PR #619, opened 2021-09-28
If prependlocal is applied to list prepend to list entries

colinschmidt:
Previous behavior was to overwrite list with prepended path
This caused very bad behavior when using that path as a list,
with the list entries being single letters


edwardcwang:
Could we have a unit test?

harrisonliew (approved):
I can attest to running into this issue before. Amended a couple of unit tests to test this and verified that it works.

PR #622, opened 2021-09-29
Add hammer-config and meta documentation

edwardcwang:


doc/Hammer-Use/Hammer-Config.rst
```diff
@@ -0,0 +1,193 @@
...
+        adc: "yes"
+        dac: "no"
+```
+
+The basic idea of Hammer IR in YAML/JSON format is centered around a hierarchically nested tree of YAML/JSON dictionaries. For example, the above YAML snippet is translated to two variables which can be queried in code - `foo.bar.adc` would have `yes` and `foo.bar.dac` would have `no`.
+
+Overriding
+----------
+
+Hammer IR snippets frequently "override" each other. For example, a technology plugin might provide some defaults which a specific project can override with a project YAML snippet.
```
harrisonliew, line 23: This should be more specific (has been a huge source of confusion). We should note that there is a built-in order of precedence, which from lowest to highest (I think, though 2/3 might be backwards?) is 1) Hammer default, 2) tech plugin, 3) tool plugin, 4) User's Hammer IR. In 4), subsequent JSON/YAML files specified with `-p` in the command line have higher precedence, and keys appearing later if duplicated in a file also take precedence.

I think we discussed having a the debug json annotate the final source file for each key, but I can't find which issue it's in right now.

doc/Hammer-Use/Hammer-Config.rst
```diff
@@ -0,0 +1,193 @@
...
+For example, say that the technology plugin provides:
+
+```
+vlsi.tech.foobar65.bad_cells: ["NAND4X", "NOR4X"]
+```
+
+And let's say that in our particular project, we find it undesirable to use the `NAND2X` and `NOR2X` cells. However, if we simply put the following in our project YAML, the references to NAND4X and NOR4X disappear and we don't want to have to copy the information from the base plugin, which may change, or which may be proprietary, etc.
+
+```
+vlsi.tech.foobar65.bad_cells: ["NAND8X", "NOR8X"]
```
harrisonliew, line 41: These should be `NAND2X` and `NOR2X` based on above text?

doc/Hammer-Use/Hammer-Config.rst
```diff
@@ -0,0 +1,193 @@
...
+```
+
+This will yield the desired result of `["NAND4X", "NOR4X", "NAND8X", "NOR8X"]` when `get_settings("vlsi.tech.foobar65.bad_cells")` is called in the end.
+
+Applying multiple meta actions
+------------------------------
+
+Multiple meta actions can be applied sequentially if the `_meta` variable is an array. Example:
+
+Layer 1:
```
harrisonliew, line 60: "Layer" doesn't make sense to me. If these are names of files containing IR I think `prependlocal` makes more sense.

doc/Hammer-Use/Hammer-Config.rst
```diff
@@ -0,0 +1,193 @@
...
+foo.pipeline_meta: ['subst', 'prependlocal']
+```
+
+
+Result: `get_setting("foo.pipeline")` = `/opt/foo/CELL_yes.lef`
+
+Common meta actions
+-------------------
+
+* `append`: append the elements provided to the base list
```
harrisonliew, line 77: Maybe inconsistent that `append` doesn't have an example but `subst` does, even though both are introduced above?

doc/Hammer-Use/Hammer-Config.rst
```diff
@@ -0,0 +1,193 @@
...
+
+Meta:
+```
+foo.pipeline: "${foo.flash}man"
+foo.pipeline_meta: subst
+```
+
+Result: `get_setting("foo.flash")` = `yesman`
+
+* `lazysubst`: by default, variables are only substituted from previous configs. Using `lazysubst` allows us to deter the substitution until the very end.
```
harrisonliew, line 93: I think a general motivation of the `lazy...` modifier (waiting til all overriding is reconciled) is warranted in the section about key overriding or introducing meta variables?

doc/Hammer-Use/Hammer-Config.rst
```diff
@@ -0,0 +1,193 @@
+.. _config:
+
+The hammer-config library
```
harrisonliew, line 3: I think this section is better described as "Hammer IR and Meta Variables" and the hammer-config library can be a subsection explaining the Hammer IR <-> JSON/YAML translation.

Perhaps the `get_setting` referenced many times later can also be introduced as being available to everything (technologies, tool steps, hooks).

harrisonliew (changes requested):
This needs to be added in `doc/Hammer-Use/index.rst`. Probably should go first in that section. Also, section 1.1.1 Main Hammer should link to this doc for an in-depth look at how keys are parsed and modified.

doc/Hammer-Use/Hammer-Config.rst
```diff
@@ -0,0 +1,193 @@
...
+  quux: "32${foo.bar}"
+foo.bar_meta: deepsubst
+```
+
+Result: `get_setting("foo.bar.baz")` = `12345` and `get_setting("foo.bar.baz")` = `32123`
+
+Reference
+---------
+
+For a more comprehensive view, please consult the `hammer_config` API documentation as well as
```
harrisonliew, line 190: I kind of wish there were more examples, but instead of immediately delving into code, perhaps list out some formulaic combinations of the `lazy`, `cross`, etc. modifiers for `append`, `ref`, etc. and give a small gist of what they might do?

harrisonliew (changes requested):

doc/Hammer-Use/Hammer-Config.rst
edwardcwang: I agree - how do you think we could list it in a way that doesn't just turn into another list dump?

doc/Hammer-Use/Hammer-Config.rst
edwardcwang: I agree. What do you think would be a better description? What I'm trying to describe here is that `layer 1` would represent e.g. base/defaults, `layer 2` a tech plugin, `layer 3` a project YAML, etc.

harrisonliew: Oh I see. I think "level" is better (like "levels of precedence"). I think if you just denote in the section about the way precedence is handled that the examples below will use "Level #" to denote it from lowest to highest, that would be sufficient.

doc/Hammer-Use/Hammer-Config.rst
harrisonliew: Don't list out the actual combinations, rather something like this (though some don't quite fit in like transclude?)? You can use your judgment.

Base modifiers:
- lazy = leave til the end
- cross = <what is this?>
- ...

Base actions:
- append = add to end of list
- ref = look somewhere else
- ...

Most supported directives are a combination of these base modifiers and actions and the files below implement them:
...

doc/Hammer-Use/Hammer-Config.rst
harrisonliew: Fixed myself.

doc/Hammer-Use/Hammer-Config.rst
harrisonliew: Rescinding this after consideration.

doc/Hammer-Use/Hammer-Config.rst
harrisonliew: Turns out it's not so formulaic. Closing this.

harrisonliew:
I fixed/tested all the rST syntax. Approving it now.

harrisonliew (approved):

edwardcwang:
Thank you @harrisonliew !

PR #625, opened 2021-12-17
Hierarchical flow docs

harrisonliew:
First draft of documentation for hierarchical flow needed for 290C tapeout class.

The limitations at the bottom may need to be checked more thoroughly.

### DIFF
--- a/doc/Hammer-Use/Hierarchical.rst
+++ b/doc/Hammer-Use/Hierarchical.rst
@@ -87,9 +87,9 @@ In a bottom-up hierarchical flow, is is important to remember that submodules do
 Special Notes & Limitations
 ---------------------------
 
-#. Hammer IR keys propagate up through the hierarchical tree. For example, if ``vlsi.inputs.clocks`` was specified in the constraints for ``ModuleAA`` but not for ``ModuleA``, ``ModuleA`` will inherit ``ModuleAA``'s constraints. Take special care of where your constraints come from, especially for a parent module with more than one submodule.
+#. Hammer IR keys propagate up through the hierarchical tree. For example, if ``vlsi.inputs.clocks`` was specified in the constraints for ``ModuleAA`` but not for ``ModuleA``, ``ModuleA`` will inherit ``ModuleAA``'s constraints. Take special care of where your constraints come from, especially for a parent module with more than one submodule. To avoid confusion, it is recommended to specify the same set of keys for every module.
 
-#. Hammer IR keys specified at the root level (i.e. outside of ``vlsi.inputs.hierarchical.constraints``) do not override the corresponding submodule constraints. However, if you add a Hammer IR file using ``-p`` on the command line (after the file containing ``vlsi.inputs.hierarchical.constraints``), those keys are global and override submodule constraints unless a meta action is specified.
+#. Hammer IR keys specified at the root level (i.e. outside of ``vlsi.inputs.hierarchical.constraints``) do not override the corresponding submodule constraints. However, if you add a Hammer IR file using ``-p`` on the command line (after the file containing ``vlsi.inputs.hierarchical.constraints``), those keys are global and override submodule constraints unless a meta action is specified. To avoid confusion, it is recommended to specify all constraints with ``vlsi.inputs.hierarchical.constraints``.
 
 #. Due to the structure of ``vlsi.inputs.hierarchical.constraints`` as a list structure, currently, there are the following limitations:
 
--- a/doc/Hammer-Use/Hierarchical.rst
+++ b/doc/Hammer-Use/Hierarchical.rst
@@ -1,2 +1,110 @@
+.. _hierarchical:
+
 Hierarchical Hammer Flow
 ============================================
+
+Hammer supports a bottom-up hierarchical flow. This is beneficial for very large designs to reduce the computing power by partitioning it into submodules, especially if there are many repeated instances of those modules.
+
+Hierarchical Hammer Config
+--------------------------
+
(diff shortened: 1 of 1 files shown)


doc/Hammer-Use/Hierarchical.rst
```diff
@@ -1,2 +1,110 @@
...
+Special Notes & Limitations
+---------------------------
+
+#. Currently, Hammer IR keys do not propagate up through the hierarchical tree. For example, if ``vlsi.inputs.clocks`` was specified in the constraints for ``ModuleAA`` but not for ``ModuleAA``, ``ModuleAA`` will not inherit ``ModuleAA``'s constraints; instead, they will take Hammer's defaults.
+
+#. Hammer IR keys specified at the root level (i.e. outside of ``vlsi.inputs.hierarchical.constraints``) are global, i.e. they override submodule constraints unless a meta action is specified.
+
+#. Due to the structure of ``vlsi.inputs.hierarchical.constraints`` as a list structure, currently, there are the following limitations:
+
+    * You must include all of the constraints in a single file. The config parser is unable to combine constraints from differnt files because meta actions do not work on list items. This will make it harder for collaboration, and unfortnately, changes to module constraints at a higher level of hierarchy after submodules are hardened will trigger the Make dependencies, so you will need to modify the generated Makefile or use redo-targets.
```
colinschmidt, line 96: I believe the constraint is more subtle than this.
Meta actions can work on list items, if you want this you need to perform a `deepsusbt`.

doc/Hammer-Use/Hierarchical.rst
```diff
@@ -1,2 +1,110 @@
...
+
+    * Similarly, certain meta actions do not work as expected, especially those where string substitutions and file transclusion are needed. This is because the meta actions are resolved between flow actions but the meta actions themselves persist.
```
colinschmidt, line 98: Hopefully this can be removed if we merge #626

doc/Hammer-Use/Hierarchical.rst
```diff
@@ -1,2 +1,110 @@
...
+* You may need to specify pin delays ``vlsi.inputs.delays`` for many pins to optimize the partitioning of sequential signals that cross the submodule boundary. By default, without pin delay constraints, the input and output delay are constrained to be coincident with its related clock arrival at the module boundary.
+
+* Custom SDC constraints that originate from a higher level (e.g. false paths from async inputs) need to be specified in submodules as well.
+
+* Custom CPFs will need to be written if differently-named power nets need to globally connected between submodules. Similarly, hierarchical flow with custom CPFs can also be used to fake a multi-power domain topology until Hammer properly supports this feature.
+
+Special Notes & Limitations
+---------------------------
+
+#. Currently, Hammer IR keys do not propagate up through the hierarchical tree. For example, if ``vlsi.inputs.clocks`` was specified in the constraints for ``ModuleAA`` but not for ``ModuleAA``, ``ModuleAA`` will not inherit ``ModuleAA``'s constraints; instead, they will take Hammer's defaults.
```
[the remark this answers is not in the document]
harrisonliew: Ah, yeah it's a typo, but what you said is something I wanted checked. I think some amount of inheritance is desired, but we do need to rework how far that propagates through the graph.

doc/Hammer-Use/Hierarchical.rst
```diff
@@ -1,2 +1,110 @@
...
+
+#. Hammer IR keys specified at the root level (i.e. outside of ``vlsi.inputs.hierarchical.constraints``) are global, i.e. they override submodule constraints unless a meta action is specified.
```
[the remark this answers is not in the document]
harrisonliew: Oh, yeah you are correct. But, I found that if you add additional config fragments via `-p`, those global constraints do override. Anyways, I think this behavior is somewhat confusing, and we need to rework the precedence of constraints.

doc/Hammer-Use/Hierarchical.rst
harrisonliew: Sure, `deepsubst` would work, but most others that you would expect to use in non-hierarchical flows wouldn't?

doc/Hammer-Use/Hierarchical.rst
harrisonliew: Going to delete this now since fixed by that PR.

doc/Hammer-Use/Hierarchical.rst
```diff
@@ -1,2 +1,106 @@
...
+* You may need to specify pin delays ``vlsi.inputs.delays`` for many pins to optimize the partitioning of sequential signals that cross the submodule boundary. By default, without pin delay constraints, the input and output delay are constrained to be coincident with its related clock arrival at the module boundary.
+
+* Custom SDC constraints that originate from a higher level (e.g. false paths from async inputs) need to be specified in submodules as well.
+
+* Custom CPFs will need to be written if differently-named power nets need to globally connected between submodules. Similarly, hierarchical flow with custom CPFs can also be used to fake a multi-power domain topology until Hammer properly supports this feature.
+
+Special Notes & Limitations
+---------------------------
+
+#. Hammer IR keys propagate up through the hierarchical tree. For example, if ``vlsi.inputs.clocks`` was specified in the constraints for ``ModuleAA`` but not for ``ModuleA``, ``ModuleA`` will inherit ``ModuleAA``'s constraints. Take special care of where your constraints come from, especially for a parent module with more than one submodule.
```
jerryz123, line 90: Are there cases where this behavior is useful? 

It would also be helpful to include YAML snippets demonstrating the issue, for this and the next bullet point.

harrisonliew: I don't believe this behavior is actually useful, and these bullets are meant to warn against situations where observed results don't match against the input config. Fixing this aligns with our internal discussions to parse + propagate IR on a module-by-module basis in hierarchical mode. As such, I don't see the value in including a YAML example snippet for these, but I will update this to recommend explicitly writing all the same keys for every module to avoid any confusion.

jerryz123 (approved):

PR #626, opened 2021-12-27
Support relative directory meta application inside hierarchical constraints

colinschmidt:
We add the special key for tracking relative directories to each dictionary that contains a `vlsi.inputs.hierarchical.constraints`.
This allows any future use of this dictionary to be able to resolve `prependlocal` meta actions or any future path relative meta actions.
We avoid adding this path variable if it already exists in the case that this procedure has already occured. Overwriting the path in that case would cause us to update it and then potentially disconnect the relative path from its original directory.
This happens in the normal usage with syn-to-par step creating a new hammer_config file in a different level of the directory hierarchy.

### DIFF
--- a/src/hammer-vlsi/hammer_vlsi/driver.py
+++ b/src/hammer-vlsi/hammer_vlsi/driver.py
@@ -1284,6 +1284,24 @@ class HammerDriver:
 
             hier_placement_constraints = {key: list(map(partial(PlacementConstraint.from_masters_and_dict, masters), lst))
                                           for key, lst in combined_raw_placement_dict.items()}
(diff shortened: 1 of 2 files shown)


edwardcwang:
Unit test?

harrisonliew (approved):
LGTM.

harrisonliew:
> Unit test?

I'm inclined not to gate this PR based on a unit test because we don't have any unit tests for hierarchical mode besides this width/height matching: https://github.com/ucb-bar/hammer/blob/95428fb44f3d8eda395b496dbdd4e04689e08acb/src/hammer-vlsi/test.py#L775. This is beyond the scope of this PR (opening an issue instead).

Issue #585, opened 2020-06-24
can't start or stop on steps in all tools except syn and par


ssteffl:
I would like to use HAMMER_EXTRA_ARGS="--to_step xxx" in a new drc tool. but cli_driver.py doesn't only adds these tooks to the syn/par tools on line ~1000.

harrisonliew:
Ah, you see this now due to the fact that we changed some DRC/LVS plugins to write the runset files with steps. Previously, only syn/par used steps.

But, in general I agree that all tools should be able to start/stop on steps.

colinschmidt:
Resolution: We should generalize `to/from_step` to be supported on all tools.

harrisonliew:
Closed by #629

PR #629, opened 2022-01-25
Flow Control Fixes + Enhancements

### LINKED ISSUE #554: Replacement hook can't be targeted in flow control
Possibly a bug from #543?

Let's say we `make_replacement_hook("step_a", "step_b")`. `step_b` is correctly run instead of `step_a`. However, if doing a `--from_step` or similar, you have to target `step_a` instead of `step_b`. Logger info also prints out `step_a` when showing what it runs/skips.

Perhaps related: steps that are skipped (via removal hook) are still target in `do_post_steps` because it is seemingly still part of a transition.

harrisonliew:
Closes #554 and #585:
- Replaced steps use the new name in flow control
- Removed steps no longer targetable in flow control
- Get flow control hooks for drc, lvs, sim, power actions
- `first_step` and `persistent_steps` properties, `run_persistent_steps` method
- Persistent hooks and pre-persistent hooks whose target is the first step will always run at the beginning. See ucb-bar/hammer-cadence-plugins#71 for an example of how this is used in both `do_pre_steps` and the `open_chip` script.
- No change needed to unit tests

### DIFF
--- a/src/hammer-vlsi/hammer_vlsi/cli_driver.py
+++ b/src/hammer-vlsi/hammer_vlsi/cli_driver.py
@@ -1053,11 +1053,22 @@ class CLIDriver:
         stop_step = only_step or to_step or until_step or None
         stop_incl = (only_step or to_step) is not None
         if (start_step or stop_step) is not None:
-            # TODO: do this for the rest of the actions
             driver.set_post_custom_syn_tool_hooks(HammerTool.make_start_stop_hooks(
                 HammerStartStopStep(step=start_step, inclusive=start_incl),
                 HammerStartStopStep(step=stop_step, inclusive=stop_incl)))
             driver.set_post_custom_par_tool_hooks(HammerTool.make_start_stop_hooks(
+                HammerStartStopStep(step=start_step, inclusive=start_incl),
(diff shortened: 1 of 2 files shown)


jerryz123 (commented):
I wasn't familiar with the `persistent_step` API before, but this seems reasonable. 

> Persistent hooks and pre-persistent hooks whose target is the first step will always run at the beginning. See Improved do_pre_steps and open_chip script hammer-cadence-plugins#71 for an example of how this is used in both do_pre_steps and the open_chip script.

What was the old behavior?

harrisonliew:
Assumptions:
- The first tool step is `init_design`
- There is a `persistent_hook`
- There is a `pre_persistent_hook` targeting `init_design`

Before:
1. When running the full flow, the `persistent_hook` and `pre_persistent_hook` will run before `init_design`
2. When running a subset of the flow (e.g. with `--start_after_step`), the old `do_pre_steps` (e.g. `read_db`) would run before either persistent hook.

After:
1. When running the full flow, the `persistent_hook` and `pre_persistent_hook` will run before `init_design`
2. When running a subset of the flow (e.g. with `--start_after_step`), the both persistent hooks run before the `read_db`.

jerryz123 (approved):

PR #631, opened 2022-01-27
Add bridge targets to Hammer Makefile

jerryz123:


harrisonliew (approved):

harrisonliew:
Oops, buildbot came back online and the Makefile tests are now broken because these new bridge targets are not available. Will PR a fix.

PR #633, opened 2022-02-12
Sky130

nayiri-k:
1) add endcaps to PAR
2) connect global nets VDD/VSS to the pg_pins in the sky130 standard cells (VPWR, VGND, VPB, VNB)
3) other minor edits to plugin

### DIFF
--- a/src/hammer-vlsi/technology/sky130/__init__.py
+++ b/src/hammer-vlsi/technology/sky130/__init__.py
@@ -8,7 +8,7 @@
 import sys
 import re
 import os, shutil
-from pathlib import Path 
+from pathlib import Path
 from typing import NamedTuple, List, Optional, Tuple, Dict, Set, Any
 
 import hammer_tech
@@ -44,9 +44,9 @@ class SKY130Tech(HammerTechnology):
         if not cdl_old_path.exists():
             raise FileNotFoundError(f"CDL not found: {cdl_old_path}")
 
-        cache_tech_dir_path = Path(self.cache_dir) 
+        cache_tech_dir_path = Path(self.cache_dir)
         os.makedirs(cache_tech_dir_path, exist_ok=True)
-        cdl_new_path = cache_tech_dir_path / f'{self.library_name}.cdl' 
+        cdl_new_path = cache_tech_dir_path / f'{self.library_name}.cdl'
 
         f_old = open(cdl_old_path,'r')
         f_new = open(cdl_new_path,'w')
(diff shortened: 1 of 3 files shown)


src/hammer-vlsi/technology/sky130/__init__.py
```diff
@@ -264,5 +269,30 @@ def sky130_power_nets(ht: HammerTool) -> bool:
                 ht.verbose_append("connect_global_net {tie} -type net -net_base_name {net}".format(tie=pwr_gnd_net.tie, net=pwr_gnd_net.name))
     return True
 
-tech = SKY130Tech()
+def sky130_connect_nets(ht: HammerTool) -> bool:
```
harrisonliew, line 272: Hm, this really should not need to be a hook. Question: are you using a CPF, or allowing Hammer to auto-generate the power intent? If it's the former, these global net connections should be in the CPF, If it's the latter, we really should add this `connect_global_net` here instead: https://github.com/ucb-bar/hammer-cadence-plugins/blob/4ecbce344302a5e697c2d54fd2fd80c3f70db186/par/innovus/__init__.py#L298

harrisonliew (changes requested):

src/hammer-vlsi/technology/sky130/__init__.py
harrisonliew: Also, I think the DRC & LVS hooks should be made default (comments in ucb-bar/chipyard#1115).

src/hammer-vlsi/technology/sky130/__init__.py
```diff
@@ -264,5 +269,30 @@ def sky130_power_nets(ht: HammerTool) -> bool:
...
+    assert isinstance(
```
harrisonliew, line 273: You also need an assertion that `ht` is a `TCLTool` to get rid of the type checker failures against `append` and `verbose_append`.

harrisonliew (changes requested):

src/hammer-vlsi/technology/sky130/__init__.py
nayiri-k: Made DRC/LVS hooks default now, and will move the sky130_power_nets hook to the auto-generated CPF in a future PR.

src/hammer-vlsi/technology/sky130/defaults.yml
```diff
@@ -18,8 +18,9 @@ technology.sky130:
 
 # Mentor environment variables
 # Override this in project
+mentor.extra_env_vars_meta: lazydeepsubst
 mentor.extra_env_vars:
-  - PDK_HOME: "<path-to-skywater-src-nda>/s8/<version#>"
+  - PDK_HOME: "${technology.sky130.sky130_nda}/s8/V2.0.1"
```
harrisonliew, line 23: Oh, turns out that this is exactly the same as `technology.sky130.pdk_home`

harrisonliew: Nvm, you can't lazy twice.

harrisonliew (approved):

PR #635, opened 2022-02-22
Yosys plugin

nayiri-k:
Initial pass at a plugin that generates the TCL script consumed by the Yosys synthesis tool.
The previous Yosys plugin called the Makefile in the Yosys git repo.
Currently generating a netlist, but the quality of this netlist has not been verified yet.

### DIFF
--- a/src/hammer-vlsi/synthesis/yosys/__init__.py
+++ b/src/hammer-vlsi/synthesis/yosys/__init__.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 #
 # Yosys synthesis plugin for Hammer (as part of OpenROAD-flow installation).
+# mostly adoped from OpenLANE scripts found here: https://github.com/The-OpenROAD-Project/OpenLane/tree/master/scripts
 #
 # See LICENSE for licence details.
 
@@ -13,7 +14,7 @@ from hammer_vlsi.vendor import OpenROADTool, OpenROADSynthesisTool
 ###########
 from hammer_vlsi import HammerTool, HammerToolStep, HammerToolHookAction, HierarchicalMode
 from hammer_vlsi.constraints import MMMCCorner, MMMCCornerType
-from hammer_utils import VerilogUtils
+from hammer_utils import VerilogUtils, optional_map
 from hammer_vlsi import HammerSynthesisTool
 from hammer_logging import HammerVLSILogging
 from hammer_vlsi import MMMCCornerType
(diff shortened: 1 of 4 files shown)


src/hammer-vlsi/synthesis/yosys/defaults.yml
```diff
@@ -2,9 +2,12 @@
...
 synthesis.yosys:
+  # Location of the binary.
+  binary: "yosys"
   # this is just to make hammer happy (although, it is never called AFAICT).
   # the version is actually calculated by the OpenROADTool class
   version: "dummy-value"
 
   # Generate the input scripts but don't run anything
-  generate_only: false
+  # TODO: make this tool-independent?
```
harrisonliew, line 12: Indeed, we have wanted to add a `-d` or `--dry_run` command line argument in CLIDriver.

src/hammer-vlsi/synthesis/yosys/__init__.py
```diff
@@ -2,20 +2,34 @@
...
 
+import specialcells
+from specialcells import CellType, SpecialCell
+
+import os
+import json
+from collections import Counter
 
-class YosysSynth(OpenROADSynthesisTool):
+class YosysSynth(HammerSynthesisTool, OpenROADTool):
```
harrisonliew, line 32: Confused about why this is no longer mixing in `OpenROADSynthesisTool`. Are some things in that class conflicting with what you need for Yosys? Since Yosys is the OpenROAD synthesis tool... should some of the below things be moved there instead?

src/hammer-vlsi/synthesis/yosys/__init__.py
```diff
@@ -2,20 +2,34 @@
 # -*- coding: utf-8 -*-
 #
 # Yosys synthesis plugin for Hammer (as part of OpenROAD-flow installation).
+# mostly adoped from OpenLANE scripts found here: https://github.com/The-OpenROAD-Project/OpenLane/tree/master/scripts
```
harrisonliew, line 5: Looking at the scripts: I see some flags related to synthesis strategies and adder topologies. Do you intend to add these as keys?

harrisonliew (changes requested):

harrisonliew:
Documentation should be updated to reflect open-source tool/tech plugin support, but that can be a separate PR if you wish.

src/hammer-vlsi/synthesis/yosys/__init__.py
```diff
@@ -59,7 +110,7 @@ def export_config_outputs(self) -> Dict[str, Any]:
         outputs["synthesis.outputs.sdc"] = self.output_sdc
         outputs["synthesis.outputs.seq_cells"] = self.output_seq_cells
         outputs["synthesis.outputs.all_regs"] = self.output_all_regs
-        outputs["synthesis.outputs.sdf_file"] = self.sdf_file
+        # outputs["synthesis.outputs.sdf_file"] = self.sdf_file
```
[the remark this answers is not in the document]
nayiri-k: I agree with option 2), while sta can be executed within openroad, it can't with yosys so would require a separate TCL script to be generated anyways

harrisonliew:
Superceded by #648

PR #648, opened 2022-05-02
Yosys + Openroad Plugins

### LINKED ISSUE #489: plugin for OpenROAD
This EDA [project](https://github.com/The-OpenROAD-Project/) seems interesting, full opensource without NDA issue.
maybe this can be a solution for public flow generating for hammer.

nayiri-k:
Initial implementations of Yosys and OpenROAD plugins

Closes #489


src/hammer-vlsi/par/openroad/__init__.py
```diff
@@ -84,148 +181,959 @@ def tool_config_prefix(self) -> str:
...
-        self.cmds += [
-          "make DESIGN_CONFIG={conf} -f {make} finish".format(
-            conf=self.design_config_path(), 
-            make=self.openroad_flow_makefile_path()
-        )]
+        self.block_append(f"""
+        ################################################################
+        # Write Design
+        """)
+        self.append(f"write_def {self.output_def_filename}")
```
harrisonliew, line 845: If klayout needs DEFs for writing GDS, isn't this one the only one that's necessary? Any advantage to writing checkpoint DEFs?

src/hammer-vlsi/par/openroad/__init__.py
```diff
@@ -84,148 +181,959 @@ def tool_config_prefix(self) -> str:
...
+
         return True
 
+    
+
+    @staticmethod
+    def generate_chip_size_constraint(width: Decimal, height: Decimal, left: Decimal, bottom: Decimal, right: Decimal,
+                                      top: Decimal, site: str) -> str:
+        """
+        Given chip width/height and margins, generate an Innovus TCL command to create the floorplan.
```
harrisonliew, line 874: nit: not Innovus

src/hammer-vlsi/par/openroad/__init__.py
```diff
@@ -84,148 +181,959 @@ def tool_config_prefix(self) -> str:
...
+        output = []
+        # initialize_floorplan removes existing tracks 
+        #   --> use the make_tracks command to add routing tracks 
+        #       to a floorplan (with no arguments it uses default from tech LEF)
+        layers = self.get_setting("par.generate_power_straps_options.by_tracks.strap_layers")
+        for metal in self.get_stackup().metals:
+            output.append(f"make_tracks {metal.name}")
+        return output
+
+    def create_floorplan_tcl(self) -> List[str]:
```
harrisonliew, line 890: It looks to me like there's a lot of stuff copied over from Innovus left in here?

src/hammer-vlsi/synthesis/yosys/__init__.py
```diff
@@ -68,88 +119,256 @@ def tool_config_prefix(self) -> str:
...
     #========================================================================
     # synthesis main steps
     #========================================================================
     def init_environment(self) -> bool:
-        # TODO: currently using OpenROAD's default synthesis script
+
+        # set variables to match global variables in OpenLANE
+        self.driving_cell = "sky130_fd_sc_hd__inv_1" # SYNTH_DRIVING_CELL
+        self.driving_cell_pin = "Y" # SYNTH_DRIVING_CELL_PIN
+        self.buffer_cell = "sky130_fd_sc_hd__buf_2 A X"
```
harrisonliew, line 210: Replace with new SpecialCells?

src/hammer-vlsi/synthesis/yosys/__init__.py
```diff
@@ -68,88 +119,256 @@ def tool_config_prefix(self) -> str:
...
             HammerVLSILogging.enable_colour = True
             HammerVLSILogging.enable_tag = True
 
         return True
 
+    def write_sdc_file(self) -> bool:
+        with open(self.mapped_sdc_path,'w') as f:
+            # TODO: why buf cell and not inv??
+            # custom sdc constraints
+            f.write("set_driving_cell sky130_fd_sc_hd__buf_1\n" )
```
harrisonliew, line 190: ditto SpecialCell

src/hammer-vlsi/synthesis/yosys/__init__.py
```diff
@@ -68,88 +119,256 @@ def tool_config_prefix(self) -> str:
...
+            f.write("set_load 5\n")
+
+    def block_append(self,commands) -> bool:
+        for line in commands.split('\n'):
+            self.append(line.strip())
+            # if line.strip().startswith('#') or (line==""):
+            #     self.append(line.strip())
+            # else:
+            #     self.verbose_append(line.strip())
```
harrisonliew, line 199: clean up

src/hammer-vlsi/synthesis/yosys/__init__.py
```diff
@@ -68,88 +119,256 @@ def tool_config_prefix(self) -> str:
...
     def init_environment(self) -> bool:
-        # TODO: currently using OpenROAD's default synthesis script
+
+        # set variables to match global variables in OpenLANE
+        self.driving_cell = "sky130_fd_sc_hd__inv_1" # SYNTH_DRIVING_CELL
+        self.driving_cell_pin = "Y" # SYNTH_DRIVING_CELL_PIN
+        self.buffer_cell = "sky130_fd_sc_hd__buf_2 A X"
+
+        time_unit = self.get_time_unit().value_prefix + self.get_time_unit().unit
+        clock_port = self.get_clock_ports()[0]
```
harrisonliew, line 213: Only 1 clock spported?

src/hammer-vlsi/synthesis/yosys/__init__.py
```diff
@@ -68,88 +119,256 @@ def tool_config_prefix(self) -> str:
...
+        time_unit = "ps" # yosys requires time units in ps
+        self.clock_period = int(clock_port.period.value_in_units(time_unit))
+        self.clock_uncertainty = int(clock_port.period.value_in_units(time_unit))
+        self.clock_transition = 0.15 # SYNTH_CLOCK_TRANSITION
+
+        self.synth_cap_load = 33.5 # SYNTH_CAP_LOAD
+        self.max_fanout = 5 # default SYNTH_MAX_FANOUT = 5
+
+        # Yosys commands only take a single lib file
+        #   so use typical corner liberty file
```
harrisonliew, line 224: Since this is the case, I would rather filter the libs for the default supply voltage in `vlsi.inputs.supplies.VDD` and not use the MMMC construct because the "extra" corner is intended for other analysis such as typical power analysis.

However, this may cause ambiguity in temperature unless you hardcode 25C, so if you decide to leave it as-is, throw an error if `corner_tt` returns `None` that tells the user that an extra corner is required for Yosys.

src/hammer-vlsi/synthesis/yosys/__init__.py
```diff
@@ -68,88 +119,256 @@ def tool_config_prefix(self) -> str:
...
+        # TODO: figure out why this is
+
+        # begin, coarse labels: automatic
+        synth -top {self.top_module} -run :fine
+        # fine label: run manually 
+        opt -fast -full
+        memory_map
+        opt -full
+        # TODO: figure out why techmap path assumed correct for conda but not server install 
+        techmap -map "/usr/local/share/yosys/techmap.v"
```
harrisonliew, line 271: What is this file?

src/hammer-vlsi/synthesis/yosys/__init__.py
```diff
@@ -68,88 +119,256 @@ def tool_config_prefix(self) -> str:
...
+
+        time_unit = self.get_time_unit().value_prefix + self.get_time_unit().unit
+        clock_port = self.get_clock_ports()[0]
+        self.clock_port_name = clock_port.name
+        time_unit = "ps" # yosys requires time units in ps
+        self.clock_period = int(clock_port.period.value_in_units(time_unit))
+        self.clock_uncertainty = int(clock_port.period.value_in_units(time_unit))
+        self.clock_transition = 0.15 # SYNTH_CLOCK_TRANSITION
+
+        self.synth_cap_load = 33.5 # SYNTH_CAP_LOAD
```
harrisonliew, line 220: How is this chosen? It might be interesting to extract it from the driving cell...

src/hammer-vlsi/synthesis/yosys/__init__.py
```diff
@@ -68,88 +119,256 @@ def tool_config_prefix(self) -> str:
...
     def syn_map(self) -> bool:
-        # TODO: currently using OpenROAD's default synthesis script
+        self.block_append(f"""
+        # Technology mapping for cells
+        # ABC supports multiple liberty files, but the hook from Yosys to ABC doesn't
+        puts "second abc pass"
+        abc -D {self.clock_period} \\
+            -constr "{self.mapped_sdc_path}" \\
+            -liberty "{self.liberty_file}" \\
+            -script "/tools/B/nayiri/openroad/gcd/sky130hd/abc_speed.script" \\
```
harrisonliew, line 304: What is this file?

src/hammer-vlsi/synthesis/yosys/__init__.py
```diff
@@ -2,20 +2,34 @@
...
 from hammer_logging import HammerVLSILogging
-from hammer_utils import deepdict
-from hammer_vlsi import HammerToolStep
-from hammer_vlsi.vendor import OpenROADSynthesisTool
+from hammer_vlsi import MMMCCornerType
+import hammer_tech
 
+from typing import Dict, List, Any, Optional
 
-class YosysSynth(OpenROADSynthesisTool):
```
harrisonliew, line 18: Why is this not an `OpenROADSynthesisTool`? Is that class conflicting/need to be deprecated?

src/hammer-vlsi/technology/sky130/sky130.tech.json
```diff
@@ -16,7 +16,7 @@
       "base var": ""
     }
   ],
-  "layer map file": "$SKY130_NDA/s8/V2.0.1/VirtuosoOA/libs/technology_library/technology_library.layermap",
+  "cadence layer map file": "$SKY130_NDA/s8/V2.0.1/VirtuosoOA/libs/technology_library/technology_library.layermap",
```
harrisonliew, line 19: Changing this key would break existing tech plugins? Also I don't see it changed in the schema or why this change is necessary.

harrisonliew (changes requested):

harrisonliew:
One additional thing: since you have used f-strings, we will need to drop support for Python 3.4 & 3.5 (I think this is OK now since Python 3.6 is over 5 years old). This needs to be reflected in the System Requirements section of the documentation.

harrisonliew:
Also, `src/test/mypy.sh` needs to type-check all of these new plugins.

src/hammer-vlsi/synthesis/yosys/__init__.py
nayiri-k: It defines internal mapping of logic cells (see [here](https://github.com/YosysHQ/yosys/blob/master/techlibs/common/techmap.v)).
This path was manually added because the yosys binary couldn't find it, but since then our yosys install was fixed and we no longer need this.

src/hammer-vlsi/synthesis/yosys/__init__.py
nayiri-k: OpenROADSynthesisTool implements routines for the old way we ran OpenROAD (via Makefile), so it didn't have any useful routines and caused an error when I tried having YosysSynth inherit from it. We could probably just remove it.

src/hammer-vlsi/technology/sky130/sky130.tech.json
nayiri-k: This was probably residual from when I was figuring out where a klayout file should be specified. Will change it back.

src/hammer-tech/specialcells.py
```diff
@@ -31,7 +33,9 @@ def __mapping(cls) -> Dict[str, "CellType"]:
             "iofiller": CellType.IOFiller,
             "stdfiller": CellType.StdFiller,
             "decap": CellType.Decap,
-            "tapcell": CellType.TapCell
+            "tapcell": CellType.TapCell,
+            "buffer": CellType.Buffer,
```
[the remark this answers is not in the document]
nayiri-k: The buffer cell is set as the driving cell in the SDC file. OpenROAD doesn't seem to have an option to specify hold buffers during PnR so it's not reused for this. Could rename this to DriverCell or something similar as well...

src/hammer-vlsi/hammer_vlsi/vendor/openroad.py
```diff
@@ -40,13 +41,16 @@ def env_vars(self) -> Dict[str, str]:
         return reduce(add_dicts, [dict(super().env_vars)] + list_of_vars, {})
 
     def validate_openroad_installation(self) -> None:
+        # TODO: why does OPENROAD need to be an env var??
```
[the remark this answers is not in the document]
nayiri-k: OpenROAD doesn't seem to have an install convention for its tools, so I wasn't sure what purpose an `OPENROAD` environment variable would serve. For `CADENCE_HOME` it makes more sense, because the default binary path for various tools is set based on this variable + the tool version.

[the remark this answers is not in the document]
harrisonliew: Oh, I also missed the fact that you're not checking the `OPENROAD` env var, just the `openroad` and `klayout` executables. Nevermind, except remove excess code.

src/hammer-vlsi/synthesis/yosys/__init__.py
nayiri-k: Unclear, but I agree. Just left the hard-coded value from the OpenROAD+sky130 flow in the interest of time.

src/hammer-vlsi/technology/sky130/__init__.py
```diff
@@ -66,60 +68,95 @@ def setup_verilog(self) -> None:
...
+        lef_files = self.read_libs([
+            hammer_tech.filters.lef_filter
+        ], hammer_tech.HammerTechnologyUtils.to_plain_item)
+
+        for i,lef_file in enumerate(lef_files):
+            # skip standard cell lef file (klayout gives an error that cells already exist)
+            # TODO: very hacky to assume std cell lef is in position 1, is there a way to only extract tech lef + macro lefs?
+            if i == 1: continue 
+            insert_lines += f"<lef-files>{lef_file}</lef-files>\n"
+
```
nayiri-k, line 150: Is there a clean way to only extract the tech lef + macro lefs?

src/hammer-vlsi/par/openroad/__init__.py
```diff
@@ -80,152 +193,1177 @@ def export_config_outputs(self) -> Dict[str, Any]:
...
+        """)
+        return True
+
+    def set_rc(self) -> bool:
+        self.block_append(f"""
+        ################################################################
+        # Repair max slew/cap/fanout violations and normalize slews
+        source "/tools/B/nayiri/openroad/hammer-openroad/src/hammer-vlsi/technology/sky130/extra/sky130hd.rc"
+        set_wire_rc -signal -layer "met2"
+        set_wire_rc -clock  -layer "met5"
```
harrisonliew, line 726: This particular method should come from the Sky130 plugin as written. Maybe solve this dependency in a future PR.

src/hammer-vlsi/par/openroad/__init__.py
```diff
@@ -80,152 +193,1177 @@ def export_config_outputs(self) -> Dict[str, Any]:
...
+
+        self.block_append(f"""
+        ################################################################
+        # Floorplan Design
+
+        # Init floorplan/Place Macros
+        source -echo -verbose {floorplan_tcl}
+
+        # TODO: see pg 44 abt how these are generated, use self.generate_make_tracks()
+        source /tools/B/nayiri/openroad/gcd/sky130hd/sky130hd.tracks
```
harrisonliew, line 407: Fix absolute path here

src/hammer-vlsi/technology/sky130/__init__.py
harrisonliew: Do something similar to this: https://github.com/ucb-bar/hammer-cadence-plugins/blob/master/power/voltus/__init__.py#L112-L124

src/hammer-vlsi/par/openroad/__init__.py
```diff
@@ -80,152 +193,1177 @@ def export_config_outputs(self) -> Dict[str, Any]:
...
+        # write gds
+        exec klayout -zz \\
+                -rd design_name={self.top_module} \\
+                -rd in_def={self.output_def_filename} \\
+                -rd in_files={" ".join(gds_files)} \\
+                -rd seal_file= \\
+                -rd tech_file={klayout_techfile} \\
+                -rd config_file= \\
+                -rd out_file={self.output_gds_filename} \\
+                -rm /tools/B/nayiri/openroad/hammer-openroad/src/hammer-vlsi/technology/sky130/extra/def2stream.py
```
harrisonliew, line 1008: remove absolute path

src/hammer-vlsi/par/openroad/__init__.py
```diff
@@ -84,148 +181,959 @@ def tool_config_prefix(self) -> str:
...
+        self.block_append("""
+        ################################################################
+        # Global placement
+        """)
+        metals=self.get_stackup().metals[1:]
+        for metal in metals:
+            self.append(f"set_global_routing_layer_adjustment {metal.name} 0.5")
+        self.block_append(f"""
+        set_routing_layers -signal {metals[0].name}-{metals[-1].name} -clock met3-met5
+        set_macro_extension 2
```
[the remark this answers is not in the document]
nayiri-k: this was hard-coded in the openroad flow but it seems to correspond to the same thing

src/hammer-vlsi/par/openroad/__init__.py
```diff
@@ -84,148 +181,959 @@ def tool_config_prefix(self) -> str:
...
+                if (layer.direction==RoutingDirection.Vertical)   and (pin_layer_name not in ver_layers):
+                    ver_layers.append(pin_layer_name)
+        return f"place_pins {random_arg} -hor_layers {{{' '.join(hor_layers)}}} -ver_layers {{{' '.join(ver_layers)}}}"
+
     def place_pins(self) -> bool:
-        # TODO: currently using OpenROAD's default floorplan script
+        self.block_append(f"""
+        ################################################################
+        # IO Placement
+        {self.place_pins_tcl()}
```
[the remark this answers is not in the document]
nayiri-k: in the openroad flow they do it twice, once with the -random flag, second time without. unsure why

src/hammer-vlsi/par/openroad/__init__.py
```diff
@@ -84,148 +181,959 @@ def tool_config_prefix(self) -> str:
...
+
+        write_def "{self.run_dir}/{self.top_module}_global_place.def"
```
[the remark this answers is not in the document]
nayiri-k: As far as I could tell...but there is very little documentation about how to write a GDS from openroad, all of the flow scripts I've seen do it this one way (DEFs+klayout+custom python script). Would be something to ask the OpenROAD folks.

src/hammer-vlsi/par/openroad/__init__.py
nayiri-k: took these out, they were just in the original flow

src/hammer-vlsi/par/openroad/__init__.py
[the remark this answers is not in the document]
harrisonliew: Let's set it to `par.blockage_spacing` based on what the command seems to imply.

src/hammer-vlsi/par/openroad/__init__.py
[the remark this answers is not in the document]
harrisonliew: Oh, maybe it's needed for the early router.

src/hammer-vlsi/par/openroad/__init__.py
```diff
@@ -351,6 +351,9 @@ def init_design(self) -> bool:
         # set up useful variables
         clock_port = self.get_clock_ports()[0]
         self.clock_port_name = clock_port.name
+        tech_base_dir=self.get_setting('vlsi.core.technology_path')[0]
```
harrisonliew, line 354: This should probably be indexed to [-1]. Oftentimes for tech plugins not in Hammer, the `technology_path` key is `_meta: append` -ed to.

src/hammer-vlsi/par/openroad/__init__.py
```diff
@@ -25,41 +35,149 @@ def steps(self) -> List[HammerToolStep]:
...
+
+    def filter_for_extra_libs(self, lib: hammer_tech.Library) -> bool:
+        return lib in list(map(lambda el: el.store_into_library(), self.technology.get_extra_libraries()))
+
+    @property
+    def fill_cells(self) -> str:
+        stdfillers = self.technology.get_special_cell_by_type(CellType.StdFiller)
+        return ' '.join(list(map(lambda c: str(c), stdfillers[0].name)))
+
+    def block_append(self,commands,file=None) -> bool:
```
harrisonliew, line 164: For future PR: this would be nice to have in HammerTool.

src/hammer-vlsi/par/openroad/__init__.py
```diff
@@ -80,152 +197,1018 @@ def export_config_outputs(self) -> Dict[str, Any]:
...
+            {' '.join(self.create_power_straps_tcl())}
+            {add_pdn_connect_tcl}
+
+            ####################################
+            # macro grids
+            ####################################
+            ####################################
+            # grid for: CORE_macro_grid_1
+            ####################################
+            define_pdn_grid -name {{CORE_macro_grid_1}} -voltage_domains {{CORE}} -macro -orient {{R0 R180 MX MY}} -halo {{2.0 2.0 2.0 2.0}} -grid_over_boundary
```
harrisonliew, line 550: This 2.0 also seems to be equiv to `par.blockage_spacing`

src/hammer-vlsi/par/openroad/__init__.py
```diff
@@ -80,152 +197,1018 @@ def export_config_outputs(self) -> Dict[str, Any]:
...
+    def write_netlist(self) -> bool:
+        self.append(f"write_verilog -include_pwr_gnd -remove_cells {{ {self.fill_cells} }} {self.output_netlist_filename}")
+        return True
+
+    # Copy and hack the klayout techfile, to add all required LEFs
+    def setup_klayout_techfile(self) -> bool:
+        setting_dir = self.get_setting("technology.sky130.sky130A")
+        setting_dir = Path(setting_dir)
+
+        source_path = setting_dir / 'libs.tech' / 'klayout' / 'sky130A.lyt'
```
harrisonliew, line 857: This source_path seems sky130 specific - but isn't it specified in the tech json?

src/hammer-vlsi/par/openroad/__init__.py
```diff
@@ -80,152 +197,1018 @@ def export_config_outputs(self) -> Dict[str, Any]:
...
+        The layer is set by technology.core.std_cell_rail_layer, which should be the highest metal layer in the std cell rails.
+
+        :param bbox: The optional (2N)-point bounding box of the area to generate straps. By default the entire core area is used.
+        :param nets: A list of power net names (e.g. ["VDD", "VSS"]). Currently only two are supported.
+        :return: A list of TCL commands that will generate power straps on rails.
+        """
+        layer_name = self.get_setting("technology.core.std_cell_rail_layer")
+        layer = self.get_stackup().get_metal(layer_name)
+        tcl=[]
+        tcl.append(f"add_pdn_stripe -grid {{grid}} -layer {{{layer.name}}} -width {0.48} -pitch {5.44} -offset 0 -followpins\n")
```
harrisonliew, line 1123: Can this width/pitch be gleaned from the stackup?

src/hammer-vlsi/par/openroad/__init__.py
```diff
@@ -84,148 +181,959 @@ def tool_config_prefix(self) -> str:
...
+                corner_names.append(corner_name)
+            
+            self.append(f"define_corners {' '.join(corner_names)}")
+            for corner,corner_name in zip(corners,corner_names):
+                lib_files=self.get_timing_libs(corner)
+                for lib_file in lib_files.split():
+                    self.verbose_append(f"read_liberty -corner {corner_name} {lib_file}")
+        self.append("")
+
+    def convert_units(self,prefix) -> str:
```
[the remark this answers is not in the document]
nayiri-k: I renamed this function to clarify...but openroad errors when reading something like "1000ps" because it expects "1ns" (the error msg says it doesn't recognize the unit "00ps"). So the function scale_units_1000x_down (renamed from convert_units) scales this down to do the conversion. Maybe I could add this to units.py eventually, but it doesn't seem useful in any other scenarios.

src/hammer-vlsi/par/openroad/__init__.py
```diff
@@ -84,148 +181,959 @@ def tool_config_prefix(self) -> str:
...
+                {' '.join(global_connections_gnd)}
+            }}
+        }}
+        ##===> Power grid strategy
+        # Ensure pitches and offsets will make the stripes fall on track
+
+        pdngen::specify_grid stdcell {{
+            name grid
+            rails {{
+                met1 {{width 0.48 offset 0}}
```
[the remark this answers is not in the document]
nayiri-k: yes, I need figure out how to generate this automatically.

src/hammer-vlsi/par/openroad/__init__.py
nayiri-k: You're right, fixed.

src/hammer-vlsi/par/openroad/__init__.py
nayiri-k: I couldn't figure out how...this 0.48 value doesn't seem to appear anywhere (searched in the LEFs as well). I'll mark this down as another question to ask the openroad developers.

src/hammer-vlsi/par/openroad/__init__.py
nayiri-k: moved the file and made a relative path

harrisonliew (approved):
Pending CI passing, it's good to go!

PR #649, opened 2022-05-03
OpenROAD Issue Packaging

harrisonliew:
This packages the run environment in order to submit OpenROAD issues. It is very loosely based on this Make target in OpenROAD-flow-scripts: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/a2a0ee8a12/flow/util/utils.mk#L65-L185

Changed:
- `submit` now gets the return code of the tool execution and calls a tool-plugin-extensible `handle_errors` method if return code >0
- `par.openroad.create_archive_mode` can be set to `after_error` (default), `none`, `always`, or `latest_run`
  - `after_error` - if OpenROAD errors, create an archive of the run
  - `always` - create an archive after every par run (regardless of whether OpenROAD errors)
  - `latest_run` - create an archive of latest par run (don't run OpenROAD)
                   (useful if OpenROAD gets stuck on endless optimization iterations but never actually "errors")
  - `none` - never create an archive
- `handle_errors` calls `create_archive` if `par.openroad.create_archive_mode` == `after_error` or `always`

OpenROAD `create_archive` does the following:
1. Generates a issue directory based on the date/time, top module, OS platform
2. Dumps the logger output
3. Generates a `runme.sh` script
4. Copies `.tcl`, `.sdc`, `.pdn`, and `.lef` files in par-rundir to the issue dir
5. Copies all input Verilog to issue dir
6. Copies all input LEFs to issue dir (some may not be used since they are hacked in `read_lef()`)
7. Copies all input LIBs to issue dir
8. Hacks copied `par.tcl` script to remove all abspaths
9. Hacks copied `par.tcl` script to comment out `write_gds` block w/ KLayout
10. Tars up issue dir

TODOs:
- [ ] Determine how OpenROAD errors out (do we need to parse trace, or is return code enough?)
- [x] Provide mechanism for user to manually run `handle_errors` without OpenROAD erroring
- [ ] Provide mechanism to selectively exclude files in the case of NDA PDKs. Will need to split up `par.tcl` to do this.


nayiri-k (approved):

harrisonliew (commented):
Thanks for finishing this off!

Issue #650, opened 2022-05-06
How to change clock_period in sim_input.yml


Waxpple:
Hi,

After I change the config in `example-asap7.yml`
```
...
vlsi.inputs.clocks: [
 {name: "clock_clock", period: "1.25ns", uncertainty: "0.1ns"}
]
```
The design sdc and `sim-syn-input.json` are using 1.25ns setting without problem but the sim-input.yml stays at the 1ns setting.
 Where can I change the `sim-input.yml` to follow the `vlsi.inputs.clocks`?

Thanks!

harrisonliew:
This should be a Chipyard issue, but here's the answer.

`vlsi.inputs.clocks` is the clock period used for synthesis, and gets translated into a timing constraint. In contrast, the clock period in simulation is only dependent on the testbench, which in the case of Chipyard is via the `CLOCK_PERIOD` variable, defined here: https://github.com/ucb-bar/chipyard/blob/main/vcs.mk#L21.

Of course, matching the two clock periods only matters if you are doing timing-annotated gate-level simulations.

Waxpple:
Got it. Thanks!

Waxpple:
Hi,

1. Is there anything else that should be modified if I want to set the clock frequency at 800MHz?
I have changed the `vcs.mk` and `example-asap7.yml`. 
The synthesized design timing is clean (TNS >0) and functional simulation passed.
However, in timing-annotated gate-level simulations, the design failed with timing violation at `frontend.icache`.
And the critical path is not seen in the 
2. I read the issue on (https://github.com/ucb-bar/hammer/issues/644). Soon we can perform post-syn-power analysis? That is great! We can read the `saif` file and get the benchmark power analysis result. I hope I can be part of the PR if it is possible?
 
Thanks.

harrisonliew:
1. Ah, due to the fact that the ASAP7 SRAM models are completely fake, they have no timing annotation for gate-level simulation, and the LIBs have fake timing too. So, there cannot be an expectation for the timing-annotated simulations to pass if there are any dummy SRAMs in the design. In this case, you see a hold violation because in synthesis, the SRAMs have a positive clk-q delay while in simulation, this delay is 0.
2. Yes, the post-syn-power PRs are coming very soon. We'll make sure to tag you for review, even if you don't have Joules in case you want to comment.
[...]

PR #645, opened 2022-04-18
Fix ASAP7 SRAMs

harrisonliew:
Multiple fixes to ASAP7 to close #642, #643, and #644:
- Dual- -> single-cycle read
- No more high-Z output (but simultaneous read/write collision now not protected)
- Generate Verilog per SRAM rather than keeping it all in 1 file (helps with VCS FGP)
- Remove `define`s and hardcode widths from Verilog generation (helps with VCS FGP)
- Blocking -> non-blocking assignments in sequential block (bug)
- Handle >32-bit wide in randomization block (resolve lints)

@Waxpple and @Leon924 please try this with the most recent Synopsys plugins version!

PR #653, opened 2022-05-19
No BYTEMASK for SRAM1RW256x64/128, Change SITE in LEFs

harrisonliew:
From the discussion in #650:

1. These 2 dummy ASAP7 SRAM macros have an extra unused BYTEMASK port, which causes post-syn simulation to fail if these macros are used. The Verilog model originally had it, but was taken out in #645. This port isn't supported in the SRAM cache description anyway, so the views have been regenerated without them.
2. Site in ASAP7 r1p7 changed to asap7sc7p5t but was not updated in SRAM LEFs.

@Waxpple please try this out and see if the issue is resolved.


nayiri-k (approved):

Waxpple:
LGTM.

PR #654, opened 2022-05-19
Add input ilms to par output_ilms for top-level in hierarchical flow

jerryz123:
Top level par should "output" ilms so lvs can see all the schematics


harrisonliew (commented):
Based on offline discussion - is there an impact on hierarchical flow with >2 levels with this (there'd be excess ILMs going into syn)? Should `lvs.inputs.ilms` be modified instead, or do you think we just merge this for now/fix later?

PR #651, opened 2022-05-06
Formal Verification (LEC at first)

harrisonliew:
Adds a "formal" action (closes #329). At first, only logical equivalence checking is implemented, though many other formal verification checks can be added as the need arises.

PR #655, opened 2022-05-21
add type-checking to configuration files

bdngo:
Added rudimentary type-checking to configuration files. The type-checker supports the following settings:
* primitive types (e.g. `int`, `str`)
* collections (e.g. `list`)
* lists of dictionaries (e.g. `list[dict[str, str]]`)
* optional types (e.g. `Optional[str]`)

Anyone can bolster a configuration file with types by adding a YAML file containing the same keys as in the original config file, but it now contains the _type_ of the key instead. This means that the type-checker is backwards-compatible and is optional for existing configurations.


harrisonliew:
@edwardcwang what is the way to update buildbot to check against different versions of Python? I don't believe I have the privileges to change this. I think supporting f-strings is very nice to have now, so supporting Python 3.5-3.10 should be the default.

@bdngo when this is completed, can you also update the System Requirements section of the documentation to reflect this?

src/hammer-vlsi/defaults_types.yml
```diff
@@ -0,0 +1,982 @@
+# Core configuration options and defaults for hammer-vlsi.
+# The values specified in this file are the defaults.
+# e.g. foo: "bar" in this file means that the default setting for foo is "bar".
+
+vlsi.core:
+  # Path(s) to out-of-tree hammer technology libraries.
+  technology_path: list[str]
+  technology_path_meta: str
```
harrisonliew, line 8: Hm, since `meta` applies to every key, is there a way to not have to say that they are always `str`s?

edwardcwang:
@harrisonliew See messages

src/hammer-vlsi/defaults_types.yml
```diff
@@ -0,0 +1,982 @@
+# Core configuration options and defaults for hammer-vlsi.
+# The values specified in this file are the defaults.
+# e.g. foo: "bar" in this file means that the default setting for foo is "bar".
+
+vlsi.core:
+  # Path(s) to out-of-tree hammer technology libraries.
```
harrisonliew, line 6: In general, I don't think it's necessary to mirror all the comments, especially all the allowable values/TODOs in this types file.

src/hammer_config_test/test.py
```diff
@@ -935,5 +935,156 @@ def test_meta_deepsubst_subst(self) -> None:
...
+        db.update_core([base])
+        db.update_types(base_types)
+        
+        self.assertEqual(db.get_setting("foo.bar.adc"), 3.14)
+        with self.assertRaises(TypeError):
+            db.get_setting("foo.bar.quu")
+
+    def test_default_types(self) -> None:
+        """
+        Test that all HAMMER defaults are properly type-checked.
```
harrisonliew, line 1071: This is great to have for all default keys in Hammer. Can this be made into a method inheritable by all `HammerTool`s and `HammerTechnology`s so that we can auto check in plugin repos? If so, this should also just generate a warning if any `defaults_types.yml` is not found in conjunction with a `defaults.yml` file telling the user to implement it.

harrisonliew (changes requested):
Thank you so much for your first PR!!! This is awesome!

Besides a few things, there needs to be directions added to the documentation about checking against `defaults_types.yml` in addition to `defaults.yml`, as well as guidance when making new plugins that they should also have accompanying types.

src/hammer-vlsi/defaults_types.yml
bdngo: I was thinking we could actually use a subset of meta-actions for the types, since a lot of keys are reused, such as `vlsi.submit`. We could extend this to the meta-actions as well.

harrisonliew: I'm not sure I understand... do you mean that you can use what the meta actions support as input types to determine if a key's meta action is valid? Then you wouldn't need to say that the meta key is a `str`?

bdngo: Yes, that was what I was thinking. For example, this is what is currently in the [`synthesis.submit`](https://github.com/ucb-bar/hammer/blob/master/src/hammer-vlsi/defaults.yml#L538-L542) key in `defaults.yml`:

```yaml
synthesis.submit:
  command: "${vlsi.submit.command}"
  command_meta: lazysubst
  settings_meta: lazycrossref
```

In this suggestion, that entire block would remain unchanged between `defaults.yml` and `defaults_types.yml` and we would only need to replace the relevant substituted blocks with the corresponding types. This would give the meta-actions a use within the data type configurations, so we don't have to say they are always `str`. It would also mean more flexible configurations, since a change to a root type would mean we wouldn't have to change it across all possible children.

harrisonliew: I see. That is indeed one way of doing it. I don't think I understand this though: 
>It would also mean more flexible configurations, since a change to a root type would mean we wouldn't have to change it across all possible children.

Regardless, I'm really wondering whether all metas can be omitted from `defaults_types.yml` entirely? In my mind, having the type annotated for the key itself already limits what meta keys are applicable, so there's no real point having the meta keys in the type config too.

bdngo: That was actually my first thought when I created `defaults_types.yml`, as to whether to annotate the meta keys or not. It is a bit redundant given information about the key itself as well as the already limited set of meta actions. Looking at the way configs are combined, meta keys are not even considered since they are not a part of the configuration when calling  `get_config` I will remove them from the types.

src/hammer_config_test/test.py
bdngo: Could you elaborate on this a little more? Should we check types as soon as they are updated into a given `HammerDatabase`?

harrisonliew: Oh, I see, currently this is only run when you do unit tests, right? If so, yes, I think the types should be checked whenever a `HammerDatabase` is created/updated.

doc/Hammer-Basics/Hammer-Setup.rst
```diff
@@ -5,11 +5,7 @@ Hammer has a few requirements and there are several environment variables to set
 
 System Requirements
 -----------------------------
-- Python 3.6+ recommended (minimum Python 3.3+)
-
--- For Python 3.4 and lower, the ``typing`` module must be installed. (``python3 -m pip install typing``)
-
--- For Python 3.4, the enum34 package must be installed. (``python3 -m pip install enum34``)
+- Python 3.6+ recommended (minimum Python 3.5+)
```
harrisonliew, line 8: Ah, oops I made a typo in my comment, f-strings are only supported for Python 3.6+. #656 actually updates this file so you might as well remove this change and wait to merge from master after that PR goes through.

doc/Hammer-Use/Hammer-Config.rst
```diff
@@ -211,6 +211,20 @@ Common meta actions
...
+Type Checking
+-------------
+
+Any existing configuration file can and should be accompanied with a corresponding configuration types file.
+This allows for static type checking of any key when calling ``get_setting``.
+The file should contain the same keys as the corresponding configuration file, but can contain the following as values:
+
+- primitive types (``int``, ``str``, etc.)
+- collection types (``list``)
+- collections of key-value pairs (``list[dict[str, str]]``) to ensure compatibility with custom constraints
```
harrisonliew, line 223: I think that this needs to be a bit clearer - "to ensure compatibility" doesn't seem like enough. I think we'd need to say like "these dicts are to be translated into custom constraints (PlacementConstraint, PinAssignment, etc.) but are not themselves checked for types within the key-value pairs. See `src/hammer-vlsi/hammer_vlsi/constraints.py` for the supported custom constraints to ensure that you are specifying them correctly.

bdngo: Custom constraints are either `list[dict[str, str]]` or `list[dict[str, list]]`. The first accounts for most custom constraints, and the second accounts specifically for manual placement constraints and other hierarchical constraints. Either way, it's difficult to test comprehensively since the defaults are usually empty.

src/hammer-vlsi/defaults_types.yml
```diff
@@ -0,0 +1,572 @@
...
+
+  # SRAM Generator tool to use.
+  sram_generator_tool: str
+
+  # Sim tool to use.
+  sim_tool: str
+
+  # Power analysis tool to use.
+  power_tool: str
+
```
harrisonliew, line 70: Ack, it looks like some keys that I added in #651 aren't here because you started this before that was merged (e.g. `formal_tool` would be right here). Sorry - can you make one more run-through `defaults.yml` to get the new keys?

harrisonliew (approved):
Thanks!!!

PR #656, opened 2022-05-21
Update pyyaml to support Python 3.10

edwardcwang:
Old pyyaml uses dead code in 3.10

File "/buildbot/run_unittests/build/src/tools/pyyaml/lib3/yaml/constructor.py", line 126, in construct_mapping
    if not isinstance(key, collections.Hashable):
AttributeError: module 'collections' has no attribute 'Hashable'


.gitmodules
```diff
@@ -11,3 +8,7 @@
 [submodule "src/tools/gdspy"]
 	path = src/tools/gdspy
 	url = https://github.com/heitzmann/gdspy.git
+[submodule "src/tools/pyyaml"]
+	path = src/tools/pyyaml
+	url = https://github.com/yaml/pyyaml.git
+	branch = release/6.1
```
harrisonliew, line 14: Question: if we are now able to use the release versions of pyyaml, does it still need to be a submodule, or can we just have it be a system requirement in the documentation/part of a dependency install script? This would get out of date pretty fast.

src/hammer-vlsi/technology/asap7/__init__.py
```diff
@@ -244,13 +244,16 @@ def asap7_scale_final_gds(ht: HammerTool) -> bool:
     Scale the final GDS by a factor of 4
     scale_gds_script writes the actual Python script to execute from the Tcl interpreter
     """
+    # This is from the Cadence tools
+    cadence_output_gds_name = ht.output_gds_filename  # type: ignore
```
harrisonliew, line 248: Ugh, this isn't elegant. Any proposal for how not to have to have these extra lines?

harrisonliew (commented):
LGTM, but approval pending on pyyaml question.

.gitmodules
edwardcwang: That's also fine with me if you'd prefer

src/hammer-vlsi/technology/asap7/__init__.py
edwardcwang: Hmm. One possible resolution is to add `output_gds_filename` into the core Hammer P&R abstraction. Currently this is actually problematic from a type-checker perspective, since it has no idea of knowing that `output_gds_filename` is valid because that is defined in CadenceTool, not core Hammer.

.gitmodules
harrisonliew: yeah let's do that and only submodule dependencies that are not released or need specific versions.

src/test/mypy.sh
```diff
@@ -9,7 +9,7 @@ err=0
 
 call_mypy () {
     >&2 echo "Running mypy $*"
-    output=$(mypy "$@" | grep -v "python-jsonschema-objects" | grep -v TechJSON | grep -v installs | grep -v tarballs | grep -v ccs | grep -v nldm | grep -v supplies | grep -v lef_file | grep -v qrc_techfile | grep -v serialize | grep -v "hammer_tech.Library" | grep -v milkyway_ | grep -v tluplus | grep -v jsonschema | grep -v "pyyaml/" | grep -v provides | grep -v \"libraries\" || true)
+    output=$(mypy --no-error-summary "$@" | grep -v "python-jsonschema-objects" | grep -v TechJSON | grep -v installs | grep -v tarballs | grep -v ccs | grep -v nldm | grep -v supplies | grep -v lef_file | grep -v qrc_techfile | grep -v serialize | grep -v "hammer_tech.Library" | grep -v milkyway_ | grep -v tluplus | grep -v jsonschema | grep -v "pyyaml/" | grep -v provides | grep -v \"libraries\" | grep -v "note: See" || true)
```
harrisonliew, line 12: now that you removed pyyaml submodule would the "pyyaml/" fail?

harrisonliew (approved):
Thanks!

PR #658, opened 2022-05-26
Add specify block in SRAM compiler for vcs sdf back-annotation

Waxpple:
Add specify block in SRAM compiler for vcs sdf back-annotation.


harrisonliew (approved):
Thanks for getting it working. Looks like CI had a network error, but I'll merge this anyway since it doesn't affect the tests.

PR #661, opened 2022-06-16
Don't print output config to stdout

harrisonliew:
@odxa20 reported in ucb-bar/chipyard#1173 that an X-to-Y step caused an BlockingIOError.

Replacing printing the entire output config to stdout with just a message pointing to where the output config was written solves this + makes Hammer much less verbose.


harrisonliew:
Taking the 2 thumbs up as approvals 😏

PR #662, opened 2022-06-18
Update asap7 defaults.yml to conform with type checks

odxa20:
Added quotation marks around the value of key core.node to comply with recenlty implemented type checks. Without this modification the following error arises TypeError: Expected primary type str for vlsi.core.node, got type int


odxa20:
Other solution is to change the expected type in `src/hammer-vlsi/defaults_types.yml` to int

odxa20:
Since the core.node value in both openroad and sky130 defaults.yml files are ints I think that the best solution is to change the expected type in src/hammer-vlsi/defaults_types.yml to int so I am closing this PR and proposing this change in a new one instead

harrisonliew:
See comment in #663. I'm going to close this one.

PR #663, opened 2022-06-18
Update defaults_types.yml to fix type error for vlsi.core.node

odxa20:
Changed default type of vlsi.core.node to int from str since in all the technology files `defaults.yml` the value is an int. Without this change the following error arises `Expected primary type str for vlsi.core.node, got type int`


harrisonliew:
I believe this is the correct PR, instead of #662, since `node` should be an `int`, not a `str`. Can I reopen this and merge it instead of #662?

harrisonliew:
Sure. Can you include a change to that default key value as part of this PR? You can make it 0.

odxa20:
Done :)

harrisonliew (approved):
Thanks!

PR #664, opened 2022-06-24
Fix small error in tech defaults

edwardcwang:
`vlsi.core.technology` is a str and `vlsi.core.node` is an int


harrisonliew (approved):
D'oh, how did I miss that...

PR #673, opened 2022-08-11
update asap7 std cell library version 28

Waxpple:
1. Update asap7 std cell library version 28
2. Drop don't use ICG*DC* cell. The LEF file of this cell is divisible by the site width in version 28.
There is a known issue in the version 28 std cell library which is the SEQ_SRAM Verilog file issue.
I have created a PR. See https://github.com/The-OpenROAD-Project/asap7/pull/33


raphaelrowley:
Why was this not merged?

PR #675, opened 2022-08-20
Refactor Hammer for PyPI Publishing

vighneshiyer:
This is still a work in progress. The goals are:

- To be able to publish hammer to PyPI so tool and PDK plugins (and Chipyard) can depend on the published artifact
- Use poetry to set up a consistent development environment
- Get rid of all the PYTHONPATH hacks and the need to set HAMMER_VLSI
- Unify the testing scripts so a normal pytest invocation can be used
- Make CI scripts easy to write
- To the minimal extent possible, migrate to using pydantic models in lieu of jsonschema dynamic objects or NamedTuples

The things done so far are:

- Unify the hammer namespace by migrating all subprojects inside a top-level package called hammer
    - `hammer-vlsi/hammer_vlsi` -> `hammer.vlsi`
    - `hammer-vlsi/technology` -> `hammer.technology`
    - `hammer-vlsi/lvs` -> `hammer.lvs` (and so forth)
    - `hammer-tech` -> `hammer.tech`
    - `hammer-config` -> `hammer.config`
    - `hammer-shell` -> `hammer.shell`
- Dependencies are now specified in a pyproject.toml using the latest upstream versions, all submodules removed, no sourceme or other magic required to run tests
- Tests moved into a top-level folder (some tests have been migrated already, I'm working on the others)

To test this out:

1. Checkout the 'packaging' branch
1. Install poetry (https://python-poetry.org/docs/master/#installing-with-the-official-installer)
1. `poetry install` to fetch dependencies and set up a virtualenv
1. `poetry shell` to pop into the virtualenv, run `hammer-vlsi` as usual to run the CLI top-level
1. `pytest tests` to run the tests
1. `poetry build` to build a wheel


edwardcwang:
Hey Vighnesh long time no see and nice to hear from you again! 1) Will this keep the `hammer-vlsi` CLI executable (which is quite useful)? 2) Will it be required to use `poetry` or pypi or will it still be possible to depend on hammer by source?

vighneshiyer:
Hey Edward! Thanks looking over this.

1) Yes, `hammer-vlsi` is CLI executable when `hammer` is installed as a package. These 'installable' scripts are specified in `pyproject.toml`

```toml
[tool.poetry.scripts]
hammer-generate-mdf = "hammer.shell.hammer_generate_mdf:main"
get-config = "hammer.shell.get_config:main"
hammer-vlsi = "hammer.shell.hammer_vlsi:main"
```

2) You can always depend on hammer by source just by cloning it somewhere and adding `hammer/hammer` to your `PYTHONPATH`. Of course, the nicer way to do this is to install hammer as an 'editable local package' using `pip install -e` and that will allow you to use hammer in any virtualenv (or conda env) you want while being able to edit its source files and seeing the changes reflect immediately.

dan-fritchman:
Big time support this, great work taking it on. 

A note on one of your comments: 

`Of course, the nicer way to do this is to install hammer as an 'editable local package' using pip install -e and that will allow you to use hammer in any virtualenv (or conda env) you want while being able to edit its source files and seeing the changes reflect immediately. `

The “editable installs” don’t quite place nice with Poetry / pyproject.toml, at least just yet. Or at least they didn’t as of a few months ago, when we decided to “revert” a bunch of projects (Hdl21, everything Vlsir-related) from pyproject.toml to setup.py. I can look up the problem if helpful, but I recall it was not work-around-able. 

I too would prefer the TOML config; the editable installs, which we use heavily, are the only thing holding it back. 

Good news is any of the features you want, including the ”scripts” / installed entry points, will work with either TOML or setup.py.

vighneshiyer:
I thought that newer versions of pip are able to handle delegating builds to poetry properly and still support locally editable packages that are only specified via pyproject.toml: https://github.com/python-poetry/poetry/issues/34#issuecomment-1054626460

https://github.com/python-poetry/poetry/issues/34#issuecomment-1193365526

> poetry-core also supports [PEP 660](https://peps.python.org/pep-0660/) in the stable and beta versions, which allows for a Poetry-managed project to be installed in editable mode by any PEP 517 frontend (e.g. Poetry itself, pip, etc) that also supports PEP 660.

But I haven't tried this out yet myself; I'll try it out with the tool plugin packages once I get this PR done

dan-fritchman:
That would be great. I certainly had the impression they *intended* for it to work at the time. 

Looking back, it appears we abandoned Poetry in January: 
https://github.com/dan-fritchman/Hdl21/commit/afe44015c98e61cca84358b5a507efd8d5a2be22

And the commentary on the Poetry issue has all been since. Hopefully they’ve added this support.

vighneshiyer:
@dpgrubb13 @harrisonliew This is ready for "review". I know that reviewing these changes line-by-line is going to be difficult, so it is better if you can set up the dev environment yourself and verify it works.

The documentation is here (https://github.com/ucb-bar/hammer/blob/packaging/doc/Hammer-Basics/Hammer-Setup.md). It should walk you through the process of setting up the dev environment, running tests, typechecking, testing multiple Python versions, and building the docs.

I've published this branch to TestPyPI (https://test.pypi.org/project/hammer-vlsi/) just to demonstrate the flow works.

While people are verifying that the dev flow works fine, I will use the published hammer-vlsi artifact to begin porting over the tool plugin repos. @dpgrubb13 will set up a strawman hammer-example repo which will use the published artifacts and implement a simple rtl2gds flow with asap7 and/or sky130 and the cadence toolchain.

dpgrubb13:
I can confirm the dev flow works for me.

harrisonliew (changes requested):
The dev environment works for me great as well. Thanks so much for this major refactor!

Some things I noticed:

- Running the tests with pytest (as would be expected) generates the hammer logs, temporary json files, and tech cache but does not clean them up. Can this be done for usability?
- Is the existence of `hammer/py.typed` the same hack as for ruamel?
- Is the dependence on the former https://github.com/edwardcwang/python-jsonschema-objects submodule removed? If so, a lot of existing tech JSONs will break due to removed support for keys with spaces. If we have to update all those technologies anyway, then maybe this isn't a dealbreaker?
- The optional gdstk dependency is confusing, since it's a C++ library with a Python wrapper. In the readme in https://github.com/heitzmann/gdstk, it's not clear whether the published Python package contains the built Python wrapper (which it says to install in Conda or build from source).  If it is, then we'll need to update the ASAP7 readme (https://github.com/ucb-bar/hammer/tree/packaging/hammer/technology/asap7) to reflect only needing to build the static C++ library.

vighneshiyer:
@harrisonliew 

1. Yes we should make sure the tests don't pollute the source directory. I need to hunt down where every emission comes from and make sure to redirect it by setting --obj_dir or running hammer inside a scratch test directory.
2. No, having py.typed in `hammer/hammer` is the right thing to do (https://peps.python.org/pep-0561/#packaging-type-information). I'm not sure why py.typed is not in the top-level ruamel.yaml package, but the hack works at least.
3. Yes that dependency is removed, and there is no need for json schema parsing and dynamic object construction anymore since TechJSON is just a Pydantic dataclass now (probably should rename TechJSON later when we do configuration as code). Yes, this means tech plugins need their json patched to convert spaces to underscores in keys. I've already done this for asap7 and it is pretty quick and easy so I anticipate it should be the same for other tech jsons.
4. The published gdstk includes the C++ library and the Python bindings together. The C++ .so needs to be built from source and the instructions in the asap7 README still apply (on BWRC machines we will have to compile gdstk manually with the documented hacks). I can clean up the asap7 readme with specific instructions.

harrisonliew (approved):

PR #678, opened 2022-09-30
Key description

bdngo:
This introduces a new action `info`. It spawns a tool that a user can query any default keys with. It spawns a "shell" that prompts the user for a key one level at a time until reaching a terminal key with corresponding value. The value from the project configuration is displayed along with a description of the key. In order to accommodate this change, any comments associated with a key in any configuration `.yml` files must be started immediately after a value, on the same line. Additional comments can be added on new lines.


harrisonliew (commented):
Awesome! Just a few things before I review in more detail:

1. Can you rebase on master? There are new keys in the asap7/sky130/yosys/openroad defaults, sorry!
2. General question about ruamel.yaml comment parsing: does this also work?
```
foo: bar
# this is a comment
# this is another comment
```
3. It seems like all `_meta` keys are also included. We should probably discuss this offline, but philosophically, does it make sense to query the meta value if the value of the key it applies to already has all the meta actions applied?
4. As a continuation of the previous point, is it difficult to also print the key's history? That way we definitely don't need to query meta keys.

After reviewing the list of actions (`action_map`), I believe we need to make the CLI even easier to use. I think we can discuss offline about:
- A `-h`/`--help` cmd argument. Should at least print out the available cmd arguments + help text, as well as the available actions + what they do (derived from the docstrings)
- I'm envisioning this `info` action to also be available for all other actions, and would let you query a bit deeper into the flow. E.g. a `syn-info` action could let you query keys along with the resolved tool steps and output keys that the tool will set.

harrisonliew (approved):
Thanks!!!

PR #688, opened 2022-12-10
Fix some SRAM pin sizes/locations

### LINKED ISSUE #687: Wierd pin width of two SRAMs
The dummy SRAMs: SRAM1RW256x64 and SRAM1RW256x128 have wierd pin widths. These two SRAMs have pin width of 0.384, while other SRAMs have pin width of 0.096. Will re-generating the SRAMs with sram_compiler.py shrink the pin width down to 0.096? If yes, how can I call the sram_compiler?

I'm trying to run PAR on MegaBoomConfig with Innovus and the icache has multiple SRAM1RW256x64. Innovus makes several DRC violations around pins of this SRAM. There is no DRC violations around other SRAMs.

harrisonliew:
Closes #687

@idwwwoqq808 please test this out!


dpgrubb13 (approved):
LGTM

PR #692, opened 2022-12-30
Utility Function v2

kenhoberkeley:
A quick PR to merge a simple utility method `get_settings_from_dict` originating from Xcelium PR. Goal is to simplify/standardize the process of getting multiple keys (often specified as a list of keys on which to perform `get_setting`) at once and returning a unified dictionary. Currently, such functionality is handled in multiple slightly different ways across the plugins.

Accompanied now by a unit test.


harrisonliew (approved):
Thanks for the rebase!

PR #693, opened 2023-01-01
Add vendor-required headers for generated scripts

harrisonliew:
Repurposed unused `append_contents_to_path` method and added a new `header` property in `HammerTool` that specific vendors/tools can override to insert a standardized header.

Also removed a few remaining Python 3.8 dependencies.


dpgrubb13 (approved):
LGTM

PR #706, opened 2023-02-02
New Method For Getting Timing Libs

kenhoberkeley:
- Introduce a new method `get_timing_lib_with_preference` to allow users to select their timing lib of choice when multiple are available. This defaults to NLDM -> ECSM -> CCS. 
- New key `vlsi.technology.timing_lib_pref`.
- Backing unit test.

Note that the prior `get_timing_lib_with_ecsm_filter` is unaltered as it is used in yosys and sky plugins.

Cadence Changes here: https://github.com/ucb-bar/hammer-cadence-plugins/pull/88

### DIFF
--- a/hammer/config/defaults.yml
+++ b/hammer/config/defaults.yml
@@ -93,6 +93,10 @@ vlsi.technology:
   # If this is not specified, then the tarballs will be extracted to obj/<tech_dir>/extracted/.
 
   timing_lib_pref: "NLDM"
+  # Select a timing lib preference, available options include:
+  # NLDM, ECSM, and CCS (lower or upper case acceptable). 
+  # If no preference is specified, then the following preference order is followed:
+  # NLDM -> ECSM -> CCS
 
 # General VLSI inputs.
 # These will vary per run of hammer-vlsi.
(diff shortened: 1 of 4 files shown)


hammer/config/defaults.yml
```diff
@@ -92,6 +92,8 @@ vlsi.technology:
   # the extracted contents of foobar.tar.gz.
   # If this is not specified, then the tarballs will be extracted to obj/<tech_dir>/extracted/.
 
+  timing_lib_pref: "NLDM"
```
harrisonliew, line 95: Can you add some comments below to describe what this key does and the available options?

harrisonliew: Also, this key isn't used in `get_timing_lib_with_preference`?

harrisonliew (changes requested):

hammer/config/defaults.yml
kenhoberkeley: Re 1: Added
Re 2:  It accepts a preference argument and defaults to NLDM if none is provided.

harrisonliew (approved):

PR #711, opened 2023-02-28
add retiming key

dpgrubb13:
Adds simple key interface for basic retiming to be implemented in the tool plugins.


harrisonliew (approved):
LGTM

PR #733, opened 2023-03-09
Add missing keys to defaults

### LINKED ISSUE #716: Some keys missing from defaults
During some code searching, I found that the following keys are used in tool plugins but not defined (as optional) in defaults.yml:
* `vlsi.technology.tap_cell_interval`
* `vlsi.technology.tap_cell_offset`

Need to investigate if other keys are missing.

harrisonliew:
Exhaustively searched in plugins/Hammer core, only found these 2 keys to not be defined in defaults.

**Related PRs / Issues**

Closes #716

**Type of change**:
- [X] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [X] Change to core Hammer
- [ ] Other

**Contributor Checklist**:

### DIFF
--- a/hammer/config/defaults.yml
+++ b/hammer/config/defaults.yml
@@ -60,6 +60,12 @@ vlsi.technology:
   bump_block_cut_layer: null # Top cut/via layer for blockage under bumps. (Optional[str])
   # Only used if using vlsi.inputs.bumps
   # TODO: remove this after stackup supports vias ucb-bar/hammer#354
+  
+  tap_cell_interval: 10.0 # Spacing between columns of tap cells (float)
+  # Must be overridden by technology plugin defaults or else you will have DRC/latch-up errors.
+
+  tap_cell_offset: 0.0 # Offset of the first column of tape cells from the edge of the floorplan (float)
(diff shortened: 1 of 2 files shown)


hammer/config/defaults.yml
```diff
@@ -60,6 +60,12 @@ vlsi.technology:
   bump_block_cut_layer: null # Top cut/via layer for blockage under bumps. (Optional[str])
   # Only used if using vlsi.inputs.bumps
   # TODO: remove this after stackup supports vias ucb-bar/hammer#354
+  
+  tap_cell_interval: 10.0 # Spacing between columns of tap cells (float)
```
dpgrubb13, line 64: Is there any particular reason at this point we're using a mix of Decimal and float types for keys?

harrisonliew: This is not Decimal because it's not processed anywhere in Hammer - it's directly read by tool plugins and so remains as a float.

harrisonliew: But, I could see how mixing this in the defaults file could create some confusion. Pydantic would remove this.

dpgrubb13 (approved):

PR #775, opened 2023-03-30
Added UPF Support

derrickqi2003:
Added UPF support to hammer. Generates a single power domain UPF. It should do exactly what hasCPFsupport does. Tested it by running PAR on the tapeout vlsi lab and manually feeding into innovus with the innovus command below:
read_power_intent -1801 power_spec.upf
[power_spec.pdf](https://github.com/ucb-bar/hammer/files/11116390/power_spec.pdf)
.

**Related PRs / Issues**

**Type of change**:
- [ ] Bug fix
- [X] New feature
- [ ] Other enhancement

**Impact**:
- [X] Change to core Hammer
- [ ] Other

**Contributor Checklist**:

### DIFF
--- a/hammer/vlsi/hammer_vlsi_impl.py
+++ b/hammer/vlsi/hammer_vlsi_impl.py
@@ -2059,11 +2059,52 @@ class HammerTimingTool(HammerTool):
     ### END Generated interface HammerTimingTool ###
 
 class HasUPFSupport(HammerTool):
-    """Mix-in trait with functions useful for tools with UPF style power
-    constraints"""
-    @property
-    def upf_power_specification(self) -> str:
-        raise NotImplementedError("Automatic generation of UPF power specifications is not supported yet.")
+   """Mix-in trait with functions useful for tools with UPF style power constraints"""
+   @property
+   def upf_power_specification(self) -> str:
+        #Constants and objects
+        output = [] # type: List[str]
+        domain = "AO"
+        #Header
+        output.append("set_design_top {t}".format(t=self.top_module))
+        vdd = VoltageValue(self.get_setting("vlsi.inputs.supplies.VDD"))
+        #Create Single Power Domain
+        output.append("create_power_domain {d} \\".format(d=domain))
+        output.append("\t-elements {.}")
+        #Get Supply Nets
+        power_nets = self.get_all_power_nets() 
+        ground_nets = self.get_all_ground_nets()
+        #Create Supply Ports 
+        for pg_net in (power_nets+ground_nets):
+            if(pg_net.pin != None):
+                #Create Supply Nets
+                output.append("create_supply_net {p} -domain {d}".format(p=pg_net.name, d=domain))
+                output.append("create_supply_port {p} -domain {d} \\".format(p=pg_net.name, d=domain))
+                output.append("\t-direction in")
+                #Connect Supply Net
+                output.append("connect_supply_net {p} -ports {p}".format(p=pg_net.name))
+        #Set Domain Supply Net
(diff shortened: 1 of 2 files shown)


hammer/vlsi/hammer_vlsi_impl.py
```diff
@@ -2059,11 +2059,52 @@ def sdf_file(self, value: Optional[str]) -> None:
...
+        output.append("\t-primary_ground_net {g}".format(g=ground_nets[0].name))
+        #Add Port States 
+        for p_net in power_nets:
+            if(p_net.pin != None):
+                output.append("add_port_state {g} \\".format(g=p_net.name))
+                output.append("\t-state {{default {v}}}".format(v=vdd.value))
+        for g_net in ground_nets:
+            if(g_net.pin != None):
+                output.append("add_port_state {g} \\".format(g=g_net.name))
+                output.append("\t-state {default off}")
```
harrisonliew, line 2098: I don't think this is correct according to my understanding of the command... instead it should be a voltage of 0. "off" is distinctly a different state?

hammer/vlsi/hammer_vlsi_impl.py
```diff
@@ -2059,11 +2059,52 @@ def sdf_file(self, value: Optional[str]) -> None:
...
+        #Create Power State Table
+        output.append("create_pst pwr_state_table \\")
+        output.append("\t-supplies {{{p} {g}}}".format(p=" ".join(map(lambda x: x.name, power_nets)),g=" ".join(map(lambda x: x.name, ground_nets)) ))
+        #Add Power States
+        output.append("add_pst_state {d} \\".format(d="aon"))
+        output.append("\t-pst {pwr_state_table} \\")
+        output.append("\t-state {{{s}}}".format(s=" ".join(map(lambda x: "default", power_nets+ground_nets))))
```
harrisonliew, line 2105: This map doesn't take into account the names of the nets (get `{default default}`). Might want to do that (including above when the state name is made) for clarity.

harrisonliew (changes requested):
Thanks for this PR! A couple comments:

- It seems like these are UPF 2.0 commands. You might want to include a `upf_version` command in the header. Similarly, [this](https://github.com/ucb-bar/hammer/blob/0ec427c5a8f609b061c8c111d7797014e56c7d07/hammer/config/defaults.yml#L297) comment may need to reflect that
- Building on previous point, supporting UPF 3+ could be a switch? That option there could be like `upf_2.0` or `upf_3.1`, etc. if you desire.
- Is there any corresponding updates needed for the Cadence plugins?

kenhoberkeley:
Really nitpicky here but could we switch to f-strings.  Might just be nice to stick for future code.

hammer/vlsi/hammer_vlsi_impl.py
derrickqi2003: Screen Shot 2023-03-30 at 5 49 42 PM
It appears to be a special exception according to the 2018 manual, but I will change it to be safe.

hammer/vlsi/hammer_vlsi_impl.py
derrickqi2003: Screen Shot 2023-03-30 at 6 00 57 PM
This is the intended power state table generated by those upf commands.
I believe the names of the nets are accounted for in the create_pst_state table. I intended for this to mimic the nominal condition in CPF, thus why there's no specificity. Let me know if I should do it though.

derrickqi2003:
@kenhoberkeley Added fstrings
@harrisonliew Made the voltage change from "off" to 0.0 & updated the documentation to say IEEE 1801-2009

PR #782, opened 2023-05-01
KLayout DRC plugin

nayiri-k:
KLayout is the preferred DRC checker for sky130 according to efabless.
This plugin generates the same DRC errors as Calibre for both gcd and the Chipyard TinyRocket Config, in an easy-to-understand GUI.

**Related PRs / Issues**

**Type of change**:
- [ ] Bug fix
- [x] New feature
- [ ] Other enhancement

**Impact**:
- [ ] Change to core Hammer
- [x] Change to a Hammer plugin
- [ ] Other

**Contributor Checklist**:

### DIFF
--- a/hammer/drc/klayout/__init__.py
+++ b/hammer/drc/klayout/__init__.py
@@ -1,0 +1,104 @@
+# Klayout DRC plugin for Hammer
+#
+# See LICENSE for licence details.
+
+import os
+from textwrap import dedent as dd
+from typing import List, Optional, Dict, Any
+
+from hammer.logging import HammerVLSILogging
+from hammer.utils import deepdict
+from hammer.vlsi import HammerToolStep
+from hammer.vlsi import HammerDRCTool, TCLTool
+
+class Klayout(HammerDRCTool, TCLTool):
+
+    #=========================================================================
+    # overrides from parent classes
+    #=========================================================================
+    @property
+    def steps(self) -> List[HammerToolStep]:
+        return self.make_steps_from_methods([
+            # no steps
+        ])
+
+    def do_post_steps(self) -> bool:
+        assert super().do_post_steps()
+        return self.run_klayout()
+
+    @property
+    def env_vars(self) -> Dict[str, str]:
+        new_dict = deepdict(super().env_vars)
+        return new_dict
+
+    def fill_outputs(self) -> bool:
+        return True
+    
+    def tool_config_prefix(self) -> str:
+        return "drc.magic"
+
+    def drc_results_pre_waived(self) -> Dict[str, int]:
+        return {}
+
+    def globally_waived_drc_rules(self) -> List[str]:
+        return []
+
+    def version_number(self, version:str) -> int:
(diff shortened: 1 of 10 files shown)


hammer/drc/klayout/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+
+class Klayout(HammerDRCTool, TCLTool):
+
+    #=========================================================================
+    # overrides from parent classes
+    #=========================================================================
+    @property
+    def steps(self) -> List[HammerToolStep]:
+        return self.make_steps_from_methods([
+            # no steps
```
harrisonliew, line 22: Hm, this would prevent users from adding hooks... is there any situation where one would override or add things to the deck? Or, even though it would be different from similar plugins, could `run_klayout` be the singular step and we could leave some results post-processing to `do_post_steps`?

hammer/drc/klayout/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+    @property
+    def env_vars(self) -> Dict[str, str]:
+        new_dict = deepdict(super().env_vars)
+        return new_dict
+
+    def fill_outputs(self) -> bool:
+        return True
+    
+    def tool_config_prefix(self) -> str:
+        return "drc.magic"
```
harrisonliew, line 38: Wrong prefix

hammer/technology/sky130/__init__.py
```diff
@@ -46,13 +48,24 @@ def setup_cdl(self) -> None:
         os.makedirs(cache_tech_dir_path, exist_ok=True)
         dest_path = cache_tech_dir_path / f'{self.library_name}.cdl'
 
+        # device names expected in LVS decks
+        if (self.get_setting('vlsi.core.lvs_tool') == "hammer.lvs.calibre"):
+            pmos = 'phighvt'
+            nmos = 'nshort'
```
harrisonliew, line 54: Does Calibre nominally require using the NDA PDK, hence the need to rename the transistors? What about the other Vts?

hammer/drc/klayout/default_types.yml
```diff
@@ -0,0 +1,12 @@
+# Default types of all of the configuration options for Klayout's tools
+#   specified in defaults.yml
+drc.klayout:
+  klayout_bin: str
+  
+  version: str
+
+  generate_only: bool
+
+  technology_file: Optional[str]
```
harrisonliew, line 10: Seems like an extra key?

harrisonliew (changes requested):

hammer/technology/sky130/__init__.py
nayiri-k: For some reason the only PMOS devices in the CDL netlist are called "pfet_01v8_hvt" and the NMOS are "nfet_01v8", there's no other variations of these devices. But the Calibre LVS deck calls these devices phighvt/nshort.

harrisonliew (approved):

PR #813, opened 2023-10-22
Add flag to prevent power strap generation on stdcell rail layer

jerryz123:
In some technologies, the stdcell rail layer is handled by a special tech-specific hook. This gives us more flexibility in selecting which layers show up in the auto-generated power_straps.tcl

**Related PRs / Issues**

**Type of change**:
- [ ] Bug fix
- [x] New feature
- [ ] Other enhancement

**Impact**:
- [x] Change to core Hammer
- [ ] Other

**Contributor Checklist**:

### DIFF
--- a/hammer/config/defaults.yml
+++ b/hammer/config/defaults.yml
@@ -613,6 +613,9 @@ par:
       strap_layers: [] # List of layers on which to place power straps (std cell rail layer is implied - do not include it)
       # type: List[str]
 
+      generate_rail_layer: true # If true, generates straps on the stdcell rail layer - technology.core.std_cell_rail_layer
+      # type: bool
+
 # DRC settings
 drc.inputs:
 # DRC settings
--- a/hammer/config/defaults_types.yml
+++ b/hammer/config/defaults_types.yml
@@ -326,6 +326,10 @@ par:
       # type: list[str]
       strap_layers: list[str]
 
+      # Indicates if Hammer should generate straps for the stdcell rail layer
+      # type: bool
+      generate_rail_layer: bool
+
 # DRC settings
 drc.inputs:
   # Top RTL module.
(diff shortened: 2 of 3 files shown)


hammer/config/defaults.yml
```diff
@@ -613,6 +613,9 @@ par:
       strap_layers: [] # List of layers on which to place power straps (std cell rail layer is implied - do not include it)
       # type: List[str]
 
+      generate_rail_layer: true # If true, generates straps on the stdcell rail layer - technology.core.std_cell_rail_layer
```
harrisonliew, line 616: I'd prefer this to be one level up in the key hierarchy (i.e. `par.generate_power_straps_options.generate_rail_layer`). This is because the rail layer generation doesn't use any of the `by_tracks` calculations.

harrisonliew (changes requested):

hammer/config/defaults.yml
jerryz123: Makes sense. Following that reasoning,  should we also pull `self.specify_std_cell_power_straps` out of `specify_all_power_straps_by_tracks` into `generate_power_straps_tcl`?

harrisonliew: Yes, let's do that.

jerryz123: I see that the  fields of `generate_power_straps_options` are intended to be the valid options for `generate_power_straps_method`. Do you think its a good idea to add another field to `generate_power_Straps_options` that breaks that pattern? I don't think it would cause anything to break, but it does invalidate that comment.

```
  generate_power_straps_method: "by_tracks" # If power_straps_mode is 'generate', which method to use.
  # Currently, the valid options are:
  # - by_tracks - Specify the power strap plan per layer in terms of tracks

  generate_power_straps_options:
  # Default settings for power strap generation modes
  # Keys are the valid values of the 'generate_power_straps_method' key above
    by_tracks:
    ```

harrisonliew: Ah, I see. Yeah, let's not break that pattern then. This PR is fine as-is.

harrisonliew (approved):

PR #816, opened 2023-11-03
Improvements to supplies API

jerryz123:
- Change `vlsi.inputs.supplies.power/ground.pin` to `.pins`. This generates CPFs with a `create_global_connection` from  single net to multiple pins
- Add `power_straps_options.by_tracks.power_nets` setting. This controls which supply nets power straps are generated for. The default uses all supplies, but this is not always desired
- Add `vlsi.input.supplies.power.voltage` field. The existing system would always create the power_nets with the same `vlsi.inputs.supplies.VDD` voltage. This enables per-supply voltage setting.

**Related PRs / Issues**

**Type of change**:
- [ ] Bug fix
- [x] New feature
- [ ] Other enhancement

**Impact**:
- [x] Change to core Hammer
- [ ] Other

**Contributor Checklist**:
- [ ] Did you set `master` as the base branch?

### DIFF
--- a/hammer/config/defaults.yml
+++ b/hammer/config/defaults.yml
@@ -113,7 +113,7 @@ vlsi.inputs:
   # Supply voltages and names
   # Supply Struct:
   #   name (str) - The name of your supply's net
-  #   pin (Optional[str]) - The cell pin that this net should drive, no pins are connected when omitted
+  #   pins (Optional[List[str]]) - The cell pin/s that this net should drive. If None, pin name is assumed to be net name. If empty, no pins are connected
   #   tie (Optional[str]) - Another supply this supply is tied to. Nets with this set will  not end up in your final layout.
   #   weight (Optional[int]) - The weight of this supply when building power straps (larger number = more straps).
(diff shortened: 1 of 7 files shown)


hammer/config/defaults.yml
harrisonliew, line 1: Can you also change the documentation for `pin`?
I'd like for this to still be `Optional[List[str]]`. Default behaviour if not specified should be that it would connect to pins matching the net name only.

harrisonliew (changes requested):

hammer/config/defaults.yml
jerryz123: So the behavior should be 
 - None -> connect to pins matching the net name
 - Empty list -> connect to nothing
 - List -> connect to pins named in the list?

harrisonliew: Yes, that's perfect

jerryz123: Done, can you check the UPF spec gen? I'm not sure I did the right thing here.

harrisonliew (approved):

PR #819, opened 2023-11-07
Add top_layer/pg_pins_layers/stripe_pins flags to write_ilm for hierarchical flows

jerryz123:
**Related PRs / Issues**

**Type of change**:
- [ ] Bug fix
- [x] New feature
- [ ] Other enhancement

**Impact**:
- [x] Change to core Hammer
- [ ] Other

**Contributor Checklist**:

### DIFF
--- a/hammer/config/defaults.yml
+++ b/hammer/config/defaults.yml
@@ -608,6 +608,7 @@ par:
       # type: str
 
       pin_layers: [] # Layers to put power pins on
+      # This is also used in bottom-up hierarchical flows to configure LEF generation
       # type: List[str]
 
       strap_layers: [] # List of layers on which to place power straps (std cell rail layer is implied - do not include it)
--- a/hammer/par/innovus/__init__.py
+++ b/hammer/par/innovus/__init__.py
@@ -848,7 +848,16 @@ class Innovus(HammerPlaceAndRouteTool, CadenceTool):
         self.verbose_append("time_design -post_route")
         self.verbose_append("time_design -post_route -hold")
         self.verbose_append("check_process_antenna")
-        self.verbose_append("write_lef_abstract -5.8 {top}ILM.lef".format(top=self.top_module))
+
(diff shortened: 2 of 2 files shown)


harrisonliew:
Hm, can't we just set the stripe pins and top layer arguments based on the power straps pin_layer field?

jerryz123:
Is that general enough though for all use cases? For instance, If you have a power island in a block that instantiates its own bump, what would the flags to this command need to be?

I figured there will frequently be cases where you'll need some specialized settings to achieve what you want.

harrisonliew:
In the user guide:

In a bottom-up flow (where the block was not created using the create_partition command), you normally create a simple LEF abstract that contains:
- one obstruction (OBS) shape per routing layer
- signal pins located along the block boundary
- add power pins for the top metal layer PG stripes (see –stripe_pins and –pg_pin_layers options)
- embedded bump pins shapes are also written out as LEF pin shapes

So, it seems like `-stripe_pins` and `-pg_pin_layers` is sufficient. If you want more options, that's what a replacement hook is for. Otherwise adding all these keys for command options gets unwieldy.

jerryz123:
Ok, I'll change this then.

jerryz123:
Should we at least handle the case where manual power straps are used, in which case the `by_tracks.pin_layers` option is unset?

harrisonliew:
No, I think if you're doing manual straps, you're already using hooks and being mindful of stripe pin layers. At most, add an additional comment for the `by_tracks.pin_layers` key to note that this is used when writing out LEF/DEF of the layout (user can always set this key even if they're using manual straps).

jerryz123:
> No, I think if you're doing manual straps, you're already using hooks and being mindful of stripe pin layers. At most, add an additional comment for the `by_tracks.pin_layers` key to note that this is used when writing out LEF/DEF of the layout (user can always set this key even if they're using manual straps).

harrisonliew (approved):

PR #837, opened 2023-12-15
Support multiple rounds of opt_design in Innovus plugin

jerryz123:
Most designs should only run 1 round. 
This is very occasionally useful, and should be used judiciously 

**Related PRs / Issues**

**Type of change**:
- [ ] Bug fix
- [x] New feature
- [ ] Other enhancement

**Impact**:
- [ ] Change to core Hammer
- [ ] Other

**Contributor Checklist**:
- [ ] Did you set `master` as the base branch?

### DIFF
--- a/hammer/par/innovus/__init__.py
+++ b/hammer/par/innovus/__init__.py
@@ -637,7 +637,10 @@ class Innovus(HammerPlaceAndRouteTool, CadenceTool):
         Post-route optimization and fix setup & hold time violations.
         -expanded_views creates timing reports for each MMMC view.
         """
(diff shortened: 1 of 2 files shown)


harrisonliew (commented):
Not a fan of this, for a few reasons:
- There appears to be an `-incremental` flag, although this disables hold & power opts
- Even more targeted: a `-targeted` flag to only change specific paths
- We should probably change the opt category attributes to tune this behavior. User guide also suggests this between iterations of opt.
- If hold is a concern, we can do an `opt_design -post_cts -hold` at the end of CTS
- We should also be using `opt_signoff` after `opt_design` anyway

PR #841, opened 2024-02-12
Relocate CPF Read + Power Strap Issues

kenhoberkeley:
**Related PRs / Issues**
- CPF intent can only be read when after ILM flattening. It is not possible to flatten ILMs before floorplanning. Hence, move the CPF read to later in the flow. 
-  Power straps continue to interact with routing within the halos. Proposed solution is like so: 1) Routing halo generation remains the same. Placement halo is slightly increased in size. A power routing blockage is then created in same shape as the placement halo. This allows router to avoid power straps in a non-routing-halo region, before subsequently entering the halo region (this seems to work better). Also, do this up to the top layer instead of layer-1.

**Type of change**:
- [X ] Bug fix
- [ ] New feature
- [ ] Other enhancement

**Impact**:
- [ ] Change to core Hammer
- [X ] Change to a Hammer plugin
- [ ] Other

**Contributor Checklist**:
- [X ] Did you set `master` as the base branch?
- [X ] Did you state the type-of-change/impact?
- [ X] Did you delete any extraneous prints/debugging code?
- [ X] (If applicable) Did you add documentation for the feature?
- [ X] (If applicable) Did you update the `poetry.lock` file if you updated the requirements in `pyproject.toml`?
- [ X] (If applicable) Did you add a unit test demonstrating the PR?
- [ X] (If applicable) Did you run this through the e2e integration tests?
- [ X] (If applicable) Did you update the submodules in `e2e/` if this feature depends on updated plugins?

### DIFF
--- a/hammer/par/innovus/__init__.py
+++ b/hammer/par/innovus/__init__.py
@@ -486,7 +486,8 @@ class Innovus(HammerPlaceAndRouteTool, CadenceTool):
                     assign_arg = "-assign {{ {x} {y} }}".format(x=pin.location[0], y=pin.location[1])
 
                 layers_arg = ""
-                if set(pin.layers).intersection(set(power_pin_layers)):
+                
+                if set(pin.layers or []).intersection(set(power_pin_layers)):
                     self.logger.error("Signal pins will be generated on the same layer(s) as power pins. Double-check to see if intended.")
                 
                 if pin.layers is not None and len(pin.layers) > 0:
@@ -1075,7 +1076,10 @@ class Innovus(HammerPlaceAndRouteTool, CadenceTool):
                         output.append("create_route_halo -bottom_layer {b} -space {s} -top_layer {t} -inst {inst}".format(     
                             inst=new_path, b=bot_layer, t=current_top_layer, s=spacing))
                         
-                        place_push_out = round(spacing*1.2 , 1) # Push the place halo, and therefore PG blockage, further out from route halo so router is aware of straps before entering final routing.  
+                        if(self.get_setting("par.power_to_route_blockage_ratio") < 1):
+                            self.logger.warning("The power strap blockage region is smaller than the routing halo region for hard macros. Double-check if this is intended.")
+                            
+                        place_push_out = round(spacing*self.get_setting("par.power_to_route_blockage_ratio") , 1) # Push the place halo, and therefore PG blockage, further out from route halo so router is aware of straps before entering final routing.  
 
                         output.append("create_place_halo -insts {inst} -halo_deltas {{{s} {s} {s} {s}}} -snap_to_site".format(  
                             inst=new_path, s=place_push_out))
--- a/hammer/par/innovus/__init__.py
+++ b/hammer/par/innovus/__init__.py
@@ -1074,8 +1074,6 @@ class Innovus(HammerPlaceAndRouteTool, CadenceTool):
 
                         output.append("create_place_halo -insts {inst} -halo_deltas {{{s} {s} {s} {s}}} -snap_to_site".format(  
                             inst=new_path, s=place_push_out))
-                        #output.append("set pg_blockage_shape [lindex [get_computed_shapes -output polygon [get_db [get_db hinsts {inst}][get_db insts {inst}] .place_halo_polygon] SIZE {pg_pullback}] 0]".format(
-                        #    inst=new_path, pg_pullback=pg_pullback))
                         output.append("set pg_blockage_shape [get_db [get_db hinsts {inst}][get_db insts {inst}] .place_halo_polygon]".format(
                             inst=new_path))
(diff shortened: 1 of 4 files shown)


hammer/tech/stackup.py
```diff
@@ -406,6 +406,22 @@ def get_metals_below_layer(self, name: str) -> List[Metal]:
...
+    def get_metals_incl_layer(self, name: str) -> List[Metal]:
+        """
+        Get all the metals including the specified metal layer.
+
+        :param index: Index of the metal layer
+        :return: A list of metal layer objects
+        """
+        try:
+            index = next(m.index for m in self.metals if m.name == name)
+            if index > 1:
```
harrisonliew, line 418: Seems unnecessary to check that the list has to have length > 1.

hammer/par/innovus/__init__.py
```diff
@@ -1065,14 +1066,19 @@ def generate_floorplan_tcl(self) -> List[str]:
                         current_top_layer = None
                     if current_top_layer is not None:
                         bot_layer = self.get_stackup().get_metal_by_index(1).name
-                        cover_layers = list(map(lambda m: m.name, self.get_stackup().get_metals_below_layer(current_top_layer)))
-                        output.append("create_place_halo -insts {inst} -halo_deltas {{{s} {s} {s} {s}}} -snap_to_site".format(
-                            inst=new_path, s=spacing))
-                        output.append("create_route_halo -bottom_layer {b} -space {s} -top_layer {t} -inst {inst}".format(
+                        cover_layers = list(map(lambda m: m.name, self.get_stackup().get_metals_incl_layer(current_top_layer)))
```
harrisonliew, line 1069: Doesn't this prevent power straps by abutment on the top layer?

hammer/par/innovus/__init__.py
```diff
@@ -1065,14 +1066,19 @@ def generate_floorplan_tcl(self) -> List[str]:
...
-                        output.append("create_route_blockage -pg_nets -inst {inst} -layers {{{layers}}} -cover".format(
+                        
+                        place_push_out = round(spacing*1.2 , 1) # Push the place halo, and therefore PG blockage, further out from route halo so router is aware of straps before entering final routing.  
+
+                        output.append("create_place_halo -insts {inst} -halo_deltas {{{s} {s} {s} {s}}} -snap_to_site".format(  
+                            inst=new_path, s=place_push_out))
+                        output.append("set pg_blockage_shape [get_db [get_db hinsts {inst}][get_db insts {inst}] .place_halo_polygon]".format(
+                            inst=new_path))
+                        output.append("create_route_blockage -pg_nets -layers {{{layers}}} -polygon $pg_blockage_shape".format(
                             inst=new_path, layers=" ".join(cover_layers)))
```
harrisonliew, line 1080: no `inst` in format to fix type checking

hammer/par/innovus/__init__.py
kenhoberkeley: It would, but I'm not sure how we could avoid the current issue if pins are also generated on the top layer as well.

harrisonliew: Oh, I missed that. Fundamentally, pins should not be generated on the power top layer. I would advocate for enforcing this in `place_pins` instead of doing this halo-ing.

kenhoberkeley: Sounds good, so I will add code to:
1) Stop signal pins from being generated on the top layer of a macro. 
2) Implement the haloing scheme only for layers below the top layer.

harrisonliew: I think it should be a logger error but not kill the run (in case it is intentional).

hammer/par/innovus/__init__.py
kenhoberkeley: Fixed

hammer/par/innovus/__init__.py
kenhoberkeley: Added logger error when power pins are found on same layer as signal pins.

hammer/tech/stackup.py
kenhoberkeley: Fixed. Though function not used, leaving in because it could be useful.

harrisonliew (approved):

hammer/par/innovus/__init__.py
```diff
@@ -483,6 +486,9 @@ def place_pins(self) -> bool:
                     assign_arg = "-assign {{ {x} {y} }}".format(x=pin.location[0], y=pin.location[1])
 
                 layers_arg = ""
+                if set(pin.layers).intersection(set(power_pin_layers)):
```
harrisonliew, line 489: One of these things in the set is Optional - may need a `get_or_else`

kenhoberkeley: Fixed.

PR #845, opened 2024-03-06
support for mesh powerstrap

yolowiz:
**Related PRs / Issues**

**Type of change**:
- [ ] Bug fix
- [x] New feature
- [ ] Other enhancement

**Impact**:
- [x] Change to core Hammer
- [ ] Other

**Contributor Checklist**:
- [x] (If applicable) Did you add a unit test demonstrating the PR?


harrisonliew (changes requested):
Can you update the documentation with a diagram using this new method? I'd like to see whether the utilization property makes sense (recommend 0.2 for diagram).

doc/Hammer-Use/mesh.png
harrisonliew, line 1: I'm confused about this picture. Isn't only the bottom half actually the "mesh" method? The top half is existing "coupled" with track_spacing = 1.

yolowiz: It's an illustration of how twt is used.

harrisonliew: In my opinion, that is unnecessary (you can just overlay measurement e in the bottom half).

If anything, the existing diagrams above this should indicate the differences in calculation between twwt and twt when track_spacing = 0 or > 0.

doc/Hammer-Use/mesh_description.png
harrisonliew, line 1: Much better. But isn't there a slight error for mesh straps? If it's 50% utilization, we should see:
 VDD | signal | VSS | signal | VDD | signal ...
Right now there are 2 signals between each power => 33% utilization

yolowiz: That arises from the fact that power utilization is not exact for power strap calculation: track_spacing is part of the consumed_tracks, and for mesh, track_spacing is hard-coded to 1.

harrisonliew: Oh, this is your hack to force twt calculation, right? can you force that calculation without needing to account for the extra spacing?

yolowiz: To keep the power utilization consistent with twt, yes. I can try to subtract the track_spacing after twt returns the width and spacing.

harrisonliew: Yes, please subtract that extra track_spacing in mesh mode so that the calculation is intuitive.

harrisonliew (approved):

PR #848, opened 2024-03-08
Sky130 - Convert to Pydantic, support sky130_scl

harrisonliew:
- Implemented abstract method `gen_config` for tech plugins to override and generate their own config (`TechJSON` instance, i.e. directly creating the Pydantic model)
- Removed all existing sky130-tech-gen-... scripts and outputs, static `sky130.tech.json`
- Fixed `gds_map_file` reading in OpenROAD plugin

TODOs:
- [ ] Add sky130_scl Cadence standard cell library option
- [ ] Documentation for this alternate method
- [ ] Sky130 documentation for selecting library
- [ ] e2e tests

**Related PRs / Issues**

**Type of change**:
- [ ] Bug fix
- [X] New feature
- [X] Other enhancement

**Impact**:
- [X] Change to core Hammer
- [X] Change to a Hammer plugin
- [ ] Other

**Contributor Checklist**:


hammer/utils/lib_utils.py
```diff
@@ -53,4 +53,6 @@ def get_headers(source: str) -> List[str]:
             # TODO: handle other compressed file types? Tools only seem to support gzip.
             fd = os.open(source, os.O_RDONLY)
             lines = os.pread(fd, nbytes, 0).splitlines()
-        return list(map(lambda l: l.decode('ascii', errors='ignore'), lines))
+        lines = list(map(lambda l: l.decode('ascii', errors='ignore'), lines))
+        all_lines = '\n'.join(lines).replace('\\\n','')  # combine lines that are split by \ character
```
nayiri-k, line 57: @harrisonliew This might be too hacky, but basically in one of the sky130 lib files (`sky130_ef_io__gpiov2_pad_wrapped_ss_ss_100C_1v60_3v00.lib`) the `capacitive_load_unit` declaration spanned 2 lines and it messed up the parsing in `LIBUtils.get_cap_unit()`
```
	capacitive_load_unit(1.000000, \
	  "pf");
```

nayiri-k (approved):
Passed through commercial/openroad e2e flow

hammer/utils/lib_utils.py
harrisonliew: I think that's valid. But maybe the better approach is to remove line continuations (\) before processing lines. Can you also fix the type checker error?

hammer/technology/sky130/__init__.py
```diff
@@ -194,80 +585,103 @@ def setup_io_lefs(self) -> None:
...
 
                 # Current version has two one-off error that break lef parser.
-                self.logger.info("Fixing broken sky130_ef_io__analog_esd_pad LEF definition.")
-                start_broken_macro_list = ["MACRO sky130_ef_io__analog_esd_pad\n", "MACRO sky130_ef_io__analog_pad\n"]
-                end_broken_macro_list = ["END sky130_ef_io__analog_pad\n", "END sky130_ef_io__analog_noesd_pad\n"]
-                end_fixed_macro_list = ["END sky130_ef_io__analog_esd_pad\n", "END sky130_ef_io__analog_pad\n"]
+                self.logger.info(
+                    "Fixing broken sky130_ef_io__analog_esd_pad LEF definition.")
+                start_broken_macro_list = []
+                    #"MACRO sky130_ef_io__analog_esd_pad\n", "MACRO sky130_ef_io__analog_pad\n"]
```
elamdf, line 617: these don't seem to exist in the io spice file (causing this to error), is that correct?

harrisonliew: No, from what I can tell, the reason why that's there is because the END MACRO <macro_name> doesn't match the macro name declared in the beginning of the the MACRO block.

elamdf: oh is that why this breaks the parsing? it seems to run fine as it is here

elamdf: turns out the instructional machine lef is fixed. will remove the hack

elamdf:
syn/par works on commercial/openroad tutorial flow

elamdf:
`make  tutorial=sky130-commercial power-syn BINARY=/tools/C/elamdf/chipyard_dev/tests/symmetric.riscv` works as well

elamdf:
This will be merged once the cadence stdcells can produce a drc/lvs (with calibre and ideally pegasus too) design of reasonable complexity

.github/workflows/e2e-cad.yml
```diff
@@ -0,0 +1,55 @@
...
+          # run cadence tools on design
+          echo "log file is $tempfile on $(hostname)" | tee -a $tempfile
+          echo "End-to-end CAD flow CI smoketest running on" ${{ github.head_ref }}.${{ github.sha }} > $tempfile 2>&1
+          make build >> $tempfile 2>&1
+          echo "running par.." | tee -a $tempfile
+          make par >> $tempfile 2>&1
+          echo "running drc.." | tee -a $tempfile
+          make drc >> $tempfile 2>&1
+          echo "running lvs.." | tee -a $tempfile
+          make lvs >> $tempfile 2>&1
```
abejgonzalez, line 37: I'm not a huge fan of dumping something out to a file if it's sensitive. What happens if a current Hammer user accidentally removes one of these echos on accident and info is leaked? Maybe it's better to try to disable actions from logging entirely?

elamdf: Hmm, I see your point but I don't see how that reasoning doesn't apply to any output masking/secret hiding in an action. I've added a warning comment to hopefully dissuade accidental leaking, I'm not really sure how to [actually resolve this concern](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#self-hosted-runner-security)- maybe only allow maintainers to manually kick off the workflow once they've checked if no malicious changes to the actions? That seems like a single point of failure though

elamdf:
will make a separate PR for CI smoketest

hammer/tech/__init__.py
```diff
@@ -323,6 +323,77 @@ def ensure_dirs_exist(self, path) -> None:
...
+                    cached_paths = [f for f in cached_paths if os.path.exists(f)]
+
+                    # only allow 1 valid cached path, otherwise ambiguous
+                    if len(cached_paths) > 1:
+                        self.logger.error(f"ambiguous cache override: {cached_paths}")
+                    if cached_paths:
+                        used_overrides.append((field_name, fname))
+                        new_path = cached_paths[0]
+                        self.logger.info(
+                            f"overriding {default_path} with cache entry {new_path}"
```
elamdf, line 375: this kind of blows up the log since there are a bunch of instances of the same library...

hammer/technology/sky130/__init__.py
```diff
@@ -383,86 +1083,97 @@ def sky130_connect_nets2(ht: HammerTool) -> bool:
...
 set_db add_endcaps_boundary_tap     true
 set_db add_endcaps_left_edge        {endcap_cell}
 set_db add_endcaps_right_edge       {endcap_cell}
 add_endcaps
-    '''
+    """
     )
     return True
 
-def efabless_ring_io(ht: HammerTool) -> bool:
```
elamdf, line 400: I think we should create a separate PR for a robust toplevel-ready sky130 flow. Recent tapeouts have required a bunch of additional hooks in `hammer-driver`, I think just having this in sky130/__init__ is a little misleading (since other stuff not currently in hammer is also required for a clean run)

PR #850, opened 2024-03-12
Switch Typechecker from mypy to pyright

vighneshiyer:
Pyright is a Python typechecker and LSP that is faster and has tighter IDE integration vs mypy. It also doesn't require weird hacks for ruamel and networkx to typecheck. It also is a bit more strict than mypy. This was just a fun experiment to see what things pyright might catch that mypy doesn't. Feel free to investigate and merge if you'd like.

**Related PRs / Issues**

**Type of change**:
- [ ] Bug fix
- [ ] New feature
- [x] Other enhancement

**Impact**:
- [x] Change to core Hammer
- [x] Other

**Contributor Checklist**:
- [x] (If applicable) Did you add documentation for the feature?
- [x] (If applicable) Did you update the `poetry.lock` file if you updated the requirements in `pyproject.toml`?


hammer/sim/vcs/__init__.py
```diff
@@ -259,7 +263,9 @@ def run_simulation(self) -> bool:
         if self.level == FlowLevel.RTL and saif_mode != "none":
             find_regs_run_tcl = []
             if saif_mode != "none":
+                stime: Optional[TimeValue] = None
                 if saif_mode == "time":
+                    assert saif_start_time
                     stime = TimeValue(saif_start_time[0])
```
vighneshiyer, line 269: There is something fishy here - `saif_start_time` is a string and we're just picking out the first character assuming that the string looks like "3ns"

harrisonliew: Line 269 seems like a typo. It shouldn't have `[0]` in it (see line 282 - `TimeValue` can be created directly from the string with units).

pyproject.toml
```diff
@@ -83,3 +83,9 @@ reports = true
 # Disable len-as-condition since len(lst) == 0 is more clear on types
 # where as if <var> is confusing since it's not clear what <var> is.
 disable = ["no-else-return", "len-as-condition"]
+
+[tool.pyright]
+exclude = ["doc/**", "e2e/**", "test_scripts/**", "hammer/technology/asap7/gds_scale.py"]
```
harrisonliew, line 88: Is gds_scale.py excluded due to the conditional gdstk/gdspy imports?

harrisonliew (changes requested):
Thanks!!! This looks like a pretty straightforward migration. What's still left to do (i.e. why is it in draft)?

hammer/sim/vcs/__init__.py
vighneshiyer: Yes I think you're right, I removed the `[0]`

hammer/vlsi/hammer_tool.py
```diff
@@ -273,8 +273,6 @@ def run_dir(self) -> str:
             return self._rundir
         except AttributeError:
             return self._rundir
-        except AttributeError:
```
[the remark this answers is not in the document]
vighneshiyer: Yes you're right, I reverted this change and deleted the above lines instead.

pyproject.toml
vighneshiyer: I removed this exclusion and added one type ascription to the `gds_scale.py`, which seems to satisfy pyright

harrisonliew (approved):

PR #859, opened 2024-04-26
MXHammer experimental version

Miguel4141:
Added AMS simulation support through the Xcelium plugin, with hard-coded additions to arguments and constraints as of now.

**Related PRs / Issues**

**Type of change**:
- [ ] Bug fix
- [X] New feature
- [ ] Other enhancement

**Impact**:
- [ ] Change to core Hammer
- [X] Change to a Hammer plugin
- [ ] Other

**Contributor Checklist**:

### DIFF
--- a/hammer/sim/xcelium/__init__.py
+++ b/hammer/sim/xcelium/__init__.py
@@ -15,6 +15,8 @@ import os
 import json
 import datetime
 import io
+import re
+import logging # Remove later to use hammer logging
 from typing import Dict, List, Optional, Tuple, Any
 
 import hammer.tech as hammer_tech
@@ -23,6 +25,7 @@ from hammer.vlsi import HammerSimTool, HammerToolStep, HammerLSFSubmitCommand, H
 from hammer.logging import HammerVLSILogging
 from hammer.common.cadence import CadenceTool
 
+# MXHammer version
 
 class xcelium(HammerSimTool, CadenceTool):
 
@@ -30,17 +33,20 @@ class xcelium(HammerSimTool, CadenceTool):
   def xcelium_ext(self) -> List[str]:
     verilog_ext  = [".v", ".V", ".VS", ".vp", ".VP"]
     sverilog_ext = [".sv",".SV",".svp",".SVP",".svi",".svh",".vlib",".VLIB"]
+    verilogams_ext = [".vams", ".VAMS", ".Vams", ".vAMS"]
+    vhdl_ext = [".vhdl", ".VHDL"]
+    scs_ext = [".scs", ".SCS", ".sp", ".SP"] 
     c_cxx_ext    = [".c",".cc",".cpp"]
-    gz_ext       = [ext + ".gz" for ext in verilog_ext + sverilog_ext]
-    z_ext        = [ext + ".z" for ext  in verilog_ext + sverilog_ext]
-    return (verilog_ext + sverilog_ext + c_cxx_ext + gz_ext + z_ext)
+    gz_ext       = [ext + ".gz" for ext in verilog_ext + sverilog_ext + verilogams_ext + scs_ext]
+    z_ext        = [ext + ".z" for ext  in verilog_ext + sverilog_ext + verilogams_ext + scs_ext]
+    return (verilog_ext + sverilog_ext + verilogams_ext + vhdl_ext + scs_ext + c_cxx_ext + gz_ext + z_ext)
 
   @property
   def steps(self) -> List[HammerToolStep]:
     return self.make_steps_from_methods([self.compile_xrun,
                                          self.elaborate_xrun,
                                          self.sim_xrun])
-    
+
   def tool_config_prefix(self) -> str:
     return "sim.xcelium"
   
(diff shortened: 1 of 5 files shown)


hammer/sim/xcelium/__init__.py
```diff
@@ -377,6 +381,336 @@ def generate_sim_tcl(self) -> bool:
...
+
+    # Write filled template to AMS control file.
+    f.write(template)
+
+    # Close files properly.
+    t.close()
+    f.close()
+    return True
+  
+  def retrieve_files(path, exts=[], output_type=str, relative=True):
```
harrisonliew, line 439: This seems like a generally-useful utility. Consider moving it to `hammer/utils/__init__.py`?

hammer/sim/xcelium/__init__.py
```diff
@@ -377,6 +381,336 @@ def generate_sim_tcl(self) -> bool:
...
+      discipline_opts_proc = [f"-SETD \"{opt}\"" for opt in discipline_opts]
+    else:
+      discipline_opts = []
+
+    return discipline_opts_proc
+  
+  def extract_ams_opts(self, step):
+    ams_opts = self.get_setting(f"{self.tool_config_prefix()}.ams_opts", [])
+    cwd = os.getcwd()
+    match (step):
```
harrisonliew, line 583: Technically, match/case is Python 3.10+. We can discuss if we want to bump the minimum Python version.

hammer/sim/xcelium/__init__.py
```diff
@@ -427,35 +766,415 @@ def elaborate_xrun(self) -> bool:
...
+    if self.get_setting(f"{self.tool_config_prefix()}.ams"):
+      self.run_mxh_pseudo_three_step()
+      return True
+
+
     self.run_executable(args, cwd=self.run_dir)
     return True
 
+  
+  def sift_exts(list, exts):
```
harrisonliew, line 825: Ditto this for making general utility.
But more generally, in `hammer/utils/__init__.py`, there's already a `HammerFileType` Enum. I'd recommend doing this retreival/sifting and preparation methods below by just doing map/filter/reduce operations, where the partial function checks `HammerFileType.get_filetype(<file>) == HammerFileType.VERILOG` or similar. Feel free to expand the enum as necessary.

hammer/sim/xcelium/__init__.py
```diff
@@ -427,35 +766,415 @@ def elaborate_xrun(self) -> bool:
...
+    exts_lower = [extension.lower() for extension in exts]
+    exts_proc = [f".{ext}" if ("." not in ext) else ext for ext in exts_lower]
+
+    list_proc = [os.path.splitext(path.lower()) for path in list]
+
+    sifted = [f"{name}{ext}" for (name, ext) in list_proc if ext in exts_proc]
+
+    return sifted
+
+  def retrieve_files(path, exts=[], output_type=str, relative=True):
```
harrisonliew, line 838: Duplicate method?

hammer/sim/xcelium/__init__.py
Miguel4141: This is actually an outdated method from when I was trying to have AMS to follow the same three-step method as the base plugin. It'll probably be removed in the upcoming commit. As for the minimum python version more generally, I don't know if any of the features of 3.10 are worth forcing people with 3.9 to update the version.

hammer/sim/xcelium/__init__.py
```diff
@@ -458,4 +556,337 @@ def sim_xrun(self) -> bool:
...
+
+      return header + f"{opts_proc_str}" + footer
+
+  def generate_amscf(self, template_filename, amscontrol_filename) -> bool:
+      """
+      Creates an AMS control file based on templated format with available analog models & schematics
+      """
+      
+      # Get analog models, schematics from directories specified in extralibs
+      extralib = self.get_setting("vlsi.technologies.extra_libraries")
```
harrisonliew, line 734: This key doesn't exist... what are you trying to do/specify here? Do you need fields added to the normal `vlsi.technology.extra_libraries` key? https://github.com/ucb-bar/hammer/blob/5277152e35f664a23e816c26fcc66e605173b19a/hammer/tech/__init__.py#L249

hammer/sim/xcelium/__init__.py
```diff
@@ -458,4 +556,337 @@ def sim_xrun(self) -> bool:
...
+  def scriptwriter(self, options, additional_options, collect=False, sourcedir="", blacklist=[], sourcelist=[]):
+    """
+    Writes all prepared files and arguments to the run_mxh shell script
+    """
+    runpath = os.path.join(self.run_dir, "run_mxh")
+
+    f = open(runpath, "w+")
+
+    # Write Shebang + xrun clean
+    f.write("#!/bin/csh -f\n#\nxrun -clean \\\n")
```
harrisonliew, line 827: Does this only work in csh? Generally we want users to use bash family terminals.

hammer/sim/xcelium/defaults.yml
```diff
@@ -1,11 +1,18 @@
 sim.xcelium:
-  # Tool version (e.g., "XCELIUM2103")
+   # Tool version (e.g., "XCELIUM2103")
```
harrisonliew, line 2: Revert extra space

hammer/sim/xcelium/defaults.yml
```diff
@@ -1,11 +1,18 @@
...
   version: "XCELIUM2103"
 
+  # Spectre version (e.g., "SPECTRE211")
```
harrisonliew, line 5: Would note that this is only required for AMS.

harrisonliew (changes requested):

PR #888, opened 2025-03-01
Expand environment variables in configuration files

daniellovell:
It is common to define paths to collateral files with environment variables, but currently if you include paths in your config/spec .yml file that use environment vars, you will get an error.

Consider the following example, I would like to add a new library to `vlsi.technology.extra_libraries`:

**Spec.yml**
```
vlsi.technology.extra_libraries:
    ## Sensor ADC
    - library:
        gds_file: "$WORK_DIR/vlsi/sensor_adc_top.gds"
        lef_file: "$WORK_DIR/vlsi/sensor_adc_top_abstract.lef"
        verilog_synth: "$CHIP/vlsi/collaterals/utilities/scumvtop_gen/SENSOR_PLACE.v"
        nldm_liberty_file: "$WORK_DIR/vlsi/sensor_adc_top.lib"
        spice_file: "$CHIP/vlsi/collaterals/spice_netlists/SENSOR_PLACE.src.net"
```

If I attempt any operation using this library, I will get a ValueError because the envvars are not expanded
```
ValueError: Path $WORK_DIR/vlsi/sensor_adc_top.lib with prefix id \\
   $WORK_DIR did not match any tarballs or installs
```

This PR fixes the issue, by expanding all envvars in yaml & json files that are loaded by Hammer

**Related PRs / Issues**

**Type of change**:
- [ ] Bug fix
- [x] New feature
- [ ] Other enhancement

**Impact**:
- [x] Change to core Hammer
- [ ] Other

**Contributor Checklist**:

### DIFF
--- a/hammer/config/config_src.py
+++ b/hammer/config/config_src.py
@@ -1084,6 +1084,9 @@ def load_config_from_string(contents: str, is_yaml: bool, path: str = "unspecifi
     :param path: Path to the folder/package where the config file is located.
     :return: Loaded config dictionary, unpacked.
     """
+    # Expand any environment variables.
+    contents = os.path.expandvars(contents)
+
     unpacked = unpack(load_yaml(contents) if is_yaml else json.loads(contents))
     unpacked[_CONFIG_PATH_KEY] = path
     return unpacked


elamdf (approved):
nice

sunjin-choi:
@daniellovell @elamdf I agree this is useful, but there could be some corner cases when `subst` meta action is applied to the variable?

elamdf:
if the variable doesn't exist in the environment, `expandvars` will leave it untouched. So this would only be an issue if you had an environment variable like `export technology.sky130.sky130A="/badpath"`, which I agree is possible but is pretty unlikely imo unless some serious hacking is going on (or if hammer is messing with environmental variables to keep some state)

elamdf:
that's a good point- perhaps we should be doing a check to make sure that env variables aren't overriding lazy yaml variables (or at least printing a warning or something)

